Issue #8, opened 2019-01-01
Delete a pod, the old volume is not attached to the pod correctly


jia-jerry:
1. k8s, single worker node, version 1.11.0
2. cis, version v0.3.0
3. apply example Yaml below.
4. get into nginx pod, and create a file in /data by ```touch test```
5. delete current nginx pod
6. get into the new nginx pod, the /data folder is wrongly mounted.
---
Root cause: the disk is attached from the machine.

```
apiVersion: v1
kind: PersistentVolumeClaim
metadata:
  name: disk-pvc
spec:
  accessModes:
  - ReadWriteOnce
  resources:
    requests:
      storage: 25Gi
  storageClassName: csi-disk
---
apiVersion: apps/v1
kind: Deployment
metadata:
  name: nginx-deployment1
  labels:
    app: nginx
spec:
  selector:
    matchLabels:
      app: nginx
  template:
    metadata:
      labels:
        app: nginx
    spec:
      containers:
      - name: nginx
        image: nginx:1.7.9
        ports:
        - containerPort: 80
        volumeMounts:
          - name: disk-pvc
            mountPath: "/data"
      volumes:
        - name: disk-pvc
          persistentVolumeClaim:
            claimName: disk-pvc
```

jia-jerry:
BTW, each time of amount, it is not correct to format the disk always.

jia-jerry:
Logs in csi-diskplugin:

I0111 08:32:38.904651    9281 utils.go:96] GRPC call: /csi.v0.Node/NodeGetCapabilities
I0111 08:32:38.904676    9281 utils.go:97] GRPC request:
I0111 08:32:38.904714    9281 nodeserver-default.go:56] Using default NodeGetCapabilities
I0111 08:32:38.904718    9281 utils.go:102] GRPC response: capabilities:<rpc:<> >
I0111 08:32:38.905493    9281 utils.go:96] GRPC call: /csi.v0.Node/NodeGetCapabilities
I0111 08:32:38.905505    9281 utils.go:97] GRPC request:
I0111 08:32:38.905538    9281 nodeserver-default.go:56] Using default NodeGetCapabilities
I0111 08:32:38.905542    9281 utils.go:102] GRPC response: capabilities:<rpc:<> >
I0111 08:32:38.910808    9281 utils.go:96] GRPC call: /csi.v0.Node/NodePublishVolume
[...]

Issue #7, opened 2018-12-31
Disk is not detached successfully in 0.3.0


jia-jerry:
_k8s verison: 1.11.0
csi plugin versions: v0.3.0
There are 3+ pvs mounted on a single VM._
When I try to remove  all the PVs at the same time, at least 1 PV can't be removed.

Root cause:
1. NodeUnpublishVolume returns error. it is 
```
 time="2018-12-27T08:24:27Z" level=error msg="NodeUnpublishVolume: detach error %!(EXTRA string=d-uf62khufqdz7w8hybmu9, string=Aliyun API Error: RequestId: 885218BD-391A-4C6B-BA40-0FCA3EB6019E Status Code: 403 Code: InvalidOperation.Conflict Message: null)"
E1227 08:24:27.715595    2106 utils.go:100] GRPC error: rpc error: code = Internal desc = Detach error:Aliyun API Error: RequestId: 885218BD-391A-4C6B-BA40-0FCA3EB6019E Status Code: 403 Code: InvalidOperation.Conflict Message: null
```
Though this function returns an error, Kubelete did NO retry to called. And Provisioner calls DeleteVolume directly. This will always fails because the volume is not detached from the machine.

2. Why error InvalidOperation.Conflict happens?
Because Alicloud API call has rate limeter.

fredkan:
@jia-jerry InvalidOperation.Conflict is caused by cloud disk status issue. this is a issue from backend.

" Kubelete did NO retry to called" we check the kubelet csi-plugin interface for this issue.

fredkan:
Issue resolved, close this. Thanks @jia-jerry

PR #10, opened 2019-01-14
Fix issue of delete pod and can't reattach issue. Enable NodeUnstageV…

jia-jerry:
fix issues
#8 
#7

### DIFF
--- a/pkg/disk/attach_helper.go
+++ b/pkg/disk/attach_helper.go
@@ -1,0 +1,36 @@
+package disk
+
+import (
+	"fmt"
+	"io/ioutil"
+	"strings"
+)
+
+func getDevices() []string {
+	devices := []string{}
+	files, _ := ioutil.ReadDir("/dev")
+	for _, file := range files {
+		if !file.IsDir() && strings.HasPrefix(file.Name(), "vd") {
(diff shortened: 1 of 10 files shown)


pkg/disk/attach_helper.go
```diff
@@ -0,0 +1,36 @@
...
+	for _, file := range files {
+		if !file.IsDir() && strings.HasPrefix(file.Name(), "vd") {
+			devices = append(devices, fmt.Sprintf("/dev/%s", file.Name()))
+		}
+	}
+
+	return devices
+}
+
+func calcNewDevices(old, new []string) []string {
```
peter-wangxu: @jia-jerry   Just a question: why not simply use the field `Device` of http://docs.aliyun.com/#/pub/ecs/open-api/datatype&diskitemtype

jia-jerry: Aliyun doesn't always return correct device. You may refer to @denverdino. He said it is an kernal issue and can't fix. 

As a result, the implementation of CSI is awkward. We can't simple use ControllerPublishVolume simple to attach instance. Never try this way.

peter-wangxu: I have concern if there is issue when concurrent attach occurs for the same ecs instance. any comment?

PR #13, opened 2019-02-10
Correct node id (in NodeInfo) for kube v1.12.X

jia-jerry:
Support 1.12.X

### DIFF
--- a/deploy/disk/disk-plugin.yaml
+++ b/deploy/disk/disk-plugin.yaml
@@ -45,14 +45,9 @@ spec:
             allowPrivilegeEscalation: true
           image: registry.cn-hangzhou.aliyuncs.com/plugins/csi-diskplugin:***
           args :
-            - "--nodeid=$(NODE_ID)"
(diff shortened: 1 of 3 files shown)


deploy/disk/disk-plugin.yaml
```diff
@@ -45,14 +45,9 @@ spec:
             allowPrivilegeEscalation: true
           image: registry.cn-hangzhou.aliyuncs.com/plugins/csi-diskplugin:***
           args :
-            - "--nodeid=$(NODE_ID)"
```
fredkan: consider NodeId is not always same as Instance ID;

jia-jerry: Node ID is desinged for CSI plugin to broadcase it's instance ID for later detach/attache usage, as far as I know. It is **not** k8s node name.


Anyway, this PR suggests the correct way we use for NodeID. If you want to inject a NODE ID from environment, it is still OK.

fredkan: Agree, thanks Jerry.

PR #26, opened 2019-03-09
Support Topology

jia-jerry:


fredkan:
Great features. And 2 points need to consider.

1. the example is better to provide as new floder: may be named topology.
2. AccessibleTopology should be always enabled? if the use case not configure topology, we should support to config the zoneid in storageclass.

Thanks Jerry.

PR #31, opened 2019-05-21
V1.0.0

fredkan:
update oss to support upgrade


liquanstan:
@fredkan 
if we annotate the "/nsenter --mount=/proc/1/ns/mnt /usr/libexec/kubernetes/kubelet-plugins/volume/exec/alicloud~csiplugin-connector-init/csiplugin-connector-init init > /dev/null  2>&1" , where can we run csiplugin-connector ?

PR #34, opened 2019-06-06
V1.0.0

fredkan:
merge from v1.0.0


fredkan:
upgrade oss daemon with systemd.

PR #35, opened 2019-06-10
Add a note for `v1.15.0-alpha.3`

gwind:
Since `v1.15.0-alpha.3` , the `--allow-privileged` is removed from kubelet options.

https://github.com/kubernetes/kubernetes/blob/master/CHANGELOG-1.15.md#action-required-1

```
ACTION REQUIRED: The deprecated Kubelet flag --allow-privileged has been removed. Remove any use of --allow-privileged from your kubelet scripts or manifests.
```


fredkan:
thanks, welcome to join the ding talk group.

PR #36, opened 2019-06-10
fix node select policy for kubernetes from scratch

gwind:
If the kubernetes cluster is build from scratch ,  like [KFS](https://kfs.ooclab.com/kfs/v1.15/) , need drop this node select policy.


fredkan:
disk-attacher.yaml is removed, and provisioner changes a lot. Close this pr.

Thanks.

PR #58, opened 2019-08-06
remove isformat for disk plugin

jia-jerry:
Isformat can't check disk format correctly. Use SafeFormatAndMount.FormatAndMount() directly


fredkan:
Following Kubernetes is better. Thanks Jerry.

PR #167, opened 2019-12-26
v1.0.2 support memory volume

fredkan:
v1.0.2 support memory volume


fredkan:
include no expect code, close this pr.

PR #210, opened 2020-03-27
update lvm striping

fredkan:
update lvm striping


xianlubird:
commit missing GitHub user

PR #231, opened 2020-04-22
Fail to find csi-plugin and exit the program

lujinda:


fredkan:
close and will talk about other solution.

PR #251, opened 2020-05-28
On non-VF nodes, use device serial to get the disk path

lujinda:


pkg/disk/bdf.go
```diff
@@ -359,3 +360,37 @@ func forceDetachAllowed(disk *ecs.Disk, nodeID string) (allowed bool, err error)
 	// case 2
 	return inst.Status == InstanceStatusStopped, nil
 }
+
+var vfOnce = new(sync.Once)
+
+func IsVFNode() bool {
```
fredkan, line 366: run this code when set GlobalConfigVar.DiskBdfEnable value is better. not need running for every check process.

lujinda: I think  IsVFNode is more reliable.  The configuration(GlobalConfigVar.DiskBdfEnable) is easy to cause incorrect configuration due to human problems

PR #271, opened 2020-07-26
fix MountPoint volume remove error

liuchjlu:
fix remove mountpoint volume exception when the pv's nodeaffinity is nil.


fredkan:
lgtm

PR #289, opened 2020-08-17
Feature/monitor

JiaoDean:
Add node-server monitoring


main.go
```diff
@@ -99,6 +115,11 @@ var (
 	rootDir         = flag.String("rootdir", "/var/lib/kubelet/csi-plugins", "Kubernetes root directory")
 )
 
+type globalMetricConfig struct {
+	enableMetric string
```
fredkan, line 119: why not bool type?

PR #295, opened 2020-08-27
Support upgrade from k8s 1.16 to 1.18

mowangdk:
Support upgrade from k8s 1.16 to 1.18

### DIFF
--- a/pkg/disk/utils.go
+++ b/pkg/disk/utils.go
@@ -249,6 +249,7 @@ func IsFileExisting(filename string) bool {
 	if err == nil {
 		return true
 	}
+	// Notice: this err may be is not dictionary error, it will returns true
 	if os.IsNotExist(err) {
 		return false
 	}


pkg/disk/utils.go
```diff
@@ -243,6 +243,25 @@ func GetDeviceMountNum(targetPath string) int {
...
 
+func getPrevBlockDevicePublishPath(targetPath string) string {
+	fsPathList := strings.Split(targetPath, "/")
+	return strings.Join(fsPathList[:len(fsPathList)-1], "/")
+}
+
+// CorrectPublishPath4Upgrate return prev version block publish path
+func CorrectPublishPath4Upgrate(targetPath string) string {
+	if strings.HasPrefix(targetPath, BLOCKVOLUMEPREFIX) {
+		PrevPublishPath := getPrevBlockDevicePublishPath(targetPath)
```
fredkan, line 254: use filepath.Dir() is better?

PR #372, opened 2020-12-10
add some fix

wenjun93:


pkg/local/lib/commands/commands.go
```diff
@@ -221,6 +221,9 @@ func CleanPath(ctx context.Context, path string) error {
 	}
 	errList := []error{}
 	for _, file := range files {
+		if file == "lost+found" {
```
TheBeatles1994, line 224: why skip this folder?

build/local/Dockerfile
```diff
@@ -1,8 +1,8 @@
-FROM ubuntu:18.10
+FROM alpine:3.10
```
xianlubird, line 1: why alpine

PR #453, opened 2021-05-27
[WIP]Support app consistent snapshot for disk

mowangdk:


pkg/utils/commands.go
```diff
@@ -0,0 +1,224 @@
...
+	}
+	return fmt.Errorf("ensureContainerExists:: container name [%s] not in [%s] pod", c.container, c.name)
+}
+
+func (c *PodCommands) setDefaultHookContainer(pod *corev1api.Pod) error {
+	if len(pod.Spec.Containers) < 1 {
+		return fmt.Errorf("need at least 1 container")
+	}
+
+	c.container = pod.Spec.Containers[0].Name
```
cuericlee, line 210: Containers[0] does not means the main database container for a multiple containers application, a sidecar and other container. container name should be specified in spec of rule or snapshot spec explicitly.

pkg/disk/controllerserver.go
```diff
@@ -653,10 +668,41 @@ func (cs *controllerServer) CreateSnapshot(ctx context.Context, req *csi.CreateS
...
+			log.Errorf("CreateSnapshot:: get pods count: %v by pvc: %v which is not support for now", len(pods), pvcName)
+		}
+	}
+	if preRule.Spec.ContainerName != postRule.Spec.ContainerName {
+		log.Errorf("CreateSnapshot:: container name in preRule & postRule are must be same")
+		podName = ""
+	}
+	if err == nil && preRule.Name != "" && postRule.Name != "" && podName != "" {
+		command := utils.NewPodCommands(volumeSnapshotNamespace, podName, preRule.Spec.ContainerName, GlobalConfigVar.ClientSet, GlobalConfigVar.kubeClientConfig)
+		err = command.ApplyPreCommands(preRule)
```
cuericlee, line 691: CreateSnapshot should be took place between ApplyPreCommands and ApplyPostCommands. Immediate ApplyPostCommands here will lead to release IO lock and snapshot will go to in-consistent condition.

cuericlee (commented):
let review impl of app consistent snapshot before merging it into main branch

PR #500, opened 2021-08-11
Support arm arch

mowangdk:


mowangdk:
TODO: update shell build command

PR #508, opened 2021-08-20
add kubeconfig support

dabaooline:
Hello

In some case, we need create kubeclient from kubeconfig. So I add kubeconfig support.


fredkan:
@dabaooline thanks!

PR #546, opened 2021-10-14
fix lvm plugin not expand volume sometimes

### LINKED ISSUE #545: [BUG] lvm plugin can not expand volume when volume size is less than 1GiB
Step to reproduce:
1. create a pv which size is 700MiB.
2. expand this pv to 1GiB

There is a bug in this code:
https://github.com/kubernetes-sigs/alibaba-cloud-csi-driver/blob/2be29b10347be9bb9c91c6c20d0e19c3021b3b14/pkg/lvm/nodeserver.go#L324-L337

This code compare 2 sizes, one is the pvSize, the other is the lvmSize(var `sizeInt` in this code)

Assume I have a pv which size is 700MiB, then expand this pv to 1GiB

cmd `lvdisplay xxxx` shows:
```
...
LV Size      700.00 MiB
...
```
This code will ignore the 'MiB' unit, and make `sizeInt` to `700`.
But the request `pvSize` is `1` which unit is `GiB`. 
Then `sizeInt >= pvSize` will be true, which is not correct.

joy717:
lvm plugin will not expand when pv size is less than 1G, and then expand to more than 1G
fix #545

### DIFF
--- a/pkg/lvm/nodeserver.go
+++ b/pkg/lvm/nodeserver.go
@@ -385,7 +385,7 @@ func (ns *nodeServer) getPvSize(volumeID string) (int64, int64, string) {
 
 // create lvm volume
 func (ns *nodeServer) createVolume(ctx context.Context, volumeID, vgName, pvType, lvmType string) error {
-	pvSize, unit := ns.getPvSize(volumeID)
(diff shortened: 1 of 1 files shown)


joy717:
done for cla

PR #553, opened 2021-10-25
fix repeat update pv annotations

dabaooline:
when csi run on large kubernetes cluster , controller/DeleteVolume may return error before waiting successfully delete lv. 

The extener-provisioner will retry. But the  controller/DeleteVolume not check the PV's annotations , update it to deleting.


fredkan:
thanks

PR #572, opened 2021-12-06
Adapt to openshift, read ak from /root/.alibabacloud/credentials path.

JiaoDean:
Adapt to openshift, read ak from /root/.alibabacloud/credentials path.


jsafrane:
I tested this PR with OpenShift, I got this log line from Disk CSI driver:

```
time="2021-12-13T16:07:14Z" level=info msg="Get AK: use Credential AK"
```

Therefore I assume it does what it should do. I can provision/attach/detach/delete a volume with this. I let others comment on the code.

PR #612, opened 2022-02-25
support multi-zones node for virtual node

nightmeng:
A virtual nodes can manage pods in different zones, so CSI provisioner needs to support multi-zone nodes.

After this feature is enabled, CSI provisioner can find all zones supported by the multi-zone nodes, and add affinity information compatible with multi-zone nodes for PV during PV production.


fredkan:
ok-to-test

PR #628, opened 2022-03-22
Add a parameter to the schema to automatically increase volume sizes …

kwoodson:
…to Alibaba minimum limit of 20Gi

#### What type of PR is this?
/kind feature

Currently Alibaba Cloud provider has a minimum volume size limit of 20 GiB. This current behavior causes the test cases for statefulsets (use 1KiB) as well as some PVC (use 1MiB) tests to fail.

This pull request adds a flag to the storage class, that when set, allows the cloud provider to provision the minimum size (currently 20GiB) for requested volumes that are smaller than this limit.

Fixes #

```release-note
Added allowIncrease flag to storageclass parameters for the ability to automatically increase volumes to the minimum size of 20GB. The flag is false by default.
```

```docs
```

### DIFF
--- a/pkg/disk/controllerserver.go
+++ b/pkg/disk/controllerserver.go
@@ -97,6 +97,10 @@ const (
 	DefaultVolumeSnapshotClass = "alibabacloud-disk-snapshot"
 	// annDiskID tag
 	annDiskID = "volume.alibabacloud.com/disk-id"
+	// MinimumDiskSizeInGB ...
+	MinimumDiskSizeInGB = 20
+	// MinimumDiskSizeInBytes ...
+	MinimumDiskSizeInBytes = 21474836480
 )
 
 const (
@@ -135,19 +139,20 @@ type controllerServer struct {
 
 // Alicloud disk parameters
 type diskVolumeArgs struct {
-	Type             string              `json:"type"`
-	RegionID         string              `json:"regionId"`
-	ZoneID           string              `json:"zoneId"`
-	FsType           string              `json:"fsType"`
-	ReadOnly         bool                `json:"readOnly"`
-	MultiAttach      string              `json:"multiAttach"`
-	Encrypted        bool                `json:"encrypted"`
-	KMSKeyID         string              `json:"kmsKeyId"`
-	PerformanceLevel string              `json:"performanceLevel"`
-	ResourceGroupID  string              `json:"resourceGroupId"`
-	DiskTags         string              `json:"diskTags"`
-	NodeSelected     string              `json:"nodeSelected"`
-	ARN              []ecs.CreateDiskArn `json:"arn"`
+	Type                    string              `json:"type"`
+	RegionID                string              `json:"regionId"`
+	ZoneID                  string              `json:"zoneId"`
+	FsType                  string              `json:"fsType"`
+	ReadOnly                bool                `json:"readOnly"`
+	MultiAttach             string              `json:"multiAttach"`
+	Encrypted               bool                `json:"encrypted"`
+	KMSKeyID                string              `json:"kmsKeyId"`
+	PerformanceLevel        string              `json:"performanceLevel"`
+	ResourceGroupID         string              `json:"resourceGroupId"`
+	DiskTags                string              `json:"diskTags"`
+	NodeSelected            string              `json:"nodeSelected"`
+	ARN                     []ecs.CreateDiskArn `json:"arn"`
+	VolumeSizeAutoAvailable bool                `json:"volumeSizeAutoAvailable"`
 }
 
 // Alicloud disk snapshot parameters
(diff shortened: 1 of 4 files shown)


kwoodson:
@jsafrane PTAL

pkg/cnfs/v1beta1/types.go
```diff
@@ -82,6 +82,7 @@ type Parameters struct {
 	ACLType              string  `json:"aclType,omitempty"`
 	EnableTrashCan       string  `json:"enableTrashCan,omitempty"`
 	TrashCanReservedDays string  `json:"trashCanReservedDays,omitempty"`
+	AllowIncrease        string  `json:"allowIncrease,omitempty"`
```
jsafrane, line 85: These are parameters for "cnfs" CSI driver, IMO it's not related to Disk CSI driver.

jsafrane (commented):
I like the PR!

pkg/cnfs/v1beta1/types.go
kwoodson: Will remove. Thanks!

kwoodson:
Looks like a networking issue:
```
remote: Cannot obtain refs from GitHub: cannot talk to GitHub:
```

kwoodson:
Fixing lint checks. Will attempt to replace the go.mod back to the original. The lint check happens in the ./hack/ dir but `go mod tidy && go mod vendor && go mod verify` removed lint.

kwoodson:
There is something strange with the lint package but the linting does pass locally.
```
./hack/check-golint.sh 
go get: installing executables with 'go get' in module mode is deprecated.
	To adjust and download dependencies of the current module, use 'go get -d'.
	To install using requirements of the current module, use 'go install'.
	To install ignoring the current module, use 'go install' with a version,
	like 'go install example.com/cmd@latest'.
	For more information, see https://golang.org/doc/go-get-install-deprecation
	or run 'go help get' or 'go help install'.
go get: upgraded golang.org/x/lint v0.0.0-20190409202823-959b441ac422 => v0.0.0-20210508222113-6edffad5e616
Congratulations! All Go source files have been linted.
```

pkg/disk/utils.go
```diff
@@ -1102,6 +1103,19 @@ func getDiskVolumeOptions(req *csi.CreateVolumeRequest) (*diskVolumeArgs, error)
 		diskVolArgs.ResourceGroupID = ""
 	}
 
+	// allowIncrease
+	value, ok = volOptions["allowIncrease"]
```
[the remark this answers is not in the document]
kwoodson: @fredkan Thanks for the review.

| 1, maybe "volumeSizeAutoAvailable" is better?
I don't have any preference to the naming of the flag. If you prefer this then that's fine. I have updated the flag.
| 2. this should be true by default, Otherwise the default storageclass will not work.
As @jsafrane has previously mentioned and from our conversation with Bo Teng, we have decided to allow this flag on a newly created storage class. This will also preserve the current behavior when creating volumes for customers who do not want the confusion of having larger disks than the PVC requested size.

pkg/disk/controllerserver.go
```diff
@@ -275,6 +278,10 @@ func (cs *controllerServer) CreateVolume(ctx context.Context, req *csi.CreateVol
 	}
 	volSizeBytes := int64(req.GetCapacityRange().GetRequiredBytes())
 	requestGB := int((volSizeBytes + 1024*1024*1024 - 1) / (1024 * 1024 * 1024))
+	if diskVol.AllowIncrease && requestGB < MinimumDiskSizeInGB {
+		log.Infof("CreateVolume: volume size was less than allowed limit. Setting request Size to %vGB. allowIncrease is set.", MinimumDiskSizeInGB)
+		requestGB = MinimumDiskSizeInGB
```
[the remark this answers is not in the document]
kwoodson: I have updated the `volSizeBytes` as per your request. The `volSizeBytes` is used when comparing and determining if a disk already exists.

I believe the pv size uses the `requestGB` when creating a new volume here:
https://github.com/kubernetes-sigs/alibaba-cloud-csi-driver/blob/master/pkg/disk/controllerserver.go#L321

kwoodson:
@jsafrane @fredkan 
Is there something I need to do for the linting check to pass from the github tests? When I run it locally it looks like it downloads the modules and passes the lint check.
```
./hack/check-golint.sh
go get: installing executables with 'go get' in module mode is deprecated.
	To adjust and download dependencies of the current module, use 'go get -d'.
	To install using requirements of the current module, use 'go install'.
	To install ignoring the current module, use 'go install' with a version,
	like 'go install example.com/cmd@latest'.
	For more information, see https://golang.org/doc/go-get-install-deprecation
	or run 'go help get' or 'go help install'.
Congratulations! All Go source files have been linted.
```

pkg/disk/utils.go
```diff
@@ -1102,6 +1103,19 @@ func getDiskVolumeOptions(req *csi.CreateVolumeRequest) (*diskVolumeArgs, error)
 		diskVolArgs.ResourceGroupID = ""
 	}
 
+	// volumeSizeAutoAvailable
+	value, ok = volOptions["volumeSizeAutoAvailable"]
+	if !ok {
+		diskVolArgs.VolumeSizeAutoAvailable = false
```
fredkan, line 1109: Also, here should be true as default.

kwoodson: @fredkan The plan was to not enable this by default and for this reason we set this to false by default.

fredkan:
check_golint.sh has issues, please rebase the code.
https://github.com/kubernetes-sigs/alibaba-cloud-csi-driver/blob/master/hack/check-golint.sh

kwoodson:
Rebased to clean up linting. @fredkan

PR #686, opened 2022-08-12
Add func: allow auto snapshot during expanding volume.

AlbeeSo:
#### What type of PR is this?

/kind feature

Add func : allows to use auto snapshot during expanding volume of ESSD.

Fixes #

```release-note

```

```docs

```

### DIFF
--- a/pkg/disk/controllerserver.go
+++ b/pkg/disk/controllerserver.go
@@ -1089,7 +1089,7 @@ func (cs *controllerServer) ControllerExpandVolume(ctx context.Context, req *csi
 	// do resize
 	resizeDiskRequest := ecs.CreateResizeDiskRequest()
 	resizeDiskRequest.RegionId = GlobalConfigVar.Region
-	resizeDiskRequest.DiskId = diskID
+	resizeDiskRequest.DiskId = disk.DiskName
 	resizeDiskRequest.NewSize = requests.NewInteger(requestGB)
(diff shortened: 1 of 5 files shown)


pkg/disk/controllerserver.go
```diff
@@ -103,6 +103,20 @@ const (
 	MinimumDiskSizeInGB = 20
 	// MinimumDiskSizeInBytes ...
 	MinimumDiskSizeInBytes = 21474836480
+	//volumeExpandAutoSnapshotIDKey is the key of volumeExpandAutoSnapshotID on pvc
+	volumeExpandAutoSnapshotIDKey = "volumeExpandAutoSnapshotID"
+	// volumeExpandAutoSnapshotNameKey is the key of volumeExpandAutoSnapshotIName on pvc
+	volumeExpandAutoSnapshotNameKey = "volumeExpandAutoSnapshotName"
```
mowangdk, line 109: IMO, Snapshot name is not necessary to mark on pvc?

AlbeeSo: controllerServer manages a map to record the created snapshot,  with the names of snapshot as the keys.

AlbeeSo: Sure it's okay to only mark the ID if the volumeExpandAutoSnapshots are not recorded in the controllerServer then the nodeServer will not need to delete it anymore.

mowangdk:
@AlbeeSo  Run ./hack/update-gofmt.sh  & repush to fix fmt err

PR #723, opened 2023-01-08
Fix ssd create with performaceLevel

mowangdk:

Fixes #

```release-note

```

```docs

```


huww98:
If storage class parameter is set to `type: cloud_essd,cloud_ssd`, the `CreateDisk` ECS request will have `PerformanceLevel` set, which results in ServerError InvalidDataDiskCategory.NotSupported, which results in error message:
`createDisk: err: %v, the zone:[%s] is not support specific disk type, please change the request disktype:`, despite that cloud_ssd is actually supported.

I think it is 2080b7bb4fe538ee62cda65b5aaee80fcd5b2334 that actually fixes this bug. But based on the commit message, I log the details here.

PR #793, opened 2023-06-26
bugfix: remove empty OSS mount point to avoid  mount point retain problem

AlbeeSo:

Fixes #

```release-note

```

```docs

```

### DIFF
--- a/pkg/oss/nodeserver.go
+++ b/pkg/oss/nodeserver.go
@@ -369,12 +369,14 @@ func (ns *nodeServer) NodeUnpublishVolume(ctx context.Context, req *csi.NodeUnpu
 	if err != nil {
 		return nil, err
 	}
-	
-	// check mount point with IsLikelyNotMountPoint
+
+	// check mount point with IsLikelyNotMountPoint first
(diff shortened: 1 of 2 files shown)


pkg/oss/nodeserver.go
```diff
@@ -367,9 +369,25 @@ func (ns *nodeServer) NodeUnpublishVolume(ctx context.Context, req *csi.NodeUnpu
...
+	// check mount point with IsLikelyNotMountPoint first
+	notmounted, err := ns.k8smounter.IsLikelyNotMountPoint(mountPoint)
+	if err != nil {
+		log.Errorf("NodeUnpublishVolume:: use IsLikelyNotMountPoint to check if path %s mounted failed with error %v", mountPoint, err)
+		notmounted = !IsOssfsMounted(mountPoint)
+	}
+
+	if notmounted {
+		if empty, _ := IsDirEmpty(mountPoint); empty {
+			if removeErr := os.Remove(mountPoint); removeErr != nil {
```
huww98, line 382: How about just call `os.Remove`? You don't need to check mounted or empty first, since `os.Remove` would fail in those cases.

PR #815, opened 2023-08-04
remove meaningless SafePathRemove function

huww98:
#### What type of PR is this?

/kind cleanup

remove meaningless SafePathRemove function

SafePathRemove checks whether the target path is mounted, or if it is a directory, check whether it is empty before removing. However, this approach is not atomic, thus unreliable. And os.Remove already properly returns error in these 2 cases:
* When removing a mount point, it returns "device or resource busy"
* When removing a non-empty directory, it returns "directory not empty"

```release-note

```

```docs

```


huww98:
close in favor of #833

PR #826, opened 2023-08-11
bugfix: silence metrics volume devices logic

mowangdk:
#### What type of PR is this?

/kind bug

Solve "Get dev name by diskID xxxx  failed, err:getDeviceByVolumeIDxxxxx with error Failed to run cmd: findmnt" error

None
None

None

```docs

```

### DIFF
--- a/pkg/metric/utils.go
+++ b/pkg/metric/utils.go
@@ -201,11 +201,12 @@ func getDeviceByVolumeID(pvName, volumeID string) (device string, err error) {
 	}
 
 	if device, err = getDeviceBySymlink(volumeID); err != nil {
+		// strange logic... ignore findmnt error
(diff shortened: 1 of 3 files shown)


tydra-wang (approved):

pkg/metric/utils.go
```diff
@@ -201,11 +201,12 @@ func getDeviceByVolumeID(pvName, volumeID string) (device string, err error) {
 	}
 
 	if device, err = getDeviceBySymlink(volumeID); err != nil {
+		// strange logic... ignore findmnt error
```
huww98, line 204: PolarFS maybe using volumeDevices?

huww98 (approved):
OK, but we still don't know why this `vol_data.json` is not cleaned up and referring to an non-exists disk. Anyway, `vol_data.json` should be the implementation detail of kubelet, which we should not rely on. I think we may get rid of reading `vol_data.json` in the future.

PR #827, opened 2023-08-12
Changes to get reasonable csi-sanity test results

huww98:
#### What type of PR is this?
/kind cleanup

Fix some low risk issue that prevent csi-sanity from producing reasonable result.

* `NodePublishVolume`: don't require `targetPath` to end with /mount
* in `CreateVolume`, `parameters` are optional and can be `nil`, fix the panic in this case
* `NodeUnpublishVolume` should delete `targetPath` as per [CSI spec](https://github.com/container-storage-interface/spec/blob/master/spec.md#nodeunpublishvolume):
  > The SP MUST delete the file or directory it created at this path.

```release-note
NONE
```

```docs

```

### DIFF
--- a/pkg/disk/controllerserver.go
+++ b/pkg/disk/controllerserver.go
@@ -237,6 +237,9 @@ func (cs *controllerServer) CreateVolume(ctx context.Context, req *csi.CreateVol
 	}
 
 	volumeContext := req.GetParameters()
+	if volumeContext == nil {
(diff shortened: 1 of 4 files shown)


huww98:
dropped commit "c81b86e5 oss/NodeUnpublishVolume: delete targetPath" to avoid conflict with teammates

PR #828, opened 2023-08-13
retry DeleteDisk if disk is still initializing

huww98:
#### What type of PR is this?
/kind cleanup

ECS would refuse to delete the disk if it is just created and is initializing. Although k8s would retry in this case, it could prevent csi-sanity from properly cleaning up the disks created for testing. Also, retrying ourselves would hide this error from end-users.

```release-note
disk driver now retries deleting the disk if it is still initializing
```

```docs

```

### DIFF
--- a/pkg/disk/controllerserver.go
+++ b/pkg/disk/controllerserver.go
@@ -331,9 +331,7 @@ func (cs *controllerServer) DeleteVolume(ctx context.Context, req *csi.DeleteVol
 		}
 	}
 
-	deleteDiskRequest := ecs.CreateDeleteDiskRequest()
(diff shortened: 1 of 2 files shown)


pkg/disk/controllerserver.go
```diff
@@ -274,6 +276,28 @@ func (cs *controllerServer) CreateVolume(ctx context.Context, req *csi.CreateVol
 	return &csi.CreateVolumeResponse{Volume: tmpVol}, nil
 }
 
+func deleteDisk(ecsClient *ecs.Client, diskId string) (*ecs.DeleteDiskResponse, error) {
```
mowangdk, line 279: Move deleteDisk method to cloud.go

PR #835, opened 2023-08-21
disk: update extended resource on startup

huww98:
#### What type of PR is this?
/kind feature

Gives more visibility to users.

Also open the door to more flexible scheduling. In the future, we may dynamically update this, coordinate with DBFS drivers, etc.

https://kubernetes.io/docs/tasks/administer-cluster/extended-resource-node/

terway already advertise a resource `aliyun/member-eni`, which is inconsistent with our `alibabacloud.com/disk`.
But I still want to follow the convention of using a domain name.

```release-note
publish a new resource type `alibabacloud.com/disk`, indicating the total number of disks available for pods. Not used for scheduling yet.
```

```docs

```

### DIFF
--- a/pkg/disk/constants.go
+++ b/pkg/disk/constants.go
@@ -3,6 +3,8 @@ package disk
 import (
 	"fmt"
 	"time"
+
+	k8sv1 "k8s.io/api/core/v1"
 )
 
 const (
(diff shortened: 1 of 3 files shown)


huww98:
After thinking again, extended resource maybe not suitable for this use case. Multiple pods may share the same volume on a node, which cannot be modeled by extended resource. So maybe we should just use an annotation to avoid any abuse that would cause incorrect scheduling.

huww98:
replaced by #844

PR #845, opened 2023-08-28
refactor UpdateNode

huww98:
#### What type of PR is this?

/kind cleanup

* use patch instead of update to avoid conflict and reduce traffic.
* backoff and retry for 1 hour if ECS or node patch failed.
  * the labels are important for the scheduler. Make sure we can recover automatically if something goes wrong.
  * also fix the incorrect retry logic
* add support for non-ACK node, skip query for disk types in this case.

```release-note
NONE
```

```docs

```

### DIFF
--- a/pkg/disk/constants.go
+++ b/pkg/disk/constants.go
@@ -175,10 +175,11 @@ const (
 	// RetryMaxTimes ...
 	RetryMaxTimes = 5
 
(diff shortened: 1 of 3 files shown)


huww98:
As requested by @mowangdk , eventually timeout after 1 hour. These labels are not that important. If no label is present, the scheduler will assume all disk types are supported.

PR #849, opened 2023-08-29
disk/CreateVolume: properly set error code

huww98:
#### What type of PR is this?

/kind cleanup

properly set error code for:
* not existing snapshot
* volume already created with the same name but different parameters

Fixes #

```release-note
NONE
```

```docs

```

### DIFF
--- a/pkg/disk/controllerserver.go
+++ b/pkg/disk/controllerserver.go
@@ -19,6 +19,7 @@ package disk
 import (
 	"bytes"
 	"context"
+	"errors"
 	"fmt"
 	"os"
 	"reflect"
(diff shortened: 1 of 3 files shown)


pkg/disk/controllerserver.go
```diff
@@ -224,6 +226,15 @@ func (cs *controllerServer) CreateVolume(ctx context.Context, req *csi.CreateVol
 
 	diskType, diskID, diskPL, err := createDisk(req.GetName(), snapshotID, requestGB, diskVol, req.Parameters[TenantUserUID])
 	if err != nil {
+		var aliErr *alicloudErr.ServerError
+		if errors.As(err, &aliErr) {
```
tydra-wang, line 230: it will panic because aliErr is nil?

huww98: `&aliErr` is of type `**alicloudErr.ServerError` and `&aliErr` is not nil. Besides, this is already covered by the csi-sanity test cases.

PR #856, opened 2023-09-01
always force delete disk snapshot

huww98:
#### What type of PR is this?

/kind feature

If there are some disks created from a snapshot, that snapshot cannot be deleted without `force`. Because the disks cannot be re-initialized then. We don't support re-init, so just do force delete unconditionally.

Fixes #

```release-note
`forceDelete` snapshot parameter is no longer respected. Now snapshot can always be deleted even if there are volumes created from it.
```

```docs

```


huww98:
As requested by @mowangdk , we still keep the tag `force.delete.snapshot.k8s.aliyun.com` for capability.

PR #863, opened 2023-09-14
disable DISK_FORCE_DETACHED by default

huww98:
#### What type of PR is this?

/kind bug

Currently we will detach disk first when we are attaching this disk to another node. This is very dangerous, and can detach the disk used by a normal pod, causing data damage.

attach_detach_controller in KCM will handle force detach if node is abnormal. So we should not try to detach disk here in any case. The relevant code should also be removed in the future.

We also have this env variable set to "true" in our internal component template. Consider removing it there.

```release-note
DISK_FORCE_DETACHED env variable is now deprecated and disabled by default.
```

```docs

```


huww98:
`detachDisk` can also be called from csi-plugin. But plugin cannot do detach if node is down. Maybe we need to migrate all detaching requests to controller before merging this PR.

/hold

huww98:
resolved conflict

PR #864, opened 2023-09-14
always do detach on controller

huww98:
#### What type of PR is this?

/kind cleanup

There is no risk of detaching wrong disk, so there is no need to do it on node. Also, controller is highly available, but node can be down and not able to execute detaching.

Unblocks: #863 

```release-note
NOTICE: Controller must be upgraded first before any node starts its upgrade.
```

```docs

```


huww98:
Detaching on controller would interference with the disk attach logic. We need fully sequential attach/detach to find which disk is just attached.

PR #824, opened 2023-08-09
Run CSI sanity on disk driver

huww98:
#### What type of PR is this?

/kind cleanup

#### What this PR does / why we need it:

Run [CSI sanity](https://github.com/kubernetes-csi/csi-test) on disk driver and fixing the reported failures.

This is a preview of needed changes, I may break it into smaller PRs for review and merge.

#### Does this PR introduce a user-facing change?

```release-note

```

#### Additional documentation e.g., KEPs (Kubernetes Enhancement Proposals), usage docs, etc.:

```docs

```

PR #876, opened 2023-10-01
disk: fix create volume idempotent

huww98:
#### What type of PR is this?

/kind cleanup

Please see the commit message for detail

CSI sanity case:

controller service [controller server] createvolume [it] should fail when requesting to create a volume with already existing name and different capacity

Because of the ClientToken generation algorithm is changed, there is a extremely small chance that multiple disks will be created for one PVC, if the provisioner created the disk and failed to notify k8s, then the provisioner is upgraded.

This is a part of #824 

```release-note
NONE
```

```docs

```

### DIFF
--- a/pkg/disk/cloud.go
+++ b/pkg/disk/cloud.go
@@ -18,6 +18,7 @@ package disk
 
 import (
 	"context"
+	"crypto/sha256"
 	"errors"
 	"fmt"
 	"regexp"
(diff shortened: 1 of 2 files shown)


huww98:
ECS is treating ClientToken differently when returning different error code. Hard-coding each of these behaviors is impractical and fragile. I will find another way.

PR #878, opened 2023-10-03
store disk type and performance level as slice

huww98:
#### What type of PR is this?

/kind cleanup

Avoid repeated split and join.

The original `validateDiskPerformaceLevel` does not validate the input if it does not contains comma. Fixed this.

```release-note
NONE
```

```docs

```

### DIFF
--- a/pkg/disk/utils.go
+++ b/pkg/disk/utils.go
@@ -874,7 +874,7 @@ func getDiskVolumeOptions(req *csi.CreateVolumeRequest) (*diskVolumeArgs, error)
 	// disk Type
 	diskType, err := validateDiskType(volOptions)
 	if err != nil {
-		return nil, fmt.Errorf("Illegal required parameter type: " + diskVolArgs.Type)
+		return nil, fmt.Errorf("Illegal required parameter type: " + volOptions["type"])
 	}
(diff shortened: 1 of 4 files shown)


pkg/disk/utils.go
```diff
@@ -1036,33 +1036,32 @@ func validateDiskType(opts map[string]string) (diskType string, err error) {
...
 	log.Infof("validateDiskPerformanceLevel: pl: %v", pl)
-	if strings.Contains(pl, ",") {
-		for _, cusPer := range strings.Split(pl, ",") {
-			if _, ok := CustomDiskPerfermance[cusPer]; !ok {
-				return "", fmt.Errorf("illegal performance level type: %s", cusPer)
-			}
+	performanceLevel = strings.Split(pl, ",")
+	for _, cusPer := range performanceLevel {
+		if _, ok := CustomDiskPerfermance[cusPer]; !ok {
+			return nil, fmt.Errorf("illegal performance level type: %s", cusPer)
```
mowangdk, line 1061: return default value ```[]string{}``` instead of nil?

huww98: The default value of []string is nil.

PR #879, opened 2023-10-03
Centralized Node Metadata Provider

huww98:
#### What type of PR is this?

/kind cleanup

We are querying for node metadata everywhere in the code, from multiple sources. But existing implementation is inconsistent, not robust to unavailable sources. And there are many duplicated codes in different drivers.

This PR implements a centralized metadata provider, which will fetch node metadata lazily from whatever source that is available, and cache the result. All existing sources used by disk driver is included: env variables, ECS metadata server, k8s API server, ECS OpenAPI.

This PR also integrates the new infrastructure into the disk driver, replacing all the node metadata retrieval code there.

See the commit message of first commit for the overall design.

```release-note
NONE
```

```docs

```


huww98:
Changed instance metadata endpoint to "latest". I've checked the docs, which do not mention the version. Also checked AWS EC2 SDK, it also uses "latest".

huww98:
As requested by @mowangdk , dropped the two commits about runtime

PR #882, opened 2023-10-09
oss: support CoCo direct volume feature

mozillazg:
#### What type of PR is this?

/kind feature

oss: support CoCo direct volume feature

Fixes #

https://github.com/kata-containers/kata-containers/blob/main/docs/design/direct-blk-device-assignment.md
https://github.com/kata-containers/kata-containers/issues/7943
https://github.com/kata-containers/kata-containers/pull/7965
https://github.com/confidential-containers/guest-components/pull/345/

```release-note

```

```docs

```

### DIFF
--- a/pkg/oss/coco.go
+++ b/pkg/oss/coco.go
@@ -1,0 +1,137 @@
+package oss
+
+import (
+	"context"
+	"encoding/json"
+	"fmt"
+	"strings"
+
+	"github.com/container-storage-interface/spec/lib/go/csi"
+	"github.com/kubernetes-sigs/alibaba-cloud-csi-driver/pkg/utils/kata/directvolume"
+	log "github.com/sirupsen/logrus"
+)
+
+const (
+	optDirectAssigned  = "direct"
+	optAnnotations     = "annotations"
+	optEncrypted       = "encrypted"
+	optEncPasswd       = "encPasswd"
+	optKmsKeyId        = "kmsKeyId"
+	fsTypeSecureMount  = "secure_mount"
+	volumeTypeOSS      = "alibaba-cloud-oss"
+	sealedSecretPrefix = "sealed."
+)
+
+func (ns *nodeServer) publishDirectVolume(ctx context.Context, req *csi.NodePublishVolumeRequest, opt *Options) (*csi.NodePublishVolumeResponse, error) {
+	logger := log.WithFields(map[string]interface{}{
+		"NodeServer": "OSSNodeServer",
+		"VolumeId":   req.VolumeId,
+		"TargetPath": req.TargetPath,
+	})
+
+	volumePath := req.TargetPath
+	if isDirectVolumePath(volumePath) {
+		logger.Infof("NodePublishVolume: The mount info for DirectVolume is already exist: %s", volumePath)
+		return &csi.NodePublishVolumeResponse{}, nil
+	}
+
+	device := req.TargetPath
+	volumeType := volumeTypeOSS
+	fsType := fsTypeSecureMount
+	var encrypted, encPasswd, kmsKeyId string
+	annotationsObj := map[string]string{}
+
+	if v := req.Secrets[optEncPasswd]; v != "" {
+		encPasswd = v
+	}
+	for key, value := range req.VolumeContext {
+		switch key {
+		case optEncrypted:
+			encrypted = strings.TrimSpace(value)
+		case optEncPasswd:
+			encPasswd = strings.TrimSpace(value)
+		case optKmsKeyId:
+			kmsKeyId = strings.TrimSpace(value)
+		case optAnnotations:
+			json.Unmarshal([]byte(value), &annotationsObj)
+		}
+	}
+
+	if annotationsObj["kata_fs_type"] != "" {
+		fsType = annotationsObj["kata_fs_type"]
+	}
+	if annotationsObj["kata_volume_type"] != "" {
+		volumeType = annotationsObj["kata_volume_type"]
+	}
+	if annotationsObj["kata_device"] != "" {
+		device = annotationsObj["kata_device"]
+	}
+	annotations, _ := json.Marshal(annotationsObj)
+	annotationsStr := string(annotations)
+	if len(annotations) == 0 || annotationsStr == "null" || annotationsStr == "" {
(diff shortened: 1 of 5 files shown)


pkg/utils/kata/directvolume/direct_volume.go
```diff
@@ -0,0 +1,83 @@
...
+	if err := json.Unmarshal([]byte(mountInfo), &deserialized); err != nil {
+		return err
+	}
+
+	return os.WriteFile(filepath.Join(volumeDir, mountInfoFileName), []byte(mountInfo), 0600)
+}
+
+// Remove deletes the direct volume path including all the files inside it.
+func Remove(volumePath string) error {
+	return os.RemoveAll(filepath.Join(kataDirectVolumeRootPath, b64.URLEncoding.EncodeToString([]byte(volumePath))))
```
mowangdk, line 65: Remove each file using ```os.Remove```? It's safer.

pkg/utils/kata/directvolume/direct_volume.go
```diff
@@ -0,0 +1,83 @@
...
+	// The filesystem type to be mounted on the volume.
+	FsType string `json:"fstype"`
+	// Additional metadata to pass to the agent regarding this volume.
+	Metadata map[string]string `json:"metadata,omitempty"`
+	// Additional mount options.
+	Options []string `json:"options,omitempty"`
+}
+
+// Add writes the mount info of a direct volume into a filesystem path known to Kata Container.
+func Add(volumePath string, mountInfo string) error {
```
huww98, line 40: pass mountInfo as `[]byte`?

`Add` is only called from `AddMountInfo`. Combine them to one func?

pkg/utils/kata/directvolume/direct_volume.go
```diff
@@ -0,0 +1,83 @@
...
+		if err := os.MkdirAll(volumeDir, 0700); err != nil {
+			return err
+		}
+	}
+	if stat != nil && !stat.IsDir() {
+		return fmt.Errorf("%s should be a directory", volumeDir)
+	}
+
+	var deserialized MountInfo
+	if err := json.Unmarshal([]byte(mountInfo), &deserialized); err != nil {
```
huww98, line 56: Why unmarshal it?

huww98 (changes requested):

pkg/utils/kata/directvolume/direct_volume.go
```diff
@@ -0,0 +1,83 @@
...
+			return err
+		}
+		if err := os.MkdirAll(volumeDir, 0700); err != nil {
+			return err
+		}
+	}
+	if stat != nil && !stat.IsDir() {
+		return fmt.Errorf("%s should be a directory", volumeDir)
+	}
+
```
huww98, line 54: How about just call `os.MkdirAll`? It will do the `Stat` and the checks for you.
```suggestion
	if err := os.MkdirAll(volumeDir, 0700); err != nil {
		return err
	}
```

pkg/utils/kata/directvolume/direct_volume.go
mozillazg: This file is copied from https://github.com/kata-containers/kata-containers/blob/c800d0739fe42414a15ac8967f3fa5ccb98c28d7/src/runtime/pkg/direct-volume/utils.go.
We should keep it sync with upstream without any changes.

pkg/oss/coco.go
```diff
@@ -0,0 +1,137 @@
...
+	for key, value := range req.VolumeContext {
+		switch key {
+		case optEncrypted:
+			encrypted = strings.TrimSpace(value)
+		case optEncPasswd:
+			encPasswd = strings.TrimSpace(value)
+		case optKmsKeyId:
+			kmsKeyId = strings.TrimSpace(value)
+		case optAnnotations:
+			json.Unmarshal([]byte(value), &annotationsObj)
```
huww98, line 56: Is it intentional to ignore the errors here?

pkg/oss/coco.go
```diff
@@ -0,0 +1,137 @@
...
+	if encPasswd != "" && strings.HasPrefix(encPasswd, sealedSecretPrefix) {
+		metadata[optEncPasswd] = encPasswd
+	}
+
+	mountInfo := directvolume.MountInfo{
+		VolumeType: volumeType,
+		Device:     device,
+		FsType:     fsType,
+		Metadata:   metadata,
+		Options:    strings.Fields(opt.OtherOpts),
```
huww98, line 99: by using `strings.Fields()`, we dropped the support for space in the options, e.g.: `-o key="hello world"` will not work.

pkg/utils/kata/directvolume/direct_volume.go
huww98: OK. then how about just importing it? That is a one file package.

mozillazg: We can't just import it. That will cause we depend on the whole kata runtime project.

pkg/oss/coco.go
mozillazg: I don't think the ossfs have such cases.

AlbeeSo:
will the kata containers follow the ossfs version in CSI ? ACK ossfs 1.80.1 is the min version support encrytion.

mowangdk:
/ok-to-test

pkg/oss/coco.go
huww98: Maybe. But this breaks the current interface. We are parsing the options with shell now, e.g., `-o 'key'` is parsed as `["-o", "key"]`, not `["-o", "'key'"]`

pkg/utils/kata/directvolume/direct_volume.go
huww98: I think that is not a big deal. We are using vendor mode, which will only introduce one go file into the vendor directory. But I've tried that, this requires several upgrades to other dependencies. So, it is a pity. Maybe we can import this after we sort out our go.mod, and upgrade our dependencies.

pkg/utils/kata/directvolume/direct_volume.go
mowangdk: +1

PR #887, opened 2023-10-12
feat: visit storage-monitor using clusterIP instead of domain name

tydra-wang:
#### What type of PR is this?

/kind cleanup

Fixes #

```release-note

```

```docs

```


pkg/metric/monitor_client.go
```diff
@@ -0,0 +1,99 @@
...
+	c.RLock()
+	clusterIP := c.clusterIP
+	c.RUnlock()
+	if clusterIP == "" {
+		return nil, errors.New("clusterIP not initialized")
+	}
+	params := url.Values{"multi-cnfs-nas": []string{pv}}
+	uri := url.URL{
+		Scheme:   "http",
+		Host:     net.JoinHostPort(clusterIP, "11280"),
```
mowangdk, line 59: Move port value to const.

pkg/metric/monitor_client.go
```diff
@@ -0,0 +1,99 @@
...
+		Path:     "/metrics",
+		RawQuery: params.Encode(),
+	}
+	resp, err := (&http.Client{Timeout: time.Millisecond * 500}).Get(uri.String())
```
mowangdk, line 63: same as above

huww98:
I would implement this like DNS. Save IP and lastUpdateTime in StorageMonitorClient, and synchronously update the IP if it is too old. This way, we can have a different TTL to "not found" result. Also, we don't pay for this if no one is requesting the metrics.

mowangdk:
> I would implement this like DNS. Save IP and lastUpdateTime in StorageMonitorClient, and synchronously update the IP if it is too old. This way, we can have a different TTL to "not found" result. Also, we don't pay for this if no one is requesting the metrics.

Never mind

PR #890, opened 2023-10-17
feature: support reschedule disk with resource exhausted error in del…

mowangdk:
#### What type of PR is this?
/kind feature

Support reschedule disk with resource exhausted error in delay provisioning

None
None

```release-note
None
```

```docs
None
```

### DIFF
--- a/pkg/disk/cloud.go
+++ b/pkg/disk/cloud.go
@@ -969,7 +969,7 @@ func getDiskType(diskVol *diskVolumeArgs) ([]string, []string, error) {
 		nodeInfo, err := client.CoreV1().Nodes().Get(context.Background(), diskVol.NodeSelected, metav1.GetOptions{})
 		if err != nil {
 			log.Log.Infof("getDiskType: failed to get node labels: %v", err)
-			goto cusDiskType
+			return nil, nil, status.Errorf(codes.ResourceExhausted, "CreateVolume:: get node info by name: %s failed with err: %v, start to reschedule", diskVol.NodeSelected, err)
 		}
 		re := regexp.MustCompile(`node.csi.alibabacloud.com/disktype.(.*)`)
 		for key := range nodeInfo.Labels {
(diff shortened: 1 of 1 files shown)


pkg/disk/cloud.go
```diff
@@ -980,14 +980,13 @@ func getDiskType(diskVol *diskVolumeArgs) ([]string, []string, error) {
...
 
-cusDiskType:
 	provisionDiskTypes := []string{}
 	allTypes := deleteEmpty(strings.Split(diskVol.Type, ","))
 	if len(nodeSupportDiskType) != 0 {
 		provisionDiskTypes = intersect(nodeSupportDiskType, allTypes)
 		if len(provisionDiskTypes) == 0 {
 			log.Log.Errorf("CreateVolume:: node(%s) support type: [%v] is incompatible with provision disk type: [%s]", diskVol.NodeSelected, nodeSupportDiskType, allTypes)
-			return nil, nil, status.Errorf(codes.InvalidArgument, "CreateVolume:: node support type: [%v] is incompatible with provision disk type: [%s]", nodeSupportDiskType, allTypes)
+			return nil, nil, status.Errorf(codes.ResourceExhausted, "CreateVolume:: node support type: [%v] is incompatible with provision disk type: [%s]", nodeSupportDiskType, allTypes)
```
huww98, line 989: We can only reach this line if the CSI plugin is just installed, and scheduling is performed before `UpdateNode()` finished. Is that correct?

If the above is true, we can eliminate this kind of racing by using topology to constraint disk type.

huww98 (commented):
FYI, I have a private branch that uses topology to constraint disk type, which is waiting for the dependencies to be merged. By that way, we can avoid query for "node-selected" from APIServer.

Once I open that as a PR, it will overwrite the changes here.

mowangdk:
> FYI, I have a private branch that uses topology to constraint disk type, which is waiting for the dependencies to be merged. By that way, we can avoid query for "node-selected" from APIServer.
> 
> Once I open that as a PR, it will overwrite the changes here.

It's okay. We'll go over your changes when you file your pr.

pkg/disk/cloud.go
```diff
@@ -969,7 +969,7 @@ func getDiskType(diskVol *diskVolumeArgs) ([]string, []string, error) {
 		nodeInfo, err := client.CoreV1().Nodes().Get(context.Background(), diskVol.NodeSelected, metav1.GetOptions{})
 		if err != nil {
 			log.Log.Infof("getDiskType: failed to get node labels: %v", err)
-			goto cusDiskType
+			return nil, nil, status.Errorf(codes.ResourceExhausted, "CreateVolume:: get node info by name: %s failed with err: %v, start to reschedule", diskVol.NodeSelected, err)
```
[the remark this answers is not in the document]
mowangdk: The retry time follows the backoff algorithm and has little effect on CSI. But it still has side effect on the scheduler, so I'll modify it.

pkg/disk/cloud.go
mowangdk: No, there are other scenarios. We'll leave it as it is for now.

PR #899, opened 2023-10-24
disk: remove ADController config

huww98:
#### What type of PR is this?

/kind cleanup
/kind api-change

Drop all code paths about attach/detach disks on Node.
Avoid OpenAPI authorization on Node.

Fixes #

Blocked by stage-1 (just disable it by default, #903)

```release-note
env DISK_AD_CONTROLLER and config disk-adcontroller-enable is now removed, we now always do attach/detach on controller
```

```docs

```


huww98:
/hold

We should merge this after #903, and verify it works OK with most users.

huww98:
The latest version of this PR is included in #824 . Will cherry-pick it here when it is mature enough.

PR #910, opened 2023-11-06
disk: fixes for unmountDuplicateMountPoint

huww98:
#### What type of PR is this?

/kind bug

There are several issues about the original implementation:
1. We should not just umount the mountpoints in data disk, as it will propagate to `/var/lib/kubelet`. Instead, we handle this in entrypoint.sh, by mount data disk as private then umount all mountpoint in it.
2. We should umount the 1.23- path even if no data disk present.
3. We will not mount v2 path in NodeStageVolume if v1 path is present, maybe trying to prevent two mounts to exists simultaneously. However, this does not work. We see a `FailedMount` event from kubelet, which may be related. Also, If we mount the same PV again on the same node, we will also mount the v2 path in NodePublishVolume, resulting in two mounts existing simultaneously. And we also have users upgraded to Kubernetes 1.24 before he upgrades CSI plugin to 758da746493eadb8e6bc317c1cb05869cc37f255 .

In summary, the proposed solution is:
1. Fixing mountpoints in data disk once and for all in entrypoint.sh
1. In NodeStageVolume, mount the target path as requested by CSI, either v1 or v2, allowing them to exist simultaneously.
1. Cleanup v1 if v2 is also present in NodeUnpublishVolume

Test plan:

1. Create a 1.22 ACK cluster, with 2 nodes. One with data disk, one without.
2. Upgrade the CSI plugin to this PR. Check that the mountpoints in data disk are cleaned up, and the mountpoints in `/var/lib/kubelet` are intact.
3. Create one pod with disk on each node. Check that the v1 mount created.
4. Delete the pods, Check that all mount points are cleaned up.
5. Repeat 3.
6. Upgrade the cluster to 1.24. Check that v1 and v2 mounts exist simultaneously.
7. Delete the pods, check that all mount points are cleaned up.
8. Repeat 3. Check that only v2 mount created.
9. Delete the pods, check that all mount points are cleaned up.

```release-note
fix a bug that volume cannot be unmounted after upgrade to Kubernetes 1.24, and no data disk is present on node.
```

```docs

```

### DIFF
--- a/build/lib/amd64-entrypoint.sh
+++ b/build/lib/amd64-entrypoint.sh
@@ -37,6 +37,22 @@ if [[ "$os_release_exist" = "0" ]]; then
 fi
 echo "detected host os: $host_os"
 
+# Host may have /var/lib/kubelet bind mounted to /var/lib/container/kubelet in shared mode,
+# where /var/lib/container is the data disk configured in node pool.
+# Fix this by making /var/lib/container private and unmounting the duplicate mount in it.
+if [ /var/lib/kubelet -ef /var/lib/container/kubelet ] && awk '{print $5}' /proc/1/mountinfo | grep -q "^/var/lib/container$"; then
+    # They refer to the same device and inode numbers and /var/lib/container is a mountpoint
+    echo "fixing mountpoints in /var/lib/container"
+    # Note that we make /var/lib/container private, but not /var/lib/kubelet,
+    # because the latter is also shared with CSI plugins.
+    $HOST_CMD mount --make-private /var/lib/container || exit 1
+    DUPICATE_MNTS=$(awk '{print $5}' /proc/1/mountinfo | grep "^/var/lib/container/kubelet")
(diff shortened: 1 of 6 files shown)


huww98:
Test finished.

For test plan step 6, I'm not able to upgrade the cluster with custom CSI version installed. So instead, I downgrade to public version CSI, upgrade the cluster, re-upgrade to CSI version in this PR, then restart kubelet.

Note that we should ensure the pod template is upgraded with the image. Or we will get error from kubelet when rebuilding the csi-plugin pod: `Error: failed to generate container "40eb796f653c58f77849a4a0cb9383ab021ae51970a1f0b1d68b9bc704e254d1" spec: failed to generate spec: path "/var/lib/container" is mounted on "/var/lib/container" but it is not a shared mount`. I've shifted the changes to template from commit 3 to commit 2.

Apart from this, all works as expected. @mowangdk PTAL.

huww98:
Superseded by #911 .
I will open another PR for the part that is in this PR but not in #911

PR #911, opened 2023-11-07
disk: fix unmountDuplicateMountPoint to handle no data disk node

huww98:
#### What type of PR is this?

/kind bug

We should unmount the 1.23- path even if no data disk present. Or kubelet will not unstage the volume.

#910 is not suitable for now, because we have no reliable way to remove volumeMount from csi-plugin DaemonSet in ACK.

Mount points in data disk are still always unmounted, because they also prevent kubelet from unstage the volume before 1.26 (See https://github.com/kubernetes/kubernetes/pull/112607 for the behaviour in 1.26+). Although this is incorrect: the unmount will propagate to `/var/lib/kubelet` and the real stage path is also unmounted. But we are already fixing this in NodePublishVolume. So, I left it as it is for now.

We do not mount v2 path in NodeStageVolume if v1 path is present, maybe trying to prevent two mounts to exists simultaneously. However, this does not work. If user mount the same PV again on the same node, we will also mount the v2 path in NodePublishVolume, resulting in two mounts existing simultaneously. And we also have users upgraded to Kubernetes 1.24 before he upgrades CSI plugin to https://github.com/kubernetes-sigs/alibaba-cloud-csi-driver/commit/758da746493eadb8e6bc317c1cb05869cc37f255 .
So, revert that change.

See #910 for the test plan.

```release-note
fix the bug that volume cannot be unmounted after upgrade to Kubernetes 1.24, and no data disk is present on node.
```

```docs

```

### DIFF
--- a/pkg/disk/nodeserver.go
+++ b/pkg/disk/nodeserver.go
@@ -514,21 +514,26 @@ func (ns *nodeServer) NodeStageVolume(ctx context.Context, req *csi.NodeStageVol
 		}
 	}
 
-	// Step 1: check oldVersion volumeMount
-	oldTargetPath := fmt.Sprintf("/var/lib/kubelet/plugins/kubernetes.io/csi/pv/%s/globalmount", req.VolumeId)
-	returned, err := ns.checkStagingPathMounted(req.VolumeId, oldTargetPath, targetPath)
-	if err != nil {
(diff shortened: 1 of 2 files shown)


huww98:
Manual tested according to plan in #910

PR #874, opened 2023-09-30
cleanup the go.mod

huww98:
#### What type of PR is this?

/kind cleanup

#### What this PR does / why we need it:

This PR brings the dependencies to a consistent state (remove all the `replace` in go.mod).

#### Which issue(s) this PR fixes:

Fixes #

#### Special notes for your reviewer:

I've grouped the changes into multiple commits according to the risk introduced: the first 2 commits should have relatively low risk, and may be merged first if you agree.

#### Does this PR introduce a user-facing change?

```release-note
NONE
```

#### Additional documentation e.g., KEPs (Kubernetes Enhancement Proposals), usage docs, etc.:

```docs

```

PR #914, opened 2023-11-22
Refactor nas plugin and support accesspoint

tydra-wang:
#### What type of PR is this?

/kind feature
/kind cleanup

Fixes #

```release-note

```

```docs

```


huww98:
Maybe you can use #873 to implement v2 Auth

huww98:
Please see #874 for upgrade deps.

pkg/nas/cloud/nas_client_v2.go
```diff
@@ -0,0 +1,175 @@
...
+
+	openapi "github.com/alibabacloud-go/darabonba-openapi/v2/client"
+	sdk "github.com/alibabacloud-go/nas-20170626/v3/client"
+	"github.com/alibabacloud-go/tea/tea"
+	"github.com/kubernetes-sigs/alibaba-cloud-csi-driver/pkg/utils"
+	"github.com/sirupsen/logrus"
+	"go.uber.org/ratelimit"
+)
+
+func newNasClientV2(region string) (*sdk.Client, error) {
```
mowangdk, line 17: Please add some comments indicates the difference between v1 & v2

pkg/nas/filesystem_controller.go
```diff
@@ -0,0 +1,529 @@
...
+	AddDefaultTagsError string = "AddDefaultTagsError"
+	// LosetupType tag
+	LosetupType = "losetup"
+	// SkipMountType tag
+	SkipMountType = "skipmount"
+
+	csiAlibabaCloudName = "csi.alibabacloud.com"
+)
+
+var pvcMountTargetMap = map[string]string{}
```
mowangdk, line 79: these three maps may has concurrency race issues?

pkg/nas/internal/config.go
```diff
@@ -0,0 +1,163 @@
...
+		}
+		if config.NodeIP == "" {
+			return nil, errors.New("enabled losetup mode but failed to get node IP")
+		}
+	}
+
+	return config, nil
+}
+
+func parseBool(str string) (bool, error) {
```
mowangdk, line 155: Can we use `strconv.ParsePool()` instead?

pkg/utils/auth.go
```diff
@@ -406,3 +419,104 @@ func getCredentialAK() AccessControl {
...
+	}
+}
+
+func (cred *managedAddonTokenCredv2) needUpdate() bool {
+	if cred.token == nil {
+		return true
+	}
+	duration := time.Since(cred.lastUpdateAt)
+	expiration := cred.token.ExpireAt.Sub(cred.lastUpdateAt)
+	return duration >= time.Duration(float64(expiration)*cred.scale)
```
huww98, line 442: I think my version at https://github.com/kubernetes-sigs/alibaba-cloud-csi-driver/pull/873/files#diff-7292cc76b3b0f97dd0d6676f1d72baa164adebc25b265417dd5dfc724236e780R152-R159 is better.

Since this is a local file, reloading has relatively low cost, so I just reload every 5 minutes, in case of user re-created the RAM role so that the token is invalidated earlier than expected. And I add a fixed 10 minutes offset to the expiration time, which I think is more reasonable than scale. Because this offset is used to compensate for clock drift, and this drift should not relate to the token valid duration.

Or are you willing to use my version of v2 cred in #873 ? I (or you) can cherry-pick the relevant commits.

huww98:
Please rebase and get rid of the commit `Revert "remove staging nas sdk"`, maybe combine it with `remove staging nas sdk`

pkg/nas/accesspoint_controller.go
```diff
@@ -0,0 +1,263 @@
...
+	}
+	resp := &csi.CreateVolumeResponse{
+		Volume: &csi.Volume{
+			VolumeId:      req.Name,
+			VolumeContext: volumeContext,
+		},
+	}
+
+	// set dir quota
+	if parameters["volumeCapacity"] == "true" || parameters["allowVolumeExpansion"] == "true" {
```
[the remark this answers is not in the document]
tydra-wang: I don't think we can directly remove it. A lot of users already got used to it. I will update the alicloud documentation and remove it later.

pkg/nas/internal/config.go
tydra-wang: No because it can not recognize "enabled" or "yes" which are legal values in our doc.

pkg/utils/auth.go
tydra-wang: Good! I will take a look.

pkg/nas/accesspoint_controller.go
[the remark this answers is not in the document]
mowangdk: good

pkg/nas/filesystem_controller.go
tydra-wang: Changed to `sync.Map` and plan to get rid of them by using clientToken to keep idempotence in the future.

pkg/nas/accesspoint_controller.go
```diff
@@ -0,0 +1,268 @@
...
+		VpcId:           &vpcId,
+		VswId:           &vswId,
+		AccessPointName: &name,
+		RootDirectory:   tea.String(path.Join(basePath, name)),
+		OwnerUserId:     tea.Int32(0),
+		OwnerGroupId:    tea.Int32(0),
+	}
+
+	var invalidParameters []string
+	parseParameter := func(key string, target any) {
```
mowangdk, line 137: elegant 👍

huww98: I would prefer to just switch on key. So that we can report errors on unknown keys. And we can get a compilation error when unknown value type is encountered (maybe due to future SDK update).

pkg/nas/accesspoint_controller.go
```diff
@@ -0,0 +1,268 @@
...
+		}
+	}
+	parseParameter("accessGroupName", &req.AccessGroup)
+	parseParameter("accesspointEnableRam", &req.EnabledRam)
+	parseParameter("accesspointOwnerGroupId", &req.OwnerGroupId)
+	parseParameter("accesspointOwnerUserId", &req.OwnerUserId)
+	parseParameter("accesspointPermission", &req.Permission)
+	parseParameter("accesspointPosixUserId", &req.PosixUserId)
+	parseParameter("accesspointPosixGroupId", &req.PosixGroupId)
+	parseParameter("accesspointPosixSecondaryGroupIds", &req.PosixSecondaryGroupIds)
```
huww98, line 169: BTW, what is the expected format of this parameter? This looks strange, I would expect the type to be `[]int32`, but it is *string.

pkg/nas/cloud/nas_client_v2.go
```diff
@@ -0,0 +1,183 @@
...
+		AccessPointId: &accessPointId,
+		FileSystemId:  &filesystemId,
+	})
+}
+
+func IsAccessPointNotFoundError(err error) bool {
+	if err == nil {
+		return false
+	}
+	sdkErr, ok := err.(*tea.SDKError)
```
huww98, line 181: prefer errors.As()

pkg/nas/internal/config.go
```diff
@@ -0,0 +1,163 @@
...
+		CNFSGetter:      cnfsGetter,
+	}
+
+	// check if enable nfs port check
+	if value := os.Getenv("NAS_PORT_CHECK"); value != "" {
+		config.EnablePortCheck, _ = parseBool(value)
+	}
+
+	// get csi-plugin configmap
+	cm, err := kubeClient.CoreV1().ConfigMaps(configMapNamespace).Get(context.Background(), configMapName, metav1.GetOptions{})
```
huww98, line 102: We can just read `/etc/csi-plugin/config` on node

PR #921, opened 2023-12-04
cleanup unmountDuplicateMountPoint

huww98:
#### What type of PR is this?

/kind cleanup

When we raise the minimum supported k8s version above 1.26, we can drop the hack of unmount old staging path from NodeUnpublishVolume. Instead, we unmount all the old path at startup, since kubelet will not need it anymore. And duplicated mount point in data disk would not block unstaging volume.

Fixes #

```release-note
NONE
```

```docs

```


huww98:
/ok-to-test

Part of this PR is already merged in #1011

PR #924, opened 2023-12-08
fix: configure mlocate cron.daily

AlbeeSo:
#### What type of PR is this?
/kind bug

Avoid daily updatedb scan on kublet-rootdir triggered by mlocate.

Fixes #

```release-note

```

```docs

```

### DIFF
--- a/build/lib/amd64-entrypoint.sh
+++ b/build/lib/amd64-entrypoint.sh
@@ -83,10 +83,8 @@ if [ -z "$KUBELET_ROOT_DIR" ]; then
 fi
 if [ "$SKIP_UPDATEDB_CONFIG" != "true" ]; then
     ## check cron.daily dir
-    stat /host/etc/cron.daily/mlocate >> /dev/null
-    if [ $? -eq 0 ]; then
-        stat /host/etc/updatedb.conf >> /dev/null
-        if [ $? -eq 0 ]; then
+    if [ -f /host/etc/cron.daily/mlocate ]; then
+        if [ -f /host/etc/updatedb.conf ]; then
             PRUNEFS=`sed '/^PRUNEFS =/!d; s/.*= //' /host/etc/updatedb.conf | sed 's/\"//g'`
             PRUNEPATHS=`sed '/^PRUNEPATHS =/!d; s/.*= //' /host/etc/updatedb.conf | sed 's/\"//g'`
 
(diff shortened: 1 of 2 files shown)


build/lib/amd64-entrypoint.sh
```diff
@@ -77,6 +77,39 @@ do
...
+                echo "original PRUNEPATHS=${PRUNEPATHS}"
+                if [ "$KUBELET_ROOT_DIR" == "/var/lib/kubelet" ]; then
+                    PRUNEPATHS="\"${PRUNEPATHS} /var/lib/kubelet /var/lib/container/kubelet\""
+                    echo "add PRUNEPATHS:/var/lib/kubelet /var/lib/container/kubelet in /etc/updatedb.conf"
+                else
+                    PRUNEPATHS="\"${PRUNEPATHS} ${KUBELET_ROOT_DIR}\""
+                    echo "add PRUNEPATHS:${KUBELET_ROOT_DIR} in /etc/updatedb.conf"
+                fi
+                sed -i 's#PRUNEPATHS.*$#PRUNEPATHS = '"${PRUNEPATHS}"'#g' /host/etc/updatedb.conf
+            fi
```
huww98, line 108:
```suggestion
            if ! grep "^PRUNEPATHS" /host/etc/updatedb.conf | grep -q --fixed-strings "$KUBELET_ROOT_DIR"; then
                PRUNEPATHS="$KUBELET_ROOT_DIR"
                if [ "$KUBELET_ROOT_DIR" = "/var/lib/kubelet" ]; then
                    PRUNEPATHS="$PRUNEPATHS /var/lib/container/kubelet"
                fi
                echo "add PRUNEPATHS: $PRUNEPATHS to /etc/updatedb.conf"
                sed -i "s|PRUNEPATHS\s*=\s*\"|&${PRUNEPATHS//|/\\|} |" /host/etc/updatedb.conf
            fi
```
Multiple space may around `=`. Use `--fixed-strings` to avoid injection. Avoid extra `wc` by reading `grep` exit code. Properly escape `PRUNEPATHS` to avoid injection. Avoid getting the original setting, just prepending our path is easier.

mowangdk: lgtm, @AlbeeSo ptal

AlbeeSo:
> Do we need `PRUNEFS` in addition to `PRUNEPATHS`?

i think so, some users mount ossfs on host directly, then used by pods with hostpath volume, in order to share the mountpoint with processes barely run on host.

huww98 (approved):

PR #927, opened 2023-12-14
feat: containerized ossfs auth enhanced

AlbeeSo:
#### What type of PR is this?
/kind feature

support RRSA & csi-secret-store authType for containerized ossfs.

Fixes #

additional rbac: serviceaccount/csi-fuse-ossfs get 

```release-note

```

```docs

```


pkg/mounter/ossfs.go
```diff
@@ -246,37 +254,111 @@ func CleanupOssfsCredentialSecret(ctx context.Context, clientset kubernetes.Inte
...
+		if rrsaCfg == nil {
+			return
+		}
+		spec.ServiceAccountName = authCfg.ServiceAccountName
+		rrsaMountDir := "/var/run/secrets/ack.alibabacloud.com/rrsa-tokens"
+		rrsaVolume := corev1.Volume{
+			Name: "rrsa-oidc-token",
+			VolumeSource: corev1.VolumeSource{
+				Projected: &corev1.ProjectedVolumeSource{
+					DefaultMode: tea.Int32(420),
```
huww98, line 278:
```suggestion
					DefaultMode: tea.Int32(0600),
```
Should this be more obvious?

pkg/mounter/ossfs.go
```diff
@@ -246,37 +254,111 @@ func CleanupOssfsCredentialSecret(ctx context.Context, clientset kubernetes.Inte
...
 			},
-		},
-	}
-	spec.Volumes = append(spec.Volumes, passwdSecretVolume)
-	passwdVolumeMont := corev1.VolumeMount{
-		Name:      passwdSecretVolume.Name,
-		MountPath: passwdMountDir,
+		}
+		spec.Volumes = append(spec.Volumes, rrsaVolume)
+		rrsaVolumeMont := corev1.VolumeMount{
```
huww98, line 292:
```suggestion
		rrsaVolumeMount := corev1.VolumeMount{
```

pkg/mounter/fuse_containerized_mounter.go
```diff
@@ -201,6 +217,22 @@ func (mounter *ContainerizedFuseMounter) labelsAndListOptionsFor(target string)
 	return labels, listOptions
 }
 
+func (mounter *ContainerizedFuseMounter) ensureSeriveAccount(ctx context.Context) error {
```
huww98, line 220: Do we really want this? Would this requires additional RBAC permission? maybe we should just add a service account into our installation yaml.

pkg/oss/controllerserver.go
```diff
@@ -61,6 +61,10 @@ func getOssVolumeOptions(req *csi.CreateVolumeRequest) *Options {
 			ossVolArgs.UseSharedPath = true
 		} else if key == "authtype" {
 			ossVolArgs.AuthType = value
+		} else if key == "rolename" {
+			ossVolArgs.RoleName = value
```
huww98, line 65: The user is responsible for creating this RAM role and authorizing the desired OSS permission. right?

pkg/oss/nodeserver.go
```diff
@@ -247,13 +255,28 @@ func (ns *nodeServer) NodePublishVolume(ctx context.Context, req *csi.NodePublis
...
 	case mounter.AuthTypeSTS:
 		if opt.FuseType == OssFsType {
 			mountOptions = append(mountOptions, GetRAMRoleOption())
 		} else if opt.FuseType == JindoFsType {
 			mountOptions = append(mountOptions, "fs.oss.provider.endpoint=ECS_ROLE")
 		}
+	case mounter.AuthTypeRRSA:
+		if opt.FuseType == OssFsType {
+			if regionID == "" {
+				mountOptions = append(mountOptions, fmt.Sprintf("rrsa_endpoint=https://sts.aliyuncs.com"))
```
huww98, line 269:
```suggestion
				mountOptions = append(mountOptions, "rrsa_endpoint=https://sts.aliyuncs.com")
```

main.go
```diff
@@ -119,6 +119,8 @@ var (
 	nodeID          = flag.String("nodeid", "", "node id")
 	runAsController = flag.Bool("run-as-controller", false, "Only run as controller service")
 	driver          = flag.String("driver", TypePluginDISK, "CSI Driver")
+	clusterID       = flag.String("clusterid", "", "cluster id")
+	aliUID          = flag.String("aliuid", "", "aliyun user id")
```
huww98, line 123: Should we use env? Env seems to be easier for kustomize.
aliUID is also available at `http://100.100.100.200/latest/meta-data/owner-account-id` or `http://100.100.100.200/latest/dynamic/instance-identity/document`

huww98 (changes requested):
Is it possible to use OIDC in an non-ACK cluster? It is **Open** ID connect anyway.

pkg/mounter/fuse_containerized_mounter.go
```diff
@@ -10,6 +10,7 @@ import (
 
 	"github.com/sirupsen/logrus"
 	corev1 "k8s.io/api/core/v1"
+	apiErr "k8s.io/apimachinery/pkg/api/errors"
```
tydra-wang, line 13:
```suggestion
	apierrors "k8s.io/apimachinery/pkg/api/errors"
```

pkg/mounter/ossfs.go
AlbeeSo: Just to be consistent with ack-pod-identity-webhook

pkg/mounter/fuse_containerized_mounter.go
AlbeeSo: additional RBAC policy is a problem, but should we create such a SA by default? i think most of the users will not use RRSA

pkg/mounter/ossfs.go
huww98: Then that one should be changed as well. 0600 is exactly the same as 420. 420 does not make sense in the context of file mode.

pkg/mounter/fuse_containerized_mounter.go
huww98: I think it is OK. This SA does not have any RBAC role binding.

AlbeeSo:
> Is it possible to use OIDC in an non-ACK cluster? It is **Open** ID connect anyway.

IMO,  it is too customized and will bring extra params. if users have no strong demand on RRSA, why not use IBMIAM auth with ossfs mount options directly instead?

huww98:
> will bring extra params

I think we can just replace the clusterId parameter with full PROVIDER_ARN. We can still build this ARN for ACK user at installation.

AlbeeSo:
> > will bring extra params
> 
> I think we can just replace the clusterId parameter with full PROVIDER_ARN. We can still build this ARN for ACK user at installation.

current version:

- get aliuid from env first, retryGetMetadata if missing
- get clusterid from env first, get cluster profile configmap if missing
- get provider name from env first, used "ack-rrsa-clusterid" if missing

pkg/cloud/metadata/ecs.go
```diff
@@ -23,6 +23,7 @@ type InstanceIdentityDocument struct {
 	InstanceID   string `json:"instance-id"`
 	InstanceType string `json:"instance-type"`
 	SerialNumber string `json:"serial-number"`
+	AliUID       string `json:"owner-account-id"`
```
huww98, line 26:
```suggestion
	OwnerAccountID string `json:"owner-account-id"`
```
It can be confusing if the json tag is differ from field name

pkg/mounter/fuse_containerized_mounter.go
```diff
@@ -232,6 +247,15 @@ func (mounter *ContainerizedFuseMounter) labelsAndListOptionsFor(target string)
 	return labels, listOptions
 }
 
+func (mounter *ContainerizedFuseMounter) checkSeriveAccount(ctx context.Context) error {
+	saClient := mounter.client.CoreV1().ServiceAccounts(mounter.namespace)
+	_, err := saClient.Get(ctx, mounter.authCfg.ServiceAccountName, metav1.GetOptions{})
```
mowangdk, line 252: why not merged these two statements?

pkg/mounter/helper.go
```diff
@@ -86,3 +92,49 @@ func ValidateAnnotationsSize(annotations map[string]string) error {
 	}
 	return nil
 }
+
+// GetClusterId get clusterId from env or profile for RRSA
+func GetClusterId() (clusterId string, err error) {
```
mowangdk, line 97: Put this part into metadata module

pkg/cloud/metadata/env.go
```diff
@@ -6,6 +6,7 @@ type ENVMetadata struct{}
 
 var MetadataEnv = map[MetadataKey]string{
 	RegionID: "REGION_ID",
+	AliUID:   "ALI_UID",
```
huww98, line 9:
```suggestion
	AliUID:   "ALIBABA_CLOUD_ACCOUNT_ID",
```
to be consistent with the official env for access key, etc.

BTW, can we get the UID from the access key? https://next.api.aliyun.com/api/Sts/2015-04-01/GetCallerIdentity

pkg/mounter/ossfs.go
```diff
@@ -275,37 +289,111 @@ func CleanupOssfsCredentialSecret(ctx context.Context, clientset kubernetes.Inte
 	return err
 }
 
-func buildAuthSpec(nodeName, volumeId string, authCfg *AuthConfig,
+func buildAuthSpec(nodeName, volumeId, target string, authCfg *AuthConfig, rrsaCfg *RrsaConfig,
```
mowangdk, line 292: may be we can move ```rrsaCfg``` into  ```authCfg```

pkg/mounter/helper.go
```diff
@@ -86,3 +92,49 @@ func ValidateAnnotationsSize(annotations map[string]string) error {
...
+	if err != nil {
+		return
+	}
+	clusterId = profile.Data["clusterid"]
+	return
+}
+
+// GetOIDCProvider get OIDC provider from env or ACK clusterId for RRSA
+func GetOIDCProvider() (provider string, err error) {
+	provider = os.Getenv("OIDC_PROVIDER")
```
huww98, line 120: I still would like this one to override the full ARN, and can be called "ALIBABA_CLOUD_OIDC_PROVIDER_ARN" for consistency. Can't a user potentially use a provider from others account?

And should this be a parameter in storage class? Maybe we can support multiple providers for different PV, or even different CSI drivers. Or CSI itself may gain OIDC capability in the future. Put it here can be ambiguous.

huww98 (changes requested):

pkg/cloud/metadata/metadata.go
```diff
@@ -20,6 +20,7 @@ const (
 	ZoneID       MetadataKey = iota
 	InstanceID   MetadataKey = iota
 	InstanceType MetadataKey = iota
+	AliUID       MetadataKey = iota
```
huww98, line 23:
```suggestion
	AccountID    MetadataKey = iota
```
I remember that we use AccountID to reference main account UID across OpenAPIs. We may follow that. "Ali" does not make much sense since this repo is all about alibaba cloud.

pkg/mounter/helper.go
AlbeeSo: If role_arn or provider_arn does not conform to this fixed format, it means that the user is trying to access the OSS from other cloud vendor through RRSA. But ossfs itself does not have the ability to parse the returns of STS services from other cloud vendors. CSI can accept a full arn but it will still not work in ossfs.  

And for multiple providers, I think supporting multiple roles is enough. @mowangdk PTAL

huww98: I mean the user may also want to override the account ID part of ARN. Consider that user A may want to access an OSS bucket from account B. User B then adds the ACK cluster in account A to his OIDC provider, then user A will need to specify the provider from account B.

AlbeeSo: sts will return error like 
image

huww98: I've tested that I can register ACK cluster in account A to OIDC provider in account B. Then everything else should only relate to account B. Which step (which OpenAPI) would fail?

AlbeeSo: could you please give a detailed test case and the AssumeRoleWithOIDC related logs of ossfs?

huww98: I don't know much about ossfs. But I've tried my proposed flow, it works and I got token from AssumeRoleWithOIDC.
1. Login account A, create an ACK cluster, enable RRSA, copy the provider URL
2. Create a Pod in above ACK cluster, mount this volume, and copy the token in it.
   ```yaml
   - name: rrsa-oidc-token
     projected:
       defaultMode: 0600
       sources:
       - serviceAccountToken:
           audience: sts.aliyuncs.com
           expirationSeconds: 3600
           path: token
   ```
3. Login account B, go to https://ram.console.aliyun.com/providers/create/OIDC and register an OIDC provider with URL from step 1, copy the provider ARN
4. create a RAM role for OIDC, copy the role ARN
5. go to https://next.api.aliyun.com/api/Sts/2015-04-01/AssumeRoleWithOIDC, fill in token from step 2, and ARNs from step 3 and 4, I get RAM sts token.

These steps should naturally apply to ossfs.

pkg/oss/nodeserver.go
```diff
@@ -152,6 +154,12 @@ func (ns *nodeServer) NodePublishVolume(ctx context.Context, req *csi.NodePublis
 			opt.AuthType = strings.ToLower(strings.TrimSpace(value))
 		} else if key == "rolename" {
 			opt.RoleName = strings.TrimSpace(value)
+		} else if key == "rolearn" {
+			opt.RoleArn = strings.TrimSpace(value)
+		} else if key == "providerarn" {
+			opt.RoleArn = strings.TrimSpace(value)
```
huww98, line 160: typo? why providerarn is assigned to RoleArn?

huww98:
After experimenting with the OIDC process myself, I wonder if a user would use different service account for different PV to control the risk? If the service account token is compromised, then the attacker will gain access to all RAM roles associated to it.

pkg/oss/nodeserver.go
AlbeeSo: oh, sorry that's a mistake I made in merging, oidcProviderArn is used acutally.

AlbeeSo:
> After experimenting with the OIDC process myself, I wonder if a user would use different service account for different PV to control the risk? If the service account token is compromised, then the attacker will gain access to all RAM roles associated to it.

It's OK, set serviceaccountname as optional param of PV, which should created by users and CSI has no authority to double check before mounting. And csi-fuse-ossfs is still the default SA created when CSI installed.

pkg/oss/utils.go
```diff
@@ -61,6 +63,24 @@ func GetRAMRoleOption() string {
...
 }
 
+// GetRRSAConifg get oidcProviderArn and roleArn
+func GetRRSAConifg(opt *Options, m metadata.MetadataProvider) *mounter.RrsaConfig {
+	saName := fuseServieAccountName
+	if opt.ServiceAccountName != "" {
+		saName = opt.ServiceAccountName
+	}
+
+	if opt.OidcProviderArn != "" && opt.RoleArn != "" {
```
huww98, line 73: If I just set one of `OidcProviderArn` or `RoleArn`, then it is ignored silently. Why not accept them?

Can `RoleArn` replace `RoleName`?

pkg/oss/utils.go
```diff
@@ -61,6 +63,24 @@ func GetRAMRoleOption() string {
...
+		return &mounter.RrsaConfig{ServiceAccountName: saName, OidcProviderArn: opt.OidcProviderArn, RoleArn: opt.RoleArn}
+	}
+
+	accountId, _ := m.Get(metadata.AccountID)
```
huww98, line 77: We'd better propagate the error.

pkg/oss/nodeserver.go
```diff
@@ -363,6 +398,14 @@ func checkOssOptions(opt *Options) error {
 
 	switch opt.AuthType {
 	case mounter.AuthTypeSTS:
+	case mounter.AuthTypeRRSA:
+		if opt.RoleName == "" && (opt.RoleArn == "" || opt.OidcProviderArn == "") {
+			return errors.New("Oss Parametes error: use RRSA but roleName is empty ")
```
huww98, line 403: Error message should be updated. BTW, why leave a space at the end?

huww98, line 403: typo: Parametes => parameters

huww98 (changes requested):

pkg/oss/utils.go
AlbeeSo: Related param-check in nodeserver already existed to ensure "both ARNs are set, or RoleName is set."
I don't think allow RoleArn to replace RoleName is a good idea, it's better to declare both ARNs explicitly if the user want to access OSS across accounts (which is not recommended in fact, and will not be exposed to official documents). Or else,  a simple roleName is enough, which can avoid different accountIDs between roleArn and oidcProviderArn.

huww98: e.g., user may specify RoleName + OidcProviderArn. Then OidcProviderArn is ignored. ARNs are not for cross account access only.

I would sugguest:
- at least one of RoleName or RoleArn is required. When both specified, raise error on conflict, and accept if they are the same.
- OidcProviderArn is optional

pkg/oss/utils.go
```diff
@@ -61,6 +64,44 @@ func GetRAMRoleOption() string {
 	return mntCmdRamRole
 }
 
+// checkRRSAParams check parameters of RRSA
+func checkRRSAParams(opt *Options) error {
+	if opt.RoleArn != "" && opt.OidcProviderArn != "" {
+		return nil
+	}
+	if opt.RoleArn != "" || opt.OidcProviderArn != "" {
+		return errors.New("Oss Parameters error: use RRSA but one of the ARNs is empty, roleArn:" + opt.RoleArn + ", oidcProviderArn:" + opt.OidcProviderArn)
```
huww98, line 73:
```suggestion
		return errors.New("Oss parameters error: use RRSA but one of the ARNs is empty, roleArn:" + opt.RoleArn + ", oidcProviderArn:" + opt.OidcProviderArn)
```

pkg/oss/utils.go
```diff
@@ -61,6 +64,44 @@ func GetRAMRoleOption() string {
...
+	}
+	if opt.RoleName == "" {
+		return errors.New("Oss Parameters error: use RRSA but roleName is empty")
+	}
+	return nil
+}
+
+// getRRSAConifg get oidcProviderArn and roleArn
+func getRRSAConifg(opt *Options, m metadata.MetadataProvider) (rrsaCfg *mounter.RrsaConfig, err error) {
```
huww98, line 82:
```suggestion
func getRRSAConfig(opt *Options, m metadata.MetadataProvider) (rrsaCfg *mounter.RrsaConfig, err error) {
```

huww98 (changes requested):

pkg/oss/utils.go
AlbeeSo: For beginner users, RoleName is enough and much easier.
If multiple providers are in need, use both params can also achieve the goal. 
The functions are the same but it can reduce the risk of mismatch between ARNs for the across account case, and facilitate users to adjust parameters. Again, @mowangdk PTAL.

huww98 (approved):
Overvall it is OK.

Although I reserve my opinion about the RoleARN/Name. I think API should be minimal. To use only RoleARN and not RoleName as parameter:

pros:
- reduce the possibility of invalid input. simpler implementation, less surprise
- fewer things for user to understand. Especially for those who use both default and non-default provider.
- closer to our OIDC OpenAPI

cons:
- harder for users who just want to use OSS without AK. They should find and copy the ARN from the web console, instead of just typing in the role name.
- storage class yaml is a little harder to read if one don't know about role ARN.

PR #930, opened 2023-12-19
disk: wait to serialize attach & detach

huww98:
#### What type of PR is this?

/kind feature

Avoid backoff caused by InvalidOperation.Conflict thus speed up attaching.
Prioritize detach over attach to avoid InstanceDiskLimitExceeded error with best effort
set DISK_SERIAL_ATTACH env to opt-in this behaviour in controller

Fixes #

This is necessary to pass the external-storage test in k/k repository. Or we will almost certainly get timeout for volume not attached in 5 minutes.

```release-note
Attach/detach multiple volumes to one node at the same time is optimized to have higher throughput, and less conflict error.
```

```docs

```

### DIFF
--- a/pkg/disk/attachdetach_slot.go
+++ b/pkg/disk/attachdetach_slot.go
@@ -1,0 +1,102 @@
+package disk
+
+import (
+	"context"
+	"sync"
+)
+
+type AttachDetachSlots struct {
+	mu          sync.RWMutex
+	nodes       map[string]slot
+	slotFactory func() slot
+}
+
+func (a *AttachDetachSlots) GetSlotFor(node string) slot {
+	a.mu.RLock()
+	if s, ok := a.nodes[node]; ok {
+		a.mu.RUnlock()
+		return s
+	}
+	a.mu.RUnlock()
+
+	a.mu.Lock()
+	defer a.mu.Unlock()
+	if s, ok := a.nodes[node]; ok {
+		return s
+	}
+	s := a.slotFactory()
+	a.nodes[node] = s
+	return s
+}
+
+type slot interface {
+	AquireDetach(ctx context.Context) error
+	AquireAttach(ctx context.Context) error
+	Release()
+}
+
+type serialSlot struct {
+	// slot is a buffered channel with size 1.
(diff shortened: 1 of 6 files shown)


pkg/disk/attachdetach_slot.go
```diff
@@ -0,0 +1,80 @@
...
+type serialSlot struct {
+	detachSlot chan struct{}
+	normalSlot chan struct{}
+}
+
+func (s *serialSlot) AquireDetach(ctx context.Context) error {
+	select {
+	case s.detachSlot <- struct{}{}:
+		return nil
+	case s.normalSlot <- struct{}{}:
```
mowangdk, line 33: The test cases should meet the following scenarios on the same node
1. Delete the application pod after restarting the controller pod, triggering an uninstallation.
2. After restarting controller pod, delete/create different pods at the same time, trigger mounting/unmounting at the same time.
3. after restarting the controller pod, create three pods at the same time, two mount and one unmount, and the unmounting is completed first.

pkg/disk/attachdetach_slot.go
```diff
@@ -0,0 +1,80 @@
...
+}
+
+type slot interface {
+	AquireDetach(ctx context.Context) error
+	AquireAttach(ctx context.Context) error
+	Release()
+}
+
+type serialSlot struct {
+	detachSlot chan struct{}
```
mowangdk, line 25: slot & priority slot would be better

huww98: OK, I've updated the naming, and added some comments.

pkg/disk/attachdetach_slot.go
huww98: About "restarting the controller". The invariant here is the slot buffer is filled if and only if an attach or detach is in progress. This holds naturally when the process restarted.

I've run this through the official k/k external storage e2e test, with 25 cases in parallel. It works smoothly.

huww98:
I replaced sync.Map with map + sync.RWMutex, which is faster and avoids unnecessary allocations.

PR #931, opened 2023-12-19
 fix: containerized ossfs metrics

AlbeeSo:
#### What type of PR is this?
/kind bug

enable containerized ossfs metrics server

Fixes #

```release-note

```

```docs

```

### DIFF
--- a/build/fuse-clients/ossfs/install-ossfs.sh
+++ b/build/fuse-clients/ossfs/install-ossfs.sh
@@ -4,7 +4,7 @@ echo "TARGETPLATFORM: $TARGETPLATFORM"
 echo "installing ossfs"
 case $TARGETPLATFORM in
     linux/amd64)
-        yum install -y https://ack-csiplugin.oss-cn-hangzhou.aliyuncs.com/pre/ossfs/ossfs_1.88.3_centos8.0_x86_64.rpm
+        yum install -y https://ack-csiplugin.oss-cn-hangzhou.aliyuncs.com/pre/ossfs/ossfs_1.88.4_centos8.0_x86_64.rpm
         ;;
     linux/arm64)
         yum install -y \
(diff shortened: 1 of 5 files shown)


pkg/metric/fuse_stat_collector.go
```diff
@@ -12,7 +13,7 @@ import (
 
 var (
 	fsClientPathPrefix      = "/host/var/run/"
-	fsClientTypeArray       = []string{"ossfs", "efc"}
+	fsClientTypeArray       = []string{"efc", "ossfs"}
```
tydra-wang, line 16: what's the difference?

build/fuse-clients/ossfs/install-ossfs.sh
```diff
@@ -4,7 +4,7 @@ echo "TARGETPLATFORM: $TARGETPLATFORM"
 echo "installing ossfs"
 case $TARGETPLATFORM in
     linux/amd64)
-        yum install -y https://ack-csiplugin.oss-cn-hangzhou.aliyuncs.com/pre/ossfs/ossfs_1.88.3_centos8.0_x86_64.rpm
+        yum install -y https://ack-csiplugin.oss-cn-hangzhou.aliyuncs.com/pre/ossfs/ossfs_1.88.4_centos8.0_x86_64.rpm
```
tydra-wang, line 7: we are not going to support metrics on arm64 platform?

pkg/metric/fuse_stat_collector.go
AlbeeSo: No difference in fact ... I tried to rewrite a collector for ossfs firstly, so I removed it from the list :/

build/fuse-clients/ossfs/install-ossfs.sh
AlbeeSo: We've never supported ossfs metrics on arm64 ever. It will be available after ossfs rebasing 1.91 version.

PR #949, opened 2024-01-15
fix metrics collector panic

tydra-wang:
#### What type of PR is this?

/kind bug

Fixes #

kubelet stats summary collector may panic in k8s 1.20 version.

```release-note

```

```docs

```

### DIFF
--- a/pkg/metric/collector.go
+++ b/pkg/metric/collector.go
@@ -2,6 +2,7 @@ package metric
 
 import (
 	"errors"
+	"runtime/debug"
 	"sync"
 	"time"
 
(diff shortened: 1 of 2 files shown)


pkg/metric/collector.go
```diff
@@ -98,6 +99,11 @@ func (csi CSICollector) Collect(ch chan<- prometheus.Metric) {
 }
 
 func execute(name string, c Collector, ch chan<- prometheus.Metric) {
+	defer func() {
```
mowangdk, line 102: Can we just stop collector when summary collect panic, Retry is pointless.

tydra-wang: The summary collector currently will not panic anymore after fix. 

Abount this snippet, just in case any individual collector panic occurred, the `execute` should  recover from panic, allowing the metrics server to maintain accurate reporting of all other metrics.

mowangdk: gotcha

PR #950, opened 2024-01-15
skip removal if mountPoint does not exist

AlbeeSo:
#### What type of PR is this?
/kind bug

Fixes #

```release-note

```

```docs

```

### DIFF
--- a/pkg/oss/nodeserver.go
+++ b/pkg/oss/nodeserver.go
@@ -380,12 +380,18 @@ func (ns *nodeServer) NodeUnpublishVolume(ctx context.Context, req *csi.NodeUnpu
 	if notmounted {
 		mountPoints := GetOssfsMountPoints()
 		log.Warnf("NodeUnpublishVolume: mount point %s is unmounted, got ossfs mount point list: %v", mountPoint, mountPoints)
-		if removeErr := os.Remove(mountPoint); removeErr != nil {
-			log.Errorf("NodeUnpublishVolume: Could not remove mount point %s with error %v", mountPoint, removeErr)
-			return nil, status.Errorf(codes.Internal, "Could not remove mount point %s: %v", mountPoint, removeErr)
+		removeErr := os.Remove(mountPoint)
+		if removeErr == nil {
+			log.Infof("NodeUnpublishVolume: %s is unmounted and empty, removed successfully", mountPoint)
+			return &csi.NodeUnpublishVolumeResponse{}, nil
 		}
-		log.Infof("NodeUnpublishVolume: %s is unmounted and empty, removed successfully", mountPoint)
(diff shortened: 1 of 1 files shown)


pkg/oss/nodeserver.go
```diff
@@ -380,12 +380,18 @@ func (ns *nodeServer) NodeUnpublishVolume(ctx context.Context, req *csi.NodeUnpu
...
-			log.Errorf("NodeUnpublishVolume: Could not remove mount point %s with error %v", mountPoint, removeErr)
-			return nil, status.Errorf(codes.Internal, "Could not remove mount point %s: %v", mountPoint, removeErr)
+		removeErr := os.Remove(mountPoint)
+		if removeErr == nil {
+			log.Infof("NodeUnpublishVolume: %s is unmounted and empty, removed successfully", mountPoint)
+			return &csi.NodeUnpublishVolumeResponse{}, nil
 		}
-		log.Infof("NodeUnpublishVolume: %s is unmounted and empty, removed successfully", mountPoint)
-		return &csi.NodeUnpublishVolumeResponse{}, nil
+		if os.IsNotExist(err) {
```
huww98, line 388: I think `IsLikelyNotMountPoint` will return error if `mountPoint` does not exist before `os.Remove`?

AlbeeSo: yes, then it will check with func IsOssfsMounted by "mount | grep fuse.ossfs" and var notmounted is true.
so we need to double check here.

huww98: Then how about just return success if `IsLikelyNotMountPoint` returns not found?

Or do we have such cases where the mount point exists, but stat it would return "not found" (e.g., the mount target path is removed externally)?

pkg/oss/nodeserver.go
```diff
@@ -420,13 +426,17 @@ func (ns *nodeServer) NodeUnpublishVolume(ctx context.Context, req *csi.NodeUnpu
...
-		return nil, status.Errorf(codes.Internal, "Could not remove mount point %s: %v", mountPoint, removeErr)
+	removeErr := os.Remove(mountPoint)
+	if removeErr == nil {
+		log.Infof("NodeUnpublishVolume:: Umount OSS Successful: %s", mountPoint)
+		return &csi.NodeUnpublishVolumeResponse{}, nil
 	}
-
-	log.Infof("NodeUnpublishVolume:: Umount OSS Successful: %s", mountPoint)
-	return &csi.NodeUnpublishVolumeResponse{}, nil
+	if os.IsNotExist(err) {
```
mowangdk, line 434: is this a typo?  It should be ```removeErr``` here

mowangdk: lgtm

PR #953, opened 2024-01-15
disk: fix config file leakage if VolumeDevice is detached from DeleteVolume

huww98:
#### What type of PR is this?

/kind bug

disk: fix config file leakage if VolumeDevice is detached from DeleteVolume.

When kubelet restarts, it will call StageVolume again. and if the leakage does happen, we will detach and attach the disk again, because `IsDeviceUsedOthers` returns true, which is dangerous if the user of this disk is running.

Fixes #

```release-note
Fix potential I/O error after kubelet restarts.
```

```docs

```

### DIFF
--- a/pkg/disk/utils.go
+++ b/pkg/disk/utils.go
@@ -715,6 +715,23 @@ func saveVolumeConfig(volumeID, devicePath string) error {
 	if err := removeVolumeConfig(volumeID); err != nil {
 		return err
 	}
+	// cleanup all config files that is pointing to devicePath. Such files may be leaked
+	// if previous UnstageVolume is skipped. This is possible if the VolumeDevice is
+	// detached by others (e.g. DeleteVolume).
+	files, err := os.ReadDir(VolumeDir)
+	if err != nil {
+		return err
+	}
+	for _, file := range files {
+		if file.Type().IsRegular() && strings.HasSuffix(file.Name(), ".conf") {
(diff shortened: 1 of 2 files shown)


huww98:
Manually tested. This is only reproducible when upgrade from 1.22 to 1.24, where the globalmount path changed.

pkg/disk/utils.go
```diff
@@ -715,6 +715,23 @@ func saveVolumeConfig(volumeID, devicePath string) error {
...
+			if getVolumeConfig(tmpVolID) == devicePath {
+				if err := removeVolumeConfig(tmpVolID); err != nil {
+					return fmt.Errorf("failed to remove volume config for %s: %w", tmpVolID, err)
+				}
+			}
+		}
+	}
 
 	volumeFile := path.Join(VolumeDir, volumeID+".conf")
 	if err := ioutil.WriteFile(volumeFile, []byte(devicePath), 0644); err != nil {
```
mowangdk, line 737: ioutil is deprecated, we need to change here later

pkg/disk/utils.go
```diff
@@ -715,6 +715,23 @@ func saveVolumeConfig(volumeID, devicePath string) error {
 	if err := removeVolumeConfig(volumeID); err != nil {
```
mowangdk, line 715: This whole logic base on ```VolumeDir``` are strange , we will delete it eventually

PR #954, opened 2024-01-16
perf(fuse): add/remove "used-by-pod" annotations on fuse pod when publishing/unpublishing

AlbeeSo:
add/remove annotations like `used-by-pod/09ec7b80-71d1-4fe9-8fa1-ea234563b05a: "20240116195325"` on fuse pod when publishing/unpublishing, to help users understand the mapping between application pods and fuse pods

Fixes #

```release-note

```

```docs

```

### DIFF
--- a/pkg/mounter/fuse_containerized_mounter.go
+++ b/pkg/mounter/fuse_containerized_mounter.go
@@ -33,7 +33,7 @@ const (
 	FuseMountPathHashLabelKey  = "csi.alibabacloud.com/mount-path-hash"
 	FuseMountPathAnnoKey       = "csi.alibabacloud.com/mount-path"
 	FuseSafeToEvictAnnoKey     = "cluster-autoscaler.kubernetes.io/safe-to-evict"
-	FuseUsedByPodAnnoKeyPrefix = "csi.alibabacloud.com/used-by-pod"
+	FuseUsedByPodAnnoKeyPrefix = "used-by-pod"
 )
 
 type AuthConfig struct {
@@ -199,7 +199,10 @@ func (mounter *ContainerizedFuseMounter) Mount(source string, target string, fst
 			if err != nil {
 				return err
 			}
-			mounter.addPodAnnotation(ctx, target, mounter.podUid)
+			err = mounter.addPodAnnotation(ctx, source, mounter.podUid)
+			if err != nil {
+				mounter.log.Warnf("addPodAnnotation failed err: %v", err)
+			}
 			return nil
 		}
 	}
@@ -343,6 +346,9 @@ func (mounter *ContainerizedFuseMounter) launchFusePod(ctx context.Context, sour
 }
 
 func (mounter *ContainerizedFuseMounter) addPodAnnotation(ctx context.Context, target, podUid string) error {
+	if podUid == "" {
+		return fmt.Errorf("podUid is empty, skip adding pod annotation")
+	}
 	podClient := mounter.client.CoreV1().Pods(mounter.namespace)
 	_, listOptions := mounter.labelsAndListOptionsFor(target)
 	podList, err := podClient.List(ctx, listOptions)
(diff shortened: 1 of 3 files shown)


pkg/oss/utils.go
```diff
@@ -99,6 +99,16 @@ func parseOtherOpts(otherOpts string) (mountOptions []string, err error) {
 	return mountOptions, nil
 }
 
+func getPodUid(targetPath, volumeID string) (podUid string) {
```
huww98, line 102: See https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.28/#csidriverspec-v1-storage-k8s-io (podInfoOnMount) for how to get pod info. Maybe we should not rely on the format of targetPath.

pkg/mounter/fuse_containerized_mounter.go
```diff
@@ -273,9 +292,9 @@ func (mounter *ContainerizedFuseMounter) launchFusePod(ctx context.Context, sour
 		var rawPod corev1.Pod
 		rawPod.GenerateName = fmt.Sprintf("csi-fuse-%s-", mounter.name())
 		rawPod.Labels = labels
-		rawPod.Annotations = map[string]string{FuseMountPathAnnoKey: target, FuseSafeToEvictAnnoKey: "true"}
+		rawPod.Annotations = map[string]string{FuseMountPathAnnoKey: target, FuseSafeToEvictAnnoKey: "true", getUsedByPodAnnoKey(mounter.podUid): time.Now().Format("20060102150405")}
```
huww98, line 295: Why not standard format? e.g. `time.RFC3339`

huww98 (commented):
I would recommend `csi.alibabacloud.com/used-by-pod/namespace/pod-name` as annotation name.

pkg/oss/utils.go
AlbeeSo: it seems that only NodePublishVolumeRequest  has related volume_context of podInfoOnMount. But we need it on unpublish stage.
BTW, `csi.alibabacloud.com/used-by-pod/namespace/pod-name` is invalid. annotation key could only has ONE "/".

huww98:
This annotation should always have the same result as `kubectl describe pvc xxx` now, right?

But if this is combined with #957 , consider:
1. Pod "a" start, fuse pod "fuse-a" start with used-by-pod/a
2. fuse pod "fuse-a" fail
3. Pod "b" start, fuse pod "fuse-b" start with used-by-pod/b
4. Pod "b" finished, then "fuse-b" is left with no used-by-pod

This seems OK.

pkg/oss/utils.go
huww98: OK, then the part before "/" should be a domain name.

AlbeeSo:
> This annotation should always have the same result as `kubectl describe pvc xxx` now, right?
> 
> But if this is combined with #957 , consider:
> 
> 1. Pod "a" start, fuse pod "fuse-a" start with used-by-pod/a
> 2. fuse pod "fuse-a" fail
> 3. Pod "b" start, fuse pod "fuse-b" start with used-by-pod/b
> 4. Pod "b" finished, then "fuse-b" is left with no used-by-pod
> 
> This seems OK.

here is a small demo for fuse pod crash case:

1. create  an app pod with oss pv, with a fuse pod `csi.alibabacloud.com/used-by-pod-5c7af695-e754-4b5b-9f08-609a76f6eac2: "2024-01-18T20:45:02+08:00"`
2. fuse pod crash by killing fuse daemon
3. create another app pod, fuse restart with annotation `csi.alibabacloud.com/used-by-pod-c0e4839c-8f73-41de-99ad-e998cddbdaba: "2024-01-18T21:01:09+08:00"`
4. pod `5c7af695-e754-4b5b-9f08-609a76f6eac2` need to restart even it is on pvc's used-by list.

huww98:
I still have concerns about using UID in the user interface.

How about:
* label: csi.alibabacloud.com/namespace: "<namespace of pvc/pod>"
* annotation: csi.alibabacloud.com/used-by-pod: "\<pod-name1\>,\<pod-name2\>"

I know this is harder to implement, as we need to pull the old annotation to add new pod. But it may be more friendly to users.

`csi.alibabacloud.com/used-by-pod-<pod name>` have issue about the length limit on annotation name

AlbeeSo:
> I still have concerns about using UID in the user interface.
> 
> How about:
> 
> * label: csi.alibabacloud.com/namespace: "<namespace of pvc/pod>"
> * annotation: csi.alibabacloud.com/used-by-pod: "<pod-name1>,<pod-name2>"
> 
> I know this is harder to implement, as we need to pull the old annotation to add new pod. But it may be more friendly to users.
> 
> `csi.alibabacloud.com/used-by-pod-<pod name>` have issue about the length limit on annotation name

the point is we cannot get the pod name on unpublish stage unless we record it.

PR #958, opened 2024-01-23
cleanup /etc/csi-tool

huww98:
#### What type of PR is this?

/kind cleanup

`/etc` is for config files, and is not suitable for socket, rpm package...

connector: use runtime directory for socket

Also includes some cleanups, see commits.

Fixes #

```release-note
NONE
```

```docs

```


build/lib/csiplugin-connector.service
```diff
@@ -3,11 +3,9 @@ Description=csiplugin connector
 After=network.target remote-fs.target nss-lookup.target
 
 [Service]
-Type=forking
-RuntimeDirectory=csiplugin
+Type=simple
+RuntimeDirectory=csi-tool/connector
```
huww98, line 7: Two level of directory is required, because the inner one will be removed by systemd when the service is stopped/restarted.

mowangdk:
/lgtm 

@tydra-wang ptal~

PR #966, opened 2024-02-02
pass supported disk type as CSI topology

huww98:
#### What type of PR is this?

/kind cleanup

We no longer update Kubernetes Node with supported disk type labels. kubelet will add them instead.

We also no longer query the Node for supported disk types, external-provisioner will pass them as preferred topology.

We no longer set the annotation csi.alibabacloud.com/volume-topology in PV, which is opaque to user. They are set in nodeAffinity instead.

As the consequence, we can remove some non-standard operation:
- support for csi.alibabacloud.com/annotation-prefix/ to add annotations to PV
- support for reading csi.alibabacloud.com/volume-topology annotation in scheduler

This also make us depends less on Kubernetes, which we should not as a CSI driver

Fixes #

```release-note
NONE
```

```docs

```

### DIFF
--- a/pkg/disk/cloud.go
+++ b/pkg/disk/cloud.go
@@ -40,9 +40,6 @@ import (
 	log "github.com/sirupsen/logrus"
 	"google.golang.org/grpc/codes"
 	"google.golang.org/grpc/status"
-	apierrors "k8s.io/apimachinery/pkg/api/errors"
-	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
-	"k8s.io/apimachinery/pkg/util/sets"
 	"k8s.io/apimachinery/pkg/util/wait"
 )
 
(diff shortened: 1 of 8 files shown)


huww98:
Tested manually, now the PV looks like:
```yaml
spec:
  nodeAffinity:
    required:
      nodeSelectorTerms:
      - matchExpressions:
        - key: topology.diskplugin.csi.alibabacloud.com/zone
          operator: In
          values:
          - cn-beijing-g
        - key: node.csi.alibabacloud.com/disktype.cloud_essd
          operator: In
          values:
          - available
```

huww98:
/hold
once we add it as nodeAffinity, it is immutable. But we need to change it on ModifyVolume

huww98:
/remove-lifecycle stale
/ok-to-test

CI failed. It seems to land this PR, we need external-provisioner to send us disk type info even if only zone is specified in `AllowedTopologies`.

Jul 27 18:43:45.896: INFO: At 2024-07-27 18:38:45 +0000 UTC - event for pvc-phc2r: {diskplugin.csi.alibabacloud.com_csi-provisioner-6cfb78cf65-9zvzj_f6b8a5de-c90a-4aed-b1bc-3f667f49add2 } ProvisioningFailed: failed to provision volume with StorageClass "topology-3434-e2e-sc8944j": rpc error: code = InvalidArgument desc = Invalid parameters from input: disk-25e3b078-386d-45f5-8518-6db4fd19238a, with error: no disk type satisfies the topology requirement: requested [cloud_essd cloud_ssd cloud_efficiency], topology []

PR #967, opened 2024-02-02
disk: set IO limit before mount

huww98:
#### What type of PR is this?

/kind bug

ensure we retry set IO limit after it failed.

Fixes #

```release-note
Fix a bug that IO limit setting may be not effective.
```

```docs

```


dongjiang1989 (commented):
LGTM

PR #969, opened 2024-02-04
chore: fix create disks returned error

mowangdk:
#### What type of PR is this?
/kind bug

Exposing CreateVolume errors via events on pvc

None

```release-note

```

```docs

```

### DIFF
--- a/pkg/disk/cloud.go
+++ b/pkg/disk/cloud.go
@@ -930,7 +930,7 @@ func createDisk(diskName, snapshotID string, requestGB int, diskVol *diskVolumeA
 		}
 		err = rerr
 	}
-	return "", "", "", status.Errorf(codes.Internal, "createDisk: err: %v, the zone:[%s] is not support specific disk type, please change the request disktype: %s or disk pl: %s", err, diskVol.ZoneID, diskTypes, diskPLs)
+	return "", "", "", fmt.Errorf("createDisk: err: %v, the zone:[%s] is not support specific disk type, please change the request disktype: %s or disk pl: %s", err, diskVol.ZoneID, diskTypes, diskPLs)
 }
 
 // reuse rpcrequest in ecs sdk is forbidden, because parameters can't be reassigned with empty string.(ecs sdk bug)
--- a/pkg/disk/controllerserver.go
+++ b/pkg/disk/controllerserver.go
@@ -214,7 +214,7 @@ func (cs *controllerServer) CreateVolume(ctx context.Context, req *csi.CreateVol
 				return nil, status.Errorf(codes.NotFound, "snapshot %s not found", snapshotID)
 			}
 		}
-		return nil, err
+		return nil, status.Error(codes.Internal, err.Error())
 	}
 
 	volumeContext := req.GetParameters()


pkg/disk/cloud.go
```diff
@@ -930,7 +930,7 @@ func createDisk(diskName, snapshotID string, requestGB int, diskVol *diskVolumeA
 		}
 		err = rerr
 	}
-	return "", "", "", status.Errorf(codes.Internal, "createDisk: err: %v, the zone:[%s] is not support specific disk type, please change the request disktype: %s or disk pl: %s", err, diskVol.ZoneID, diskTypes, diskPLs)
+	return "", "", "", fmt.Errorf("createDisk: err: %v, the zone:[%s] is not support specific disk type, please change the request disktype: %s or disk pl: %s", err, diskVol.ZoneID, diskTypes, diskPLs)
```
huww98, line 933:
```suggestion
	return "", "", "", fmt.Errorf("createDisk: err: %w, the zone %s does not support the specified disk type, please change the request disk type: %s or disk pl: %s", err, diskVol.ZoneID, diskTypes, diskPLs)
```

huww98:
> Exposing CreateVolume errors via events on pvc

This change is good, but it may not have this claimed effect. I think external provisioner will expose event regardless of the gRPC status code.

mowangdk:
> > Exposing CreateVolume errors via events on pvc
> 
> This change is good, but it may not have this claimed effect. I think external provisioner will expose event regardless of the gRPC status code.

yep, I had checked the external-provisioner code. this pr has no influences with error event, but there is one question  need to confirmed that csi event will vanish after several retries.  could you please take a look about it @huww98

huww98:
unable to reproduce. I'm able to get something like:
```
Type     Reason              Age                  From                                                                                                  
----     ------              ----                 ----
Warning  VolumeFailedDelete  34s (x32 over 117m)  diskplugin.csi.alibabacloud.com_csi-provisioner
```
It does not vanish.

huww98:
> but there is one question need to confirmed that csi event will vanish after several retries. could you please take a look about it

I investigated this issue. And found that csi-provisioner is always issuing two event in succession: "Provisioning", "ProvisioningFailed". Since the first event will take the token, the second one has a great possibility of being filtered by:
https://github.com/kubernetes-csi/external-provisioner/blob/91e80e508a9b789d833db94196524e6ac02c7ebd/vendor/k8s.io/client-go/tools/record/events_cache.go#L148-L149

This patch should fix this. But we have to modify vendor 😮‍💨. eventRecorder is not exposed as an option. 
```diff
diff --git a/vendor/sigs.k8s.io/sig-storage-lib-external-provisioner/v9/controller/controller.go b/vendor/sigs.k8s.io/sig-storage-lib-external-provisioner/v9/controller/controller.go
index d54bf45f8..93c906f95 100644
--- a/vendor/sigs.k8s.io/sig-storage-lib-external-provisioner/v9/controller/controller.go
+++ b/vendor/sigs.k8s.io/sig-storage-lib-external-provisioner/v9/controller/controller.go
@@ -614,7 +614,21 @@ func NewProvisionController(
        component := provisionerName + "_" + id
 
        v1.AddToScheme(scheme.Scheme)
-       broadcaster := record.NewBroadcaster()
+       broadcaster := record.NewBroadcaster(record.WithCorrelatorOptions(record.CorrelatorOptions{
+               SpamKeyFunc: func(event *v1.Event) string {
+                       return strings.Join([]string{
+                               event.Source.Component,
+                               event.Source.Host,
+                               event.InvolvedObject.Kind,
+                               event.InvolvedObject.Namespace,
+                               event.InvolvedObject.Name,
+                               string(event.InvolvedObject.UID),
+                               event.InvolvedObject.APIVersion,
+                               event.Reason,
+                       },
+                               "")
+               },
+       }))
        broadcaster.StartLogging(klog.Infof)
        broadcaster.StartRecordingToSink(&corev1.EventSinkImpl{Interface: client.CoreV1().Events(v1.NamespaceAll)})
        eventRecorder := broadcaster.NewRecorder(scheme.Scheme, v1.EventSource{Component: component})
```

PR #972, opened 2024-02-19
oss-perf: modify default protocol for ossfs URL

AlbeeSo:
#### What type of PR is this?
/kind feature

In current version, CSI will not check protocol of ossfs URL. Then, ossfs will use default protocol (http) if it is not specified explicitly. 
In this PR, protocol will be set by CSI. If instance needs to access OSS with public endpoint, use https as default protocol instead.

Fixes #

```release-note

```

```docs

```

### DIFF
--- a/pkg/oss/utils.go
+++ b/pkg/oss/utils.go
@@ -126,3 +126,18 @@ func doMount(mounter mountutils.Interface, target string, opts Options, mountOpt
 	logger.Info("mounted successfully")
 	return nil
 }
+
+func modifiedURL(originURL, regionID string) (URL string, modified bool) {
+	URL = originURL
+	if strings.Contains(originURL, regionID) && !strings.Contains(originURL, "internal") && !utils.IsPrivateCloud() {
+		URL = strings.ReplaceAll(originURL, regionID, regionID+"-internal")
+		modified = true
+	}
+	if strings.HasPrefix(URL, "http://") || strings.HasPrefix(URL, "https://") {
+		return
+	}
+	if strings.HasSuffix(URL, "-internal.aliyuncs.com") {
+		return "http://" + URL, true
+	}
+	return "https://" + URL, true
+}
--- a/pkg/oss/utils.go
+++ b/pkg/oss/utils.go
@@ -129,7 +129,10 @@ func doMount(mounter mountutils.Interface, target string, opts Options, mountOpt
 
 func setNetworkType(originURL, regionID string) (URL string, modified bool) {
 	URL = originURL
-	if strings.Contains(originURL, regionID) && !strings.Contains(originURL, "internal") && !utils.IsPrivateCloud() {
+	if utils.IsPrivateCloud() || !strings.HasSuffix(strings.TrimRight(URL, "/"), ".aliyuncs.com") {
+		return
+	}
+	if strings.Contains(originURL, regionID) && !strings.Contains(originURL, "internal") {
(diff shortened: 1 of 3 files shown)


pkg/oss/utils.go
```diff
@@ -126,3 +126,18 @@ func doMount(mounter mountutils.Interface, target string, opts Options, mountOpt
 	logger.Info("mounted successfully")
 	return nil
 }
+
+func modifiedURL(originURL, regionID string) (URL string, modified bool) {
```
mowangdk, line 130: Let's split this method into  ```setNetworkType``` and ```setTransmissionProtocol```

mowangdk: and we should add unit test for these two method

pkg/oss/utils.go
```diff
@@ -126,3 +126,23 @@ func doMount(mounter mountutils.Interface, target string, opts Options, mountOpt
...
+	}
+	return
+}
+
+func setTransmissionProtocol(originURL string) (URL string, modified bool) {
+	URL = originURL
+	if strings.HasPrefix(URL, "http://") || strings.HasPrefix(URL, "https://") {
+		return
+	}
+	if strings.HasSuffix(URL, "-internal.aliyuncs.com") {
```
huww98, line 144: I think at least URL like `oss-cn-beijing-internal.aliyuncs.com/` is vaild. Paths other than '/' should be considered too. scheme and hostname parts of URL are case insensitive, and should be considered.

I think we may parse the URL with `net/url` package.

huww98 (changes requested):

pkg/oss/utils.go
AlbeeSo: [url here is can only ](https://help.aliyun.com/zh/oss/user-guide/regions-and-endpoints?spm=a2c4g.11186623.0.0.1e2030f0DijQDq#section-oao-7ao-11f)
Endpoint not matches  "(http://,https://)oss-{{region}}-{{pub/internal/null}}-aliyuncs.com" is invalid to access OSS by ossfs. Use the raw URL or the modified one will both cause an error.

huww98: I don't get it. The provided link shows endpoint, which is only the hostname part of URL. URL is a term defined by [rfc3986](https://datatracker.ietf.org/doc/html/rfc3986). Case insensitivity of scheme should be supported:
> Although schemes are case-
   insensitive, the canonical form is lowercase and documents that
   specify schemes must do so with lowercase letters.  An implementation
   should accept uppercase letters as equivalent to lowercase in scheme
   names (e.g., allow "HTTP" as well as "http") for the sake of
   robustness but should only produce lowercase scheme names for
   consistency.

DNS name should also be case insensitive (see [rfc4343](https://datatracker.ietf.org/doc/html/rfc4343)).

As for the path component. I've verified that `oss-cn-beijing-internal.aliyuncs.com/` this URL works currently. We may also want to support this too.

AlbeeSo: In fact, value of -ourl= for s3fs/ossfs is little bit different from a real URL. It's for the Endpoint part of Host, which is equal to BucketName.Endpoint. So 'URL' here can only be endpoint or the endpoint with protocol.  I fixed case like `oss-cn-beijing-internal.aliyuncs.com/` . 
https://help.aliyun.com/zh/oss/developer-reference/common-http-headers?spm=a2c4g.11186623.0.0.3b2b3864sI0iwV#section-eq1-b2y-wdb

PR #980, opened 2024-02-25
bug: fix string float conversion

mowangdk:
#### What type of PR is this?
/kind bug

Fix float64 conversion

Fixes #

```release-note

```

```docs

```

### DIFF
--- a/pkg/local/capacity.go
+++ b/pkg/local/capacity.go
@@ -156,12 +156,13 @@ func getLoopDeviceCapacity() []*StorageCapacity {
 			log.Errorf("getLoopDevice: failed to get totalcapacity: %v. err: %v", totalBytes, err)
 			return nil
 		}
-		percent := 0.9
+		percentFraction := 0.9
 		if types.GlobalConfigVar.LocalSparseTotalAvailablePercent != "" {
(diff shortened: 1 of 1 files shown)


pkg/local/capacity.go
```diff
@@ -158,8 +158,9 @@ func getLoopDeviceCapacity() []*StorageCapacity {
 		}
 		percent := 0.9
 		if types.GlobalConfigVar.LocalSparseTotalAvailablePercent != "" {
-			p, _ := strconv.Atoi(types.GlobalConfigVar.LocalSparseTotalAvailablePercent)
-			percent = float64(p)
+			if p, err := strconv.ParseFloat(types.GlobalConfigVar.LocalSparseTotalAvailablePercent, 32); err == nil && p < 1 {
```
huww98, line 161: Since this variable is named "Percent", I would not expect it to be less than 1.

Also, for sparse files, over 100% should also be acceptable?

mowangdk: It's just the space available in a directory, not the sparse-file specific.

huww98: "percent" means 1/100, which is different from implementation. Should we fix this? Either change the semantic or the name of this env to avoid confusions.

PR #992, opened 2024-03-07
disk: ensure NodeUnstageVolume is called if attached but not mounted

huww98:
#### What type of PR is this?

/kind cleanup

We need NodeUnstageVolume to detach the disk.
Although we have logic to detach the disk on deleting or attaching to another node, but they are less tested.
And detaching the disk earlier is preferred.

Fixes #

```release-note
NONE
```

```docs

```


huww98:
This issue can also cause disk .conf file leakage, which may trigger #953 .

PR #996, opened 2024-03-11
Refactor credential acquisition (part 1)

huww98:
#### What type of PR is this?

/kind cleanup

Part 1 of #873

only introduce the infrastructure and not used in any driver yet.

Fixes #

```release-note
NONE
```

```docs
ACCESS_KEY_ID and related environment variables are deprecated, please use ALIBABA_CLOUD_ACCESS_KEY_ID and related instead.
```

### DIFF
--- a/pkg/credentials/credential.go
+++ b/pkg/credentials/credential.go
@@ -1,0 +1,31 @@
+package credentials
+
+import (
+	"errors"
+	"os"
+
+	alicred_old "github.com/aliyun/credentials-go/credentials"
+	alicred "github.com/aliyun/credentials-go/credentials/providers"
+	"k8s.io/utils/clock"
+)
+
+func NewProvider() (alicred.CredentialsProvider, error) {
+	// try managed token credential
+	provider, err := NewManagedTokenProvider(clock.RealClock{}, GetManagedTokenPath())
+	if err == nil {
+		return provider, nil
+	}
+	if !errors.Is(err, os.ErrNotExist) {
+		return nil, err
+	}
+
+	return alicred.NewDefaultCredentialsProvider(), nil
+}
+
+func NewCredential() (alicred_old.Credential, error) {
+	provider, err := NewProvider()
+	if err != nil {
+		return nil, err
+	}
+	return alicred_old.FromCredentialsProvider("default", provider), nil
+}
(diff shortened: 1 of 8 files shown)


pkg/credentials/managed_token.go
```diff
@@ -0,0 +1,97 @@
...
+		klog.Warningf("token already expired at %s on load", m.tokens.Expiration)
+	}
+
+	return nil
+}
+
+func (m *ManagedTokenCredential) reloadTokenIfNeeded() error {
+	now := m.clock.Now()
+	// reload token every 5 minutes, or 10 minutes before expiration
+	if now.After(m.reloadTime.Add(5*time.Minute)) || m.expiration.Before(now.Add(10*time.Minute)) {
```
mowangdk, line 56: Is it necessary to reload every 5 minutes if we're aware of the token's expiration?

pkg/credentials/managed_token.go
```diff
@@ -0,0 +1,97 @@
...
+	alicred "github.com/aliyun/credentials-go/credentials/providers"
+	"github.com/kubernetes-sigs/alibaba-cloud-csi-driver/pkg/utils/crypto"
+	"k8s.io/klog/v2"
+	"k8s.io/utils/clock"
+)
+
+const (
+	// the secret mount file
+	DefaultTokenPath = "/var/addon/token-config"
+	TokenPathEnvName = "ALIBABA_CLOUD_ACK_ADDON_TOKEN_PATH"
```
mowangdk, line 17: I don't remember we've got this ENV, new added?

pkg/credentials/credential.go
```diff
@@ -0,0 +1,59 @@
+package credentials
+
+import (
+	"errors"
+	"os"
+
+	alicred_old "github.com/aliyun/credentials-go/credentials"
```
mowangdk, line 7: Is this old credential intended for rollback?

pkg/credentials/managed_token.go
huww98: Since reload is cheap (just read the file, no network IO), I'd like to do that often. e.g. Maybe user revoked the token we are using by removing the the RAM role?

pkg/credentials/managed_token.go
huww98: Yes, it is new. It is also useful for unit tests

pkg/credentials/credential.go
huww98: Basically, Alibaba Cloud OpenAPI SDK (NAS) is still taking the old credential. So we will still use the old interface for a while. The old credential is now a thin wrapper around the new provider interface.

PR #997, opened 2024-03-12
feature: remove system disk check failed

mowangdk:
#### What type of PR is this?
/kind cleanup

User may be using sata or nvme local volume as systemdisk. so this check is invalid in some scanario. According to csi current logic after we disable system disk check, it will return normally if device is mounted on "/" . have no influence with current node

Fixes #

```release-note

```

```docs

```

### DIFF
--- a/pkg/disk/utils.go
+++ b/pkg/disk/utils.go
@@ -1065,8 +1065,7 @@ func checkDeviceAvailable(mountinfoPath, devicePath, volumeID, targetPath string
 
 	// check the device is used for system
 	if devicePath == "/dev/vda" || devicePath == "/dev/vda1" {
-		msg := fmt.Sprintf("devicePath(%s) is system device, cannot used for Volume", devicePath)
-		return status.Error(codes.Internal, msg)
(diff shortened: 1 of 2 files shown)


pkg/disk/utils.go
```diff
@@ -1065,8 +1065,7 @@ func checkDeviceAvailable(mountinfoPath, devicePath, volumeID, targetPath string
 
 	// check the device is used for system
 	if devicePath == "/dev/vda" || devicePath == "/dev/vda1" {
-		msg := fmt.Sprintf("devicePath(%s) is system device, cannot used for Volume", devicePath)
-		return status.Error(codes.Internal, msg)
+		log.Warnf("checkDeviceAvailable: devicePath(%s) is system device, cannot used for volume: %s", devicePath, volumeID)
```
huww98, line 1068:
```suggestion
		log.Warnf("checkDeviceAvailable: devicePath(%s) may be system device: %s", devicePath, volumeID)
```
to keep the log consistent with the behavior.

huww98:
LGTM, but this check is introduced in #163 , I don't know the background of that change.

mowangdk:
It's should be fine generally

PR #1002, opened 2024-03-20
fix(nas): do not set vpc endpoint for all regions

tydra-wang:
#### What type of PR is this?

/kind bug

Fixes #

```release-note

```

```docs

```

### DIFF
--- a/pkg/nas/cloud/nas_client_v1.go
+++ b/pkg/nas/cloud/nas_client_v1.go
@@ -13,11 +13,15 @@ import (
 	"github.com/kubernetes-sigs/alibaba-cloud-csi-driver/pkg/utils"
 )
 
+func init() {
+	for region := range regionsCanUseVpcEndpoint {
+		_ = aliyunep.AddEndpointMapping(region, "Nas", fmt.Sprintf("nas-vpc.%s.aliyuncs.com", region))
+	}
+}
+
 func newNasClientV1(region string) (*nassdk.Client, error) {
 	if ep := os.Getenv("NAS_ENDPOINT"); ep != "" {
 		_ = aliyunep.AddEndpointMapping(region, "Nas", ep)
-	} else {
-		_ = aliyunep.AddEndpointMapping(region, "Nas", fmt.Sprintf("nas-vpc.%s.aliyuncs.com", region))
 	}
 
 	ac := utils.GetAccessControl()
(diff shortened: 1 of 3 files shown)


pkg/nas/cloud/utils.go
```diff
@@ -14,3 +20,40 @@ func GetFilesystemTypeByMountTargetDomain(domain string) string {
 		return ""
 	}
 }
+
+var regionsCanUseVpcEndpoint = sets.NewString(
```
huww98, line 24: How can we get an update to this list? Is this intended to be temporary?

pkg/nas/cloud/nas_client_v1.go
```diff
@@ -13,11 +13,15 @@ import (
...
+	for region := range regionsCanUseVpcEndpoint {
+		_ = aliyunep.AddEndpointMapping(region, "Nas", fmt.Sprintf("nas-vpc.%s.aliyuncs.com", region))
+	}
+}
+
 func newNasClientV1(region string) (*nassdk.Client, error) {
 	if ep := os.Getenv("NAS_ENDPOINT"); ep != "" {
 		_ = aliyunep.AddEndpointMapping(region, "Nas", ep)
-	} else {
-		_ = aliyunep.AddEndpointMapping(region, "Nas", fmt.Sprintf("nas-vpc.%s.aliyuncs.com", region))
```
huww98, line 20: Why not use `GetEndpointForRegion` here?

pkg/nas/cloud/nas_client_v2.go
```diff
@@ -32,7 +32,7 @@ func newNasClientV2(region string) (*sdk.Client, error) {
 	// set endpoint
 	ep := os.Getenv("NAS_ENDPOINT")
 	if ep == "" {
-		ep = fmt.Sprintf("nas-vpc.%s.aliyuncs.com", region)
+		ep = GetEndpointForRegion(region)
```
huww98, line 35: Will this fix the issue we met? We have test failure in `cn-wulanchabu`, in which we still use vpc endpoint in the new version.

pkg/nas/cloud/nas_client_v1.go
tydra-wang: I did not find a way to directly set endpoint in sdk v1 client config

huww98: I don't get it. Shouldn't this work?
```
_ = aliyunep.AddEndpointMapping(region, "Nas", GetEndpointForRegion(region))
```

PR #1007, opened 2024-03-23
disk: create snapshot even if VolumeSnapshot is deleted

huww98:
#### What type of PR is this?

/kind bug

So that the VolumeSnapshotContent can be deleted eventually, instead of retrying infinitely.

Fixes:

E0323 16:06:48.148596 1872601 utils.go:101] GRPC error: rpc error: code = Internal desc = failed to get VolumeSnapshot: snapshotting-3919/snapshot-2trth: volumesnapshots.snapshot.storage.k8s.io "snapshot-2trth" not found

```release-note
Fix that VolumeSnapshotContent may stuck if VolumeSnapshot is deleted before snapshot is created.
```

```docs

```

### DIFF
--- a/pkg/disk/controllerserver.go
+++ b/pkg/disk/controllerserver.go
@@ -47,6 +47,7 @@ import (
 	v1 "k8s.io/api/core/v1"
 	crdv1 "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1"
 	crd "k8s.io/apiextensions-apiserver/pkg/client/clientset/clientset"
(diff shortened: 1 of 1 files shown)


huww98:
Closing this. I suspect this is bug in external-snapshotter. Returning Internal should indicate the snapshot is not-yet-created, and the volumesnapshotcontent object is safe to delete.

huww98:
This is not nessesary if https://github.com/kubernetes-csi/external-snapshotter/pull/1064 get approved.

PR #1008, opened 2024-03-28
feat: organize node expand logics

mowangdk:
#### What type of PR is this?
/kind feature

1. Add check if block device size is as expect
2. remove filesystem resize check 

Fixes # None

None

```release-note

```

```docs

```

### DIFF
--- a/pkg/disk/nodeserver.go
+++ b/pkg/disk/nodeserver.go
@@ -22,6 +22,7 @@ import (
 	"crypto/sha256"
 	"errors"
 	"fmt"
+	"math"
 	"os"
 	"os/exec"
 	"path/filepath"
@@ -888,8 +889,8 @@ func (ns *nodeServer) NodeExpandVolume(ctx context.Context, req *csi.NodeExpandV
 	*csi.NodeExpandVolumeResponse, error) {
 	log.Infof("NodeExpandVolume: node expand volume: %v", req)
 
-	volExpandBytes := int64(req.GetCapacityRange().GetRequiredBytes())
-	requestGB := float64((volExpandBytes + 1024*1024*1024 - 1) / (1024 * 1024 * 1024))
+	volExpandBytes := float64(req.GetCapacityRange().GetRequiredBytes())
+	requestGB := math.Floor(volExpandBytes / float64(1024*1024*1024))
 
 	volumePath := req.GetVolumePath()
 	diskID := req.GetVolumeId()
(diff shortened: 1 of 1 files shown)


pkg/disk/nodeserver.go
```diff
@@ -885,8 +886,8 @@ func (ns *nodeServer) NodeExpandVolume(ctx context.Context, req *csi.NodeExpandV
 	*csi.NodeExpandVolumeResponse, error) {
 	log.Infof("NodeExpandVolume: node expand volume: %v", req)
 
-	volExpandBytes := int64(req.GetCapacityRange().GetRequiredBytes())
-	requestGB := float64((volExpandBytes + 1024*1024*1024 - 1) / (1024 * 1024 * 1024))
+	volExpandBytes := float64(req.GetCapacityRange().GetRequiredBytes())
+	requestGB := math.Floor(volExpandBytes / float64(1024*1024*1024))
```
huww98, line 890: Do all calculation with int64 and bytes? to avoid any potential issue with floating point precision lost.

pkg/disk/nodeserver.go
```diff
@@ -990,14 +991,14 @@ func (ns *nodeServer) NodeExpandVolume(ctx context.Context, req *csi.NodeExpandV
...
 	}
 	deviceCapacity := getBlockDeviceCapacity(devicePath)
+	if deviceCapacity < requestGB {
+		// After calling openapi to expansion cloud disk, the size of the underlying block device may not change. need to retry
+		log.Errorf("NodeExpandVolume:: Actual block size: %s is smaller than expected block size: %s, need to retry waiting", deviceCapacity, requestGB)
+		return nil, status.Errorf(codes.Aborted, "deviceCapacity: %v, requestGB: %v not in range", deviceCapacity, requestGB)
+	}
 	log.Infof(
 		"NodeExpandVolume:: filesystem resize context device capacity: %v, before resize fs capacity: %v resize fs capacity: %v, requestGB: %v. file system lose percent: %v",
 		deviceCapacity, beforeResizeDiskCapacity, diskCapacity, requestGB, GlobalConfigVar.FilesystemLosePercent)
```
huww98, line 1001: Remove all codes about `FilesystemLosePercent`?

huww98 (changes requested):
This is a good step forward. But I think it is even better just pass the required size to resizefs, and let it check and report error.

In case user manually resize the disk to larger size, and set PVC capacity to a smaller one, should we occupy the whole disk? (I think no, we should only occupy what is declared)

pkg/disk/nodeserver.go
mowangdk: volExpandBytes variable here is only use to generate requestGB, I think it would be fine.

pkg/disk/nodeserver.go
mowangdk: yes, I don't think It serves its purpose.

huww98:
Please also update the PR title and commit message.

PR #1009, opened 2024-03-28
disk: DescribeDisks auto batching

huww98:
#### What type of PR is this?

/kind feature
/kind cleanup

disk: introduce DiskStatusWaiter

A central place for all disks pending status changes (attaching/detaching).
Waiter will poll disk status in batch and wake up the caller if the status has changed.

The motivation is reducing OpenAPI call count, to avoid triggering throttling, thus improve system throughput.

Also include a simple fallback waiter to support multi-tanent use-case.

Shared disk support is best effort, since we cannot create new shared disks for test now.

Fixes #

```release-note
If AD-controller enabled, the number of DescribeDisks requests on a busy cluster will be reduced greatly.
```

```docs

```

### DIFF
--- a/pkg/disk/cloud.go
+++ b/pkg/disk/cloud.go
@@ -34,6 +34,7 @@ import (
 	"github.com/aliyun/alibaba-cloud-sdk-go/services/ecs"
 	"github.com/kubernetes-sigs/alibaba-cloud-csi-driver/pkg/cloud"
 	"github.com/kubernetes-sigs/alibaba-cloud-csi-driver/pkg/common"
+	"github.com/kubernetes-sigs/alibaba-cloud-csi-driver/pkg/disk/waitstatus"
 	"github.com/kubernetes-sigs/alibaba-cloud-csi-driver/pkg/utils"
 	perrors "github.com/pkg/errors"
 	"google.golang.org/grpc/codes"
@@ -62,8 +63,13 @@ const (
 	DiskMultiAttachEnabled     = "Enabled"
 )
 
+type DiskAttachDetach struct {
+	slots  AttachDetachSlots
+	waiter waitstatus.StatusWaiter[ecs.Disk]
+}
+
 // attach alibaba cloud disk
-func attachDisk(ctx context.Context, diskID, nodeID string, isSharedDisk, isSingleInstance bool, fromNode bool) (string, error) {
+func (ad *DiskAttachDetach) attachDisk(ctx context.Context, diskID, nodeID string, isSharedDisk, isSingleInstance bool, fromNode bool) (string, error) {
 	klog.Infof("AttachDisk: Starting Do AttachDisk: DiskId: %s, InstanceId: %s, Region: %v", diskID, nodeID, GlobalConfigVar.Region)
 
 	ecsClient := updateEcsClient(GlobalConfigVar.EcsClient)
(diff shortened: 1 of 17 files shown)


huww98:
I have observed that after AttachDisk returns, the status remains Available for a short period, so I'm not relying on disk.Status in this version.

huww98:
/ok-to-test
I've worked around the credential refresh issue with a hook.

huww98:
/hold
merge after #1200

pkg/disk/cloud.go
```diff
@@ -620,72 +544,35 @@ func tagDiskAsK8sAttached(diskID string, ecsClient *ecs.Client) {
 	klog.Infof("tagDiskAsK8sAttached:: add tag to disk: %s", diskID)
 }
 
-func waitForSharedDiskInStatus(ctx context.Context, retryCount int, interval time.Duration, diskID, nodeID string, expectStatus string, ecsClient *ecs.Client) error {
```
mowangdk, line 623: Do not delete the logics of sharedisk, try to make batch describe and sharedisk coexists

huww98: OK. But due to lack of testing, we cannot verify it still works.

huww98:
/unhold
I've added back best effort support for shared disk

pkg/disk/disk.go
```diff
@@ -232,21 +237,29 @@ func GlobalConfigSet(m metadata.MetadataProvider) {
...
-		GlobalConfigVar.AttachDetachSlots = NewSlots(detachConcurrency, attachConcurrency)
-	} else {
-		// if ADController is not enabled, we need serial attach to recognize old disk
-		GlobalConfigVar.AttachDetachSlots = NewSlots(1, 1)
+func newBatcher(fromNode bool) (waitstatus.StatusWaiter[ecs.Disk], batcher.Batcher[ecs.Disk]) {
+	client := desc.Disk{Client: GlobalConfigVar.EcsClient}
+	ctx := context.Background()
+	interval := 1 * time.Second
+	if fromNode {
+		interval = 2 * time.Second // We have many nodes, use longer interval to avoid throttling
```
mowangdk, line 248: We can optimize the interval based on distinct node names

pkg/disk/waitstatus/batched.go
```diff
@@ -0,0 +1,216 @@
...
+	if len(pendingReqs) == 0 {
+		delete(w.requests, id)
+		return true
+	} else {
+		w.requests[id] = pendingReqs
+		return false
+	}
+}
+
+func (w *Batched[T]) processFeedback(fb responseFeedback[T]) (next []string) {
```
mowangdk, line 147: Why not eliminate the feedback channel and incorporate the response directly into resultChan?

huww98: Because I want to avoid a mutex to protect `w.requests` and `w.idQueue`. I found it easier to  understand if I only modify `w` from one goroutine.

pkg/disk/disk.go
huww98: Sorry. I don't understand this. If batcher is initialized from csi-plugin, we cannot know how many nodes are there in the cluster. If from controller, we already have the dynamic interval mechanism from batcher to adjust the internal based on real-time load.

pkg/disk/waitstatus/batched.go
```diff
@@ -0,0 +1,216 @@
+package waitstatus
```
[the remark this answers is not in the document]
huww98: OK, I will add comments.

#1225 is using `waitstatus`. It shares several commits from this one.

pkg/disk/disk.go
mowangdk: Only from node, maybe we can use nodeName as random seed to generate interval time?

pkg/disk/waitstatus/batched.go
mowangdk: okay, gotcha, please add the explanation

PR #1010, opened 2024-03-29
feat: add featuregate

vie-serendipity:
#### What type of PR is this?

/kind feature
Currently, many functional configurations rely on environment variables. The pull request intends to introduce featuregates as a method to enable or disable these features. For now it only support the disk-ad-controller, with plans to expand to more features upon successful testing.

Fixes #

```release-note
Enable disk-ad-controller through command line parameters(--feature-gates).
```

```docs

```


pkg/disk/disk.go
```diff
@@ -240,9 +241,10 @@ func GlobalConfigSet(m metadata.MetadataProvider) *restclient.Config {
 
 	// Global Config Set
 	GlobalConfigVar = GlobalConfig{
-		Region:                metadata.MustGet(m, metadata.RegionID),
-		NodeID:                nodeID,
-		ADControllerEnable:    csiCfg.GetBool("disk-adcontroller-enable", "DISK_AD_CONTROLLER", false),
+		Region: metadata.MustGet(m, metadata.RegionID),
+		NodeID: nodeID,
+		//ADControllerEnable:    csiCfg.GetBool("disk-adcontroller-enable", "DISK_AD_CONTROLLER", false),
```
huww98, line 246: Just remove this line. We can retrieve it from git.

main.go
```diff
@@ -94,6 +95,7 @@ type globalMetricConfig struct {
 
 // Nas CSI Plugin
 func main() {
+	flag.Var(features.FunctionalMutableFeatureGate, "featuregate", "Enable or disable specific feature gates")
```
huww98, line 98: https://github.com/kubernetes-csi/external-provisioner/blob/33c5c1f8f2f78e7756969cf37c2417c339c4c216/cmd/csi-provisioner/csi-provisioner.go#L131-L132

Be consistent with other CSI sidecars?

pkg/features/features.go
```diff
@@ -0,0 +1,32 @@
...
+	DBFSMetricByPlugin     featuregate.Feature = "DBFSMetricByPlugin"
+	DBFSADController       featuregate.Feature = "DBFSADController"
+)
+
+var (
+	FunctionalMutableFeatureGate = featuregate.NewFeatureGate()
+	defaultDiskFeatureGate       = map[featuregate.Feature]featuregate.FeatureSpec{
+		DiskADController:       {Default: false, PreRelease: featuregate.Alpha},
+		DiskDetachBeforeAttach: {Default: true, PreRelease: featuregate.Alpha},
+		DiskDetachBeforeDelete: {Default: true, PreRelease: featuregate.Alpha},
```
huww98, line 21: I think these two are not suitable for feature gates. Feature gates are intended to be removed when stable. We may never remove these options.

huww98 (changes requested):

main.go
vie-serendipity: These are just the different methods of parsing input parameters. Will it cause consistency issues?

pkg/features/features.go
vie-serendipity: Ok

vie-serendipity:
> Should we keep reading from ENV, for capability to old installations?
@huww98  Just reading both from environment variables and feature gates, then do a bitwise AND operation. Do you think this approach makes sense?

main.go
huww98: my largest concern is the flag name, `feature-gates` vs `featuregate`

go.mod
```diff
@@ -123,5 +122,6 @@ replace (
 	github.com/stretchr/testify => github.com/stretchr/testify v1.3.0
 	google.golang.org/appengine => google.golang.org/appengine v1.4.0
 	gopkg.in/yaml.v2 => gopkg.in/yaml.v2 v2.2.2
+	k8s.io/component-base => k8s.io/component-base v0.23.16
```
huww98, line 125: Why replace?

huww98:
> then do a bitwise AND operation.

Do you mean OR? I think either env or feature gate should enable it.

main.go
vie-serendipity: I misunderstand, you're right. `feature-gates` is reasonable.

go.mod
vie-serendipity: Because version 0.26 of `component-base` relies on a higher version of Prometheus, I'm concerned that upgrading will introduce unexpected errors. So I find replace it with a lower version.

mowangdk: what if we directly set ```component-base``` to v0.23.16? would there be any issues?

mowangdk:
> > then do a bitwise AND operation.
> 
> Do you mean OR? I think either env or feature gate should enable it.

+1 and feature-gate should be a higher priority

vie-serendipity:
Yes, I meant OR, I typed it wrong back then. @huww98 @mowangdk

go.mod
vie-serendipity: When I set to v0.23.16, every time I execute `go mod tidy`, it will upgrade to v0.26.12. So I have to force to replace it.

huww98: I think upgrading Prometheus is fine. We have to do it sooner or later.

mowangdk:
@vie-serendipity please solve the conflict

PR #1013, opened 2024-04-04
disk: fix and cleanup NodeExpandVolume

huww98:
#### What type of PR is this?

/kind bug
/kind cleanup

fix NodeExpandVolume always return error

remove unnessesary statfs
remove float math
remove the unused parameter FILE_SYSTEM_LOSE_PERCENT
optimize error message
fix false error if req.CapacityRange is not specified

Fixes #1008 
Fixes 377c9e9bce468c996426da502fc1e157a13ac5f7

```release-note
FILE_SYSTEM_LOSE_PERCENT env is now deprecated and not used.
```

```docs

```

### DIFF
--- a/pkg/disk/nodeserver.go
+++ b/pkg/disk/nodeserver.go
@@ -23,7 +23,6 @@ import (
 	"errors"
 	"fmt"
 	"io/fs"
-	"math"
 	"os"
 	"os/exec"
 	"path/filepath"
@@ -40,6 +39,7 @@ import (
 	"google.golang.org/grpc/codes"
 	"google.golang.org/grpc/status"
 	v1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/api/resource"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/client-go/kubernetes"
@@ -92,8 +92,6 @@ const (
 	RunvRunTimeMode = "runv"
 	// InputOutputErr tag
 	InputOutputErr = "input/output error"
-	// FileSystemLoseCapacityPercent is the env of container
-	FileSystemLoseCapacityPercent = "FILE_SYSTEM_LOSE_PERCENT"
 	// DiskMultiTenantEnable Enable disk multi-tenant mode
 	DiskMultiTenantEnable = "DISK_MULTI_TENANT_ENABLE"
 	// TenantUserUID tag
@@ -903,8 +901,7 @@ func (ns *nodeServer) NodeExpandVolume(ctx context.Context, req *csi.NodeExpandV
 	*csi.NodeExpandVolumeResponse, error) {
 	log.Infof("NodeExpandVolume: node expand volume: %v", req)
 
-	volExpandBytes := float64(req.GetCapacityRange().GetRequiredBytes())
-	requestGB := math.Floor(volExpandBytes / float64(1024*1024*1024))
+	requestBytes := req.GetCapacityRange().GetRequiredBytes()
 
 	volumePath := req.GetVolumePath()
 	diskID := req.GetVolumeId()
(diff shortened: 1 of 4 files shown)


pkg/disk/nodeserver.go
```diff
@@ -937,16 +933,6 @@ func (ns *nodeServer) NodeExpandVolume(ctx context.Context, req *csi.NodeExpandV
 	}
 	devicePath := ChooseDevice(rootDevice, subDevice)
 
-	log.Infof("NodeExpandVolume:: volumeId: %s, devicePath: %s, volumePath: %s", diskID, devicePaths, volumePath)
```
mowangdk, line 940: Keep this log.

pkg/disk/nodeserver.go
```diff
@@ -959,7 +945,8 @@ func (ns *nodeServer) NodeExpandVolume(ctx context.Context, req *csi.NodeExpandV
 				if bytes.Contains(output, []byte("it cannot be grown")) || bytes.Contains(output, []byte("could only be grown by")) {
 					deviceCapacity := getBlockDeviceCapacity(devicePath)
 					rootCapacity := getBlockDeviceCapacity(rootPath)
-					log.Infof("NodeExpandVolume: Volume %s with Device Partition %s no need to grown, with requestSize: %v, rootBlockSize: %v, partition BlockDevice size: %v", diskID, devicePath, requestGB, rootCapacity, deviceCapacity)
+					log.Infof("NodeExpandVolume: Volume %s with Device Partition %s no need to grown, with requestSize: %v, rootBlockSize: %v, partition BlockDevice size: %v",
+						diskID, devicePath, requestBytes, rootCapacity, deviceCapacity)
```
mowangdk, line 949: Change the log capacity units to GB, for better readability.

pkg/disk/nodeserver.go
```diff
@@ -988,21 +975,18 @@ func (ns *nodeServer) NodeExpandVolume(ctx context.Context, req *csi.NodeExpandV
...
-		"NodeExpandVolume:: filesystem resize context device capacity: %v, before resize fs capacity: %v resize fs capacity: %v, requestGB: %v. file system lose percent: %v",
-		deviceCapacity, beforeResizeDiskCapacity, diskCapacity, requestGB, GlobalConfigVar.FilesystemLosePercent)
-	return &csi.NodeExpandVolumeResponse{}, nil
+	if requestBytes > 0 && deviceCapacity < requestBytes {
+		// After calling OpenAPI to expansion cloud disk, the size of the underlying block device may not change immediately.
+		// return error and CO will retry later.
+		return nil, status.Errorf(codes.Aborted, "requested %v, but actual block size %v is smaller. Not updated yet?",
+			resource.NewQuantity(requestBytes, resource.BinarySI), resource.NewQuantity(deviceCapacity, resource.BinarySI))
+	}
+	log.Infof("NodeExpandVolume:: Expand %s to %dB successful", diskID, deviceCapacity)
```
mowangdk, line 986: same as above

pkg/disk/nodeserver.go
huww98: How about use hex? I used to see a lot of hex in ceph log. Once we need to read the log, we value simple and precise than readability for diagnosis purpose. And hex can be converted to different unit faster by hand. e.g.:

Decimal: 22014345216
GB: 20.502456665039062
Hex: 0x520284000
resource: 21498384Ki

mowangdk: hex may be more versatile, but it's basically useless in a container scenario, and you can check that logs in kubernetes repo basically don't use hex.  keep using GiB would be fine. Accuracy issue could be ignore

huww98: > you can check that logs in kubernetes repo basically don't use hex

Yes, but they also don't use floating point numbers. I'm searching for regex `Infof.+size`, all of them using either decimal bytes, or resource. Resource is not convenient when comparing different sizes in the log, so decimal bytes should be the way to go if we really want to follow Kubernetes.

huww98 (commented):
Resolved conflict on import

PR #1028, opened 2024-04-19
disk: use xattr to check if the disk is managed by us

huww98:
#### What type of PR is this?

/kind bug

to fix the issue that when DISK_AD_CONTROLLER is enabled and disk is detaching, MaxVolumesPerNode may be smaller than expected

Fixes #

```release-note
fixed that when DISK_AD_CONTROLLER is enabled and disk is detaching, MaxVolumesPerNode may be smaller than expected
```

```docs

```

### DIFF
--- a/pkg/disk/bdf.go
+++ b/pkg/disk/bdf.go
@@ -10,6 +10,7 @@ import (
 	"os"
 	"os/exec"
 	"path/filepath"
+	"strconv"
 	"strings"
 	"sync"
 	"syscall"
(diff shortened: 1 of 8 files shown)


huww98:
https://github.com/torvalds/linux/commit/b09e0fa4b4ea66266058eead43350bd7d55fec67

This tmpfs xattr support is added in Linux Kernel 3.0

huww98:
I've made some changes to the design:
- The xattr is now added at NodeUnstageVolume, because:
  - this also works for volumes attached before CSI upgraded
  - we may switch drivers for the disk for some use-case, and the xattr may be lost.
- I now iterate all the `/dev` to find xattr, because some disk may lack of serial number, and cannot be found by link.
- Detaching disks are now excluded from `attachedDisks`, because we can't know whether it's device in /dev is present. Even if it is really an unmanaged disk, it is going to be detached, so it does not matter.
- To avoid race condition, the informations are now retrieved in the order of: xattr -> OpenAPI -> Node status. This is the only valid order from my analyses.

PR #1040, opened 2024-04-25
feat: tag disk when creating static volume

vie-serendipity:
#### What type of PR is this?

/kind feature
When 'diskid' is present in the annotations and volume can be created statically, add custom tags and some default tags to the disk.

Fixes #

```release-note
NONE
```

```docs

```

### DIFF
--- a/pkg/disk/cloud.go
+++ b/pkg/disk/cloud.go
@@ -537,7 +537,7 @@ func tagDiskUserTags(diskID string, tags map[string]string, tenantUID string) {
 		log.Warnf("tagDiskUserTags: AddTags error: %s, %s", diskID, err.Error())
 		return
 	}
-	log.Info("tagDiskUserTags: Success to tag disk %s", diskID)
+	log.Infof("tagDiskUserTags: Success to tag disk %s", diskID)
 }
 
 // tag disk with: k8s.aliyun.com=true
--- a/pkg/disk/cloud.go
+++ b/pkg/disk/cloud.go
@@ -521,6 +521,24 @@ func getDisk(diskID string, ecsClient *ecs.Client) []ecs.Disk {
 	return diskResponse.Disks.Disk
 }
 
+func tagDiskUserTags(diskID string, ecsClient *ecs.Client, tags map[string]string) {
+	addTagsReq := ecs.CreateAddTagsRequest()
+	var userTags []ecs.AddTagsTag
+	for k, v := range tags {
+		userTags = append(userTags, ecs.AddTagsTag{Key: k, Value: v})
+	}
+	addTagsReq.Tag = &userTags
+	addTagsReq.ResourceType = "disk"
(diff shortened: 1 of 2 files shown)


pkg/disk/cloud.go
```diff
@@ -521,6 +521,24 @@ func getDisk(diskID string, ecsClient *ecs.Client) []ecs.Disk {
...
+	addTagsReq := ecs.CreateAddTagsRequest()
+	var userTags []ecs.AddTagsTag
+	for k, v := range tags {
+		userTags = append(userTags, ecs.AddTagsTag{Key: k, Value: v})
+	}
+	addTagsReq.Tag = &userTags
+	addTagsReq.ResourceType = "disk"
+	addTagsReq.ResourceId = diskID
+	addTagsReq.RegionId = GlobalConfigVar.Region
+	_, err := ecsClient.AddTags(addTagsReq)
```
huww98, line 534: Need to call `getEcsClientByID` before calling ECS OpenAPI, to support multi-tenant use case, or refresh the credential.

pkg/disk/cloud.go
```diff
@@ -521,6 +521,24 @@ func getDisk(diskID string, ecsClient *ecs.Client) []ecs.Disk {
...
+	if err != nil {
+		log.Warnf("tagDiskUserTags: AddTags error: %s, %s", diskID, err.Error())
+		return
```
huww98, line 537: This error is otherwise ignored. Is this intentional? Should the failure block PV creating?

pkg/disk/cloud.go
```diff
@@ -521,6 +521,24 @@ func getDisk(diskID string, ecsClient *ecs.Client) []ecs.Disk {
 	return diskResponse.Disks.Disk
 }
 
+func tagDiskUserTags(diskID string, ecsClient *ecs.Client, tags map[string]string) {
+	addTagsReq := ecs.CreateAddTagsRequest()
+	var userTags []ecs.AddTagsTag
```
huww98, line 526: use `make([]ecs.AddTagsTag, 0, len(tags))` should be more efficient.

vie-serendipity: okay

pkg/disk/cloud.go
vie-serendipity: Normally, a failed tag shouldn't impact the PV. I'm just referencing when `tagDiskAsK8sAttached` failed while attaching, it's ignored too. @mowangdk

pkg/disk/cloud.go
vie-serendipity: okay

pkg/disk/cloud.go
mowangdk: yep, failed tag has no impact to PV. we should keep the original logic

mowangdk:
/lgtm
/approve

PR #1044, opened 2024-04-28
Metrics cleanup

huww98:
#### What type of PR is this?

/kind cleanup

#### What this PR does / why we need it:

remove the dependency on github.com/emirpasic/gods/sets/hashset
remove old globalmount path support.

#### Which issue(s) this PR fixes:

Fixes #

#### Does this PR introduce a user-facing change?

```release-note
NONE
```

#### Additional documentation e.g., KEPs (Kubernetes Enhancement Proposals), usage docs, etc.:

```docs

```

PR #1046, opened 2024-04-29
metric: disk hang alert and some cleanup

huww98:
#### What type of PR is this?

/kind cleanup
/kind feature

##### metric/disk: refactor and add more reliable I/O hang alert

This add back I/O hang alert removed by commit 656877777b5dd98a8d8e3c9594cdc1e7b0306d75.

The previous implementation is prone to false positive if an IO is started just before we collect stats.
In this version, we ensure there is no progress for at least 30s before alert.

This patch also switch to prometheus package for parsing /proc/diskstats, and avoids hard codes magic indices everywhere.

Fix the metrics `{read,write}_time_milliseconds_total` to be really milliseconds (they are in seconds before). Also fix the metric help text.

Fix that metrics `{capacity_bytes,inodes}_{available,total,used}` should be gauge instead of counter.

##### metric/disk: use pvc ref from pv in event

should reduce the number of args, and also pass along PVC UID

##### metric/disk: remove unnessesary find disk logic

just get the device number from mountpoint

Fixes #

The first 4 commits are already in #1044, please start reviewing from the 5th commit.

/hold
merge this after #1044

```release-note
Disk IOHang event is added back.
{read,write}_time_milliseconds_total metrics changed to use millisecond as unit.
{capacity_bytes,inodes}_{available,total,used}` metrics changed to gauge
```

```docs

```

### DIFF
--- a/pkg/metric/common.go
+++ b/pkg/metric/common.go
@@ -42,19 +42,10 @@ const (
 	diskDefaultsLantencyThreshold           = 10
 	diskDefaultsCapacityPercentageThreshold = 85
 	nfsDefaultsCapacityPercentageThreshold  = 85
-	float64EqualityThreshold                = 1e-9
-	diskStatsFileName                       = "diskstats"
 	nfsStatsFileName                        = "/proc/self/mountstats"
 	latencyTooHigh                          = "LatencyTooHigh"
 	capacityNotEnough                       = "NotEnoughDiskSpace"
 	ioHang                                  = "IOHang"
-)
-
-const (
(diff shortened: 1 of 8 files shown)


huww98:
In this version, I split the first commit into two, for easier review.
I also added support for old kernel, based on the observation that the `IOsTotalTicks` may not increase during IO hung.

huww98:
/unhold
https://github.com/kubernetes-sigs/alibaba-cloud-csi-driver/pull/1044 already merged

PR #1047, opened 2024-04-30
helm: nas driver no longer needs /mnt

huww98:
#### What type of PR is this?

/kind cleanup

remove hostPath /mnt from controller. It is no longer needed since 33875a5bf3 (refactor nas plugin and support accesspoint).

local driver still needs this for quota path.

Fixes #

/cc @tydra-wang 

```release-note
NONE
```

```docs

```

### DIFF
--- a/deploy/chart/templates/controller.yaml
+++ b/deploy/chart/templates/controller.yaml
@@ -336,9 +336,6 @@ spec:
               name: addon-token
               readOnly: true
 {{- end }}
-            - mountPath: /mnt
-              mountPropagation: Bidirectional
-              name: host-mnt
             - mountPath: /host/etc
               name: etc
           resources:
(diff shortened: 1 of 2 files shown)


mowangdk:
@tydra-wang ptal

tydra-wang:
The NAS driver stores loopdevice image files under `/mnt/nasplugin.alibabacloud.com` when using the `losetup` mode. However this feature is not enabled by default.

huww98:
We have a case where the user have his own mountpoint on /mnt on the host. /mnt is so common that I think we'd better not use it for our purpose.

huww98:
> Yes, but I suppose losetup would still work even if /mnt is not a hostPath?

This will cause problem when CSI plugin restarts.

But I still think this change is desired. Users who want to enable losetup should add the mount themselves.

PR #1049, opened 2024-05-03
disk/device_manager: replace syscall with unix package

huww98:
#### What type of PR is this?

/kind cleanup

docs of syscall recommends the unix package. And latter enables simpler code.

Fixes #

```release-note
NONE
```

```docs

```

### DIFF
--- a/pkg/disk/device_manager.go
+++ b/pkg/disk/device_manager.go
@@ -6,7 +6,6 @@ import (
 	"os"
 	"path/filepath"
 	"strings"
-	"syscall"
 
 	"github.com/kubernetes-sigs/alibaba-cloud-csi-driver/pkg/utils"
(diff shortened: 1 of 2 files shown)


mowangdk:
Can you paste the syscall doc link on this PR

huww98:
https://pkg.go.dev/syscall

> NOTE: Most of the functions, types, and constants defined in this package are also available in the [golang.org/x/sys](https://pkg.go.dev/golang.org/x/sys) package. That package has more system call support than this one, and most new code should prefer that package where possible.

huww98:
There is one more reason leading to these changes: `unix.Major(rdev), unix.Minor(rdev)` do not have corresponding alternatives in `syscall`, which are in fact a lot more complex than I thought.

huww98:
rebased to resolve compile error.

PR #1054, opened 2024-05-08
redirect logs to klog

huww98:
#### What type of PR is this?

/kind feature

to be consistent with other k8s projects.

We should migrate to structured and contextual logging gradually, as the k/k repo does.

Fixes #

```release-note
log format is switched to klog
```

https://github.com/kubernetes/community/blob/master/contributors/devel/sig-instrumentation/migration-to-structured-logging.md

```docs

```

### DIFF
--- a/main.go
+++ b/main.go
@@ -34,6 +34,7 @@ import (
 	"github.com/kubernetes-sigs/alibaba-cloud-csi-driver/pkg/disk"
 	"github.com/kubernetes-sigs/alibaba-cloud-csi-driver/pkg/ens"
 	"github.com/kubernetes-sigs/alibaba-cloud-csi-driver/pkg/features"
+	csilog "github.com/kubernetes-sigs/alibaba-cloud-csi-driver/pkg/log"
 	"github.com/kubernetes-sigs/alibaba-cloud-csi-driver/pkg/metric"
 	"github.com/kubernetes-sigs/alibaba-cloud-csi-driver/pkg/nas"
 	"github.com/kubernetes-sigs/alibaba-cloud-csi-driver/pkg/om"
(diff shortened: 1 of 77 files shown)


huww98:
Added a benchmark. klog is more efficient than logrus. But the redirection causes about 3x overhead. Mostly caused by stack unwind, the call stack between logrus and its hook is too deep.

```
goos: linux
goarch: amd64
pkg: github.com/kubernetes-sigs/alibaba-cloud-csi-driver/pkg/log
cpu: Intel(R) Xeon(R) Platinum
BenchmarkLogrus-4                       	  497246	      2421 ns/op
BenchmarkLogrusStructured-4             	  338302	      3603 ns/op
BenchmarkKlog-4                         	  661293	      1652 ns/op
BenchmarkKlogStructured-4               	  387448	      2936 ns/op
BenchmarkLogrusRedirected-4             	  110848	     10740 ns/op
BenchmarkLogrusStructuredRedirected-4   	  120429	     10234 ns/op
PASS
```
image

We can cut down the overhead by half if we don't want useful backtrace for redirected logs
```
BenchmarkLogrusRedirected-4             	  330480	      3520 ns/op
BenchmarkLogrusStructuredRedirected-4   	  259669	      4505 ns/op
```

main.go
```diff
@@ -95,9 +100,29 @@ type globalMetricConfig struct {
 
 // Nas CSI Plugin
 func main() {
+	csilog.RedirectLogrusToLogr(logrus.StandardLogger(), klog.Background())
```
mowangdk, line 103: Given that all callers have been switched to klog, what purpose does this function serve?

main.go
```diff
@@ -95,9 +100,29 @@ type globalMetricConfig struct {
...
+
 	flag.Var(features.FunctionalMutableFeatureGate, "feature-gates", "A set of key=value pairs that describe feature gates for alpha/experimental features. "+
 		"Options are:\n"+strings.Join(features.FunctionalMutableFeatureGate.KnownFeatures(), "\n"))
 	flag.Parse()
+
+	// https://github.com/kubernetes/klog/blob/main/examples/coexist_glog/coexist_glog.go
+	// replace this when we remove csi-common
```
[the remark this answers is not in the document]
huww98: Yes. I will resolve this one when rebasing.

main.go
mowangdk: Fine, Is there a possibility we can remove this in the future?

PR #1056, opened 2024-05-10
feat: add helm deploy for edge

vie-serendipity:
#### What type of PR is this?

/kind feature
Deploy multi daemonsets using values-nodepools according to affinity. And csi config will overwrite the original config.

Fixes #

```release-note

```

```docs

```


deploy/chart/templates/plugin.yaml
```diff
@@ -1,17 +1,20 @@
...
+{{- range .Values.nodepools -}}
+{{- if $.Values.plugin.enabled }}
+{{- $copy := deepCopy .csi }}
+{{- $csi := mergeOverwrite $.Values.csi $copy }}
 kind: DaemonSet
 apiVersion: apps/v1
 metadata:
-  name: csi-plugin
-  namespace: {{ .Release.Namespace }}
+  name: {{ .name }}
```
huww98, line 8: Can we just use the key in .Values.nodepools as the name? e.g. 
```helm
{{ if eq $key "default" }}
name: csi-plugin
{{ else }}
name: csi-pluign-{{ $key }}
{{ end }}
```
We can avoid possible duplicated name in this way.

deploy/chart/templates/plugin.yaml
```diff
@@ -1,17 +1,20 @@
...
+  namespace: {{ $.Release.Namespace }}
 spec:
   selector:
     matchLabels:
-      app: csi-plugin
+      app: {{ .name }}
```
huww98, line 13: I suggest use two labels
```
app: csi-plugin
nodepool: {{ $.key }}
```
so that we can select all the csi-plugin pods with kubectl while debugging.

deploy/chart/templates/plugin.yaml
```diff
@@ -1,17 +1,20 @@
-{{- if .Values.plugin.enabled -}}
+{{- range .Values.nodepools -}}
+{{- if $.Values.plugin.enabled }}
+{{- $copy := deepCopy .csi }}
+{{- $csi := mergeOverwrite $.Values.csi $copy }}
```
huww98, line 4:
```suggestion
{{- range $key, $nodepool := .Values.nodepools -}}
{{- if and $.Values.plugin.enabled (ne $nodepool nil) }}
{{- $nodepool := merge $nodepool $.Values }}
```
* I want to support configs other than `.csi`, e.g. cloud nodes may use VPC registry URL, but edge node must use Internet URL.
* Allow user to set a node pool to nil to disable it.
* Merge values into nodepool, so that we don't need deep copy.

huww98:
values-edge.yaml is the config we use in our ACK Edge cluster. I think we'd better keep its name.

deploy/chart/templates/plugin.yaml
vie-serendipity: I personally think it's more readable to explicitly clarify names of each node pool. It makes no real difference compared to using $key as we do now.

deploy/chart/templates/plugin.yaml
vie-serendipity: It makes sense.

deploy/chart/templates/plugin.yaml
huww98: OK. But this way we lost a level of abstraction. e.g. `kubectl get pods -l nodepool=edge` seems more reasonable than `-l nodepool=csi-plugin-edge`

deploy/chart/templates/plugin.yaml
vie-serendipity: It‘s a useful advice to merge values into nodepool directly so that we can rewrite any field of original values.
> Allow user to set a node pool to nil to disable it.
I'm confused about this suggestion. If users want to disable a nodepool, just delete it from nodepools, why setting it to nil?

huww98: Because it is straightforward to write `helm install --set nodepools.default=nil`. But not clear how to delete it from CLI.

deploy/chart/templates/plugin.yaml
```diff
@@ -46,9 +47,9 @@ spec:
       dnsPolicy: ClusterFirst
       initContainers:
         - name: init
-          image: {{ include "workerImageSpec" (list $.Values.images "pluginInit") }}
+          image: {{ include "workerImageSpec" (list $nodepool.images "pluginInit") }}
```
huww98, line 50: I think you can still omit these `$nodepool`

deploy/chart/templates/plugin.yaml
```diff
@@ -1,17 +1,21 @@
-{{- if .Values.plugin.enabled -}}
+{{- range $key, $nodepool := .Values.nodepools -}}
+{{- if and $.Values.plugin.enabled (ne $nodepool nil) }}
+{{- $nodepool := mergeOverwrite $.Values $nodepool }}
```
huww98, line 3: I think this is still modifying `$.Values`, which is not desired. Why not `merge $nodepool $.Values`?

vie-serendipity: When I use merge, I get an unexpected result which is not same as official doc. So I choose to use mergeOverwrite.

vie-serendipity: So if I use mergeOverwrite, It's still necessary to deepCopy `.Values`.

huww98: OK, so there is some unexpected behavior here https://github.com/Masterminds/sprig/issues/255
Please proceed with deepCopy + mergeOverwrite

deploy/chart/templates/plugin.yaml
```diff
@@ -24,6 +28,15 @@ spec:
...
                 values:
                 - virtual-kubelet
+{{- if $nodepool.affinity }}
+            - matchExpressions:
+              - key: {{ $nodepool.affinity.key }}
+                operator: {{ $nodepool.affinity.operator }}
+                values:
+                {{- range $nodepool.affinity.values }}
+                - {{ . | quote }}
+                {{- end -}}
```
huww98, line 38:
```
            nodeSelectorTerms: {{- toYaml $nodepool.nodeSelectorTerms | nindent 12 }}
```
Node pool selector should goes into the same `matchExpressions` as virtual-kubelet, so that they can be ANDed.
I think we can just take the hole nodeSelectorTerms from values. And move the `virtual-kubelet` things to default node pool.

deploy/chart/values-edge.yaml
```diff
@@ -1,11 +1,29 @@
...
+      local:
+        enabled: true
+      oss:
+        controller:
+          enabled: false
+    affinity:
+      key: alibabacloud.com/is-edge-worker
+      operator: NotIn
+      values:
+        val1: "false"
```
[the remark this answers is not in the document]
vie-serendipity: done

deploy/chart/templates/plugin.yaml
vie-serendipity: done

deploy/chart/values-edge.yaml
```diff
@@ -1,11 +1,33 @@
...
+          operator: In
+          values:
+            - "false"
+  edge:
+    pluginname: csi-plugin-edge
+    csi:
+      disk:
+        enabled: false
+      local:
+        enabled: true
```
huww98, line 23: We should move this to the top level `csi`, because controller only looking for that, and we need local plugin controller.

deploy/chart/values-edge.yaml
```diff
@@ -1,11 +1,33 @@
 deploy:
   ecs: false
```
huww98, line 2: This can be moved to edge nodepool

deploy/chart/values-edge.yaml
```diff
@@ -1,11 +1,33 @@
...
+  edge:
+    pluginname: csi-plugin-edge
+    csi:
+      disk:
+        enabled: false
+      local:
+        enabled: true
+      oss:
+        controller:
+          enabled: false
```
huww98, line 26: The same here, move to top level. Configuring controller in nodepool makes no sense.

deploy/chart/values-edge.yaml
```diff
@@ -1,11 +1,33 @@
...
+        - key: type
+          operator: NotIn
+          values:
+            - virtual-kubelet
+        - key: alibabacloud.com/is-edge-worker
+          operator: In
+          values:
+            - "false"
+  edge:
+    pluginname: csi-plugin-edge
```
huww98, line 18: For capability with existing edge cluster, should we swap the name? that is, name edge daemonset `csi-plugin`, and default daemonset `csi-plugin-cloud`

deploy/chart/values-edge.yaml
vie-serendipity: why original local.controller equals to true, it seems to make no sense. how about set it to false? 
`{{- if and .Values.csi.local.enabled .Values.csi.local.controller.enabled }}`

deploy/chart/values-edge.yaml
vie-serendipity: okay

deploy/chart/values-edge.yaml
vie-serendipity: okay

deploy/chart/values-edge.yaml
huww98: Because `.csi.local.controller.enabled` is introduced later. And we should maintain capability to old configs, both for users who enabled or disabled local driver.

deploy/chart/values-edge.yaml
```diff
@@ -1,11 +1,38 @@
...
+# deployment
 csi:
-  disk:
-    enabled: false
   local:
     enabled: true
   oss:
     controller:
-      enabled: false
+      enabled: true
```
huww98, line 7:
```suggestion
      enabled: false
```

deploy/chart/values-edge.yaml
huww98: Now you moved it to top level, remove this one?

deploy/chart/values-edge.yaml
vie-serendipity: sorry for this minor mistake.

deploy/chart/templates/plugin.yaml
```diff
@@ -339,4 +340,6 @@ spec:
     rollingUpdate:
       maxUnavailable: 20%
     type: RollingUpdate
+---
```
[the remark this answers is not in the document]
vie-serendipity: because range will produce two daemonset yaml. So I think using --- to separate two resources is necessary. Otherwise,

deploy/chart/values-edge.yaml
vie-serendipity: ok

huww98 (approved):

huww98 (changes requested):
Please rebase and apply these change
```diff
diff --git a/deploy/chart/templates/plugin.yaml b/deploy/chart/templates/plugin.yaml
index 2ee66ad7..3092b4dd 100644
--- a/deploy/chart/templates/plugin.yaml
+++ b/deploy/chart/templates/plugin.yaml
@@ -148,7 +148,7 @@ spec:
               mountPath: /var/lib/kubelet/
               mountPropagation: "Bidirectional"
 {{- range $key := tuple "disk" "nas" "oss" }}
-  {{- with index $.Values.csi $key -}}
+  {{- with index $nodePool.csi $key -}}
     {{- if .enabled }}
             - name: {{ $key }}-plugin-dir
               mountPath: /csi/{{ $key }}plugin.csi.alibabacloud.com
@@ -301,7 +301,7 @@ spec:
           hostPath:
             path: /var/lib/kubelet
             type: Directory
-{{- range $key, $val := .Values.csi }}
+{{- range $key, $val := $nodePool.csi }}
 {{- if $val.enabled }}
         - name: {{ $key }}-plugin-dir
           hostPath:
```

huww98 (approved):

PR #994, opened 2024-03-07
[WIP] support disk group volumesnapshot

AlbeeSo:
#### What type of PR is this?
/kind feature

#### Which issue(s) this PR fixes:

Fixes #

#### Does this PR introduce a user-facing change?

```release-note

```

#### Additional documentation e.g., KEPs (Kubernetes Enhancement Proposals), usage docs, etc.:

```docs

```

PR #1060, opened 2024-05-15
dbfs&oss: return unimplemented error instead

AlbeeSo:
https://github.com/kubernetes-sigs/alibaba-cloud-csi-driver/pull/994#discussion_r1515810514
as discussion here, it's better to return unimplemented error if api is not support.

Fixes #

```release-note

```

```docs

```


huww98:
/remove-lifecycle rotten

This is still valuable. But I think we should now just remove these useless functions since #1144 has been merged. Returning unimplemented is the default behavior if you don't implement them.

huww98:
#1241 should supersede this

/close

PR #1063, opened 2024-05-21
helm: less dependent on /var/lib/kubelet

huww98:
#### What type of PR is this?

/kind bug
/kind cleanup

Removed all unnecessary occurrence of /var/lib/kubelet, controller should not depends on this.

Registrars don't need the whole kubelet-dir

Fixes local plugin crash if /var/lib/kubelet/csi-plugins/localplugin.csi.alibabacloud.com does not exist in advance.

Fixes #

```release-note
NONE
```

```docs

```

### DIFF
--- a/deploy/chart/templates/controller.yaml
+++ b/deploy/chart/templates/controller.yaml
@@ -79,7 +79,7 @@ spec:
               cpu: 500m
               memory: 1024Mi
           args:
-            - "--csi-address=$(ADDRESS)"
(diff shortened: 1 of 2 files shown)


huww98:
Fix this:

F0521 08:29:13.511328   80898 server.go:89] Failed to listen: listen unix /var/lib/kubelet/csi-plugins/localplugin.csi.alibabacloud.com/csi.sock: bind: no such file or directory

PR #1067, opened 2024-05-27
chore: reduce error logs from metrics collector

tydra-wang:
If storage-monitor is not installed, the NAS metrics collector could raise a lot of "services not found" errors.

#### What type of PR is this?

/kind cleanup

Fixes #

```release-note

```

```docs

```


huww98:
How about at least log it once so that the cluster admin will have clues about why their NAS metrics are not available?

tydra-wang:
> How about at least log it once so that the cluster admin will have clues about why their NAS metrics are not available?

Changed to discard errors that are duplicated in a limited duration.

huww98:
That looks better. How about logging the duplicated errors at debug level? `log.StandardLogger().Logf()` accepts a level as argument.

huww98:
Please rebase the latest master for CI to pass

PR #1072, opened 2024-06-02
combine the init script of different arch into one

huww98:
they are very similar, combine them to reduce maintenance work in the future.

#### What type of PR is this?

/kind cleanup

Fixes #

```release-note
NONE
```

```docs

```


huww98:
I have confirmed with @tydra-wang that this `yum install -y fuse-devel` is not necessary for ARM. So removed in this PR.

tydra-wang (approved):

PR #1078, opened 2024-06-05
Add support for ACK-one register clusters

tydra-wang:
#### What type of PR is this?

/kind feature

Add Helm values for the ACK Register Clusters to enable the use of NAS/OSS on the nodes of the register cluster.

Fixes #

```release-note

```

```docs

```


deploy/chart/values-external.yaml
```diff
@@ -0,0 +1,56 @@
...
+controller:
+  nodeAffinity:
+    requiredDuringSchedulingIgnoredDuringExecution:
+      nodeSelectorTerms:
+      - matchExpressions:
+        - key: type
+          operator: NotIn
+          values:
+          - virtual-kubelet
+        # csi-provisioner should be scheduled to on-cloud nodes
```
huww98, line 14: Do all register clusters have at least one on-cloud node?

tydra-wang: If no on-cloud nodes exist, they don't need to install csi-provisioner.

huww98: OK, so they can just install csi-plugin from ACK console.

deploy/chart/values.yaml
```diff
@@ -46,10 +53,13 @@ deploy:
...
   regionID: null  # useful when deploying on non-ECS, but need to access Alibaba Cloud OpenAPI
 
-  # To access Alibaba Cloud OpenAPI in self-deployed Kubernetes cluster, use a secret with "id" and "secret" set.
+  # To access Alibaba Cloud OpenAPI in self-deployed Kubernetes cluster, use a Secret with "id" and "secret" set.
+  # Customize the keys in the Secret by setting the idField and secretField.
   accessKey:
     enabled: false
     secretName: csi-access-key
+    idField: id
+    secretField: secret
```
huww98, line 62:
```suggestion
    idKey: id
    secretKey: secret
```
as it is used in `secretKeyRef.key`

deploy/chart/values.yaml
```diff
@@ -37,6 +37,13 @@ defaultStorageClass:
   enabled: true
 
 deploy:
+  # By default, the controller server does not delete NAS subpaths when DeleteVolume, that means
+  # the files created in the volume will remain in the filesystem after PV deleted.
+  # You can set "enableNasSubpathFinalizer" to true in ACK clusters with storage-operator addon installed, 
+  # then the controller server will patch a finalzier on NAS subpath-type PV when DeleteVolume 
+  # and the storage-operator will handle the cleanup of the subpath in NAS filesystem.
+  enableNasSubpathFinalizer: false
```
huww98, line 45: how about just name it `deploy.withStorageController`? And default it to `deploy.ack`.

deploy/chart/values-external.yaml
```diff
@@ -0,0 +1,65 @@
+# Values for register clusters
+# NOTE: on-cloud nodes created by register clusters will have alibabacloud.com/external=true label
+
+deploy:
+  accessKey:
+    enabled: true
+    secretName: alibaba-addon-secret
+    idField: access-key-id
+    secretField: access-key-secret
```
huww98, line 9: Maybe we should also add yet another value `deploy.addonToken`, default to `deploy.ack`, and override it to false on external cluster.

huww98 (changes requested):

deploy/chart/values-external.yaml
huww98: Or just disable addonToken automatically if accessKey is enabled?

deploy/chart/values.yaml
huww98: We should not disable it here, or we will break existing ACK installations.

deploy/chart/values-external.yaml
tydra-wang: I don't think there is a problem, because the addon token volume is set as optional. csi will prioritize to use credentials from env variables.

deploy/chart/values.yaml
tydra-wang: Users would better install csi from ACK addons, where we can set the dependencies on. Using the chart directly with this enabled, users could be confused when their PV stuck in terminating if storage-controller not working as expect.

huww98: But the manifests we use in ACK addon will be generated from this `values.yaml`. Or should we introduce another values-ack.yaml?

huww98:
I think we should disable addonToken automatically if accessKey is enabled. But that is an enhancement, and can be deferred to another PR.

/lgtm

deploy/chart/templates/controller.yaml
```diff
@@ -300,8 +304,10 @@ spec:
             - name: PRIVATE_CLOUD_TAG
               value: "true"
 {{- end }}
+{{- if .Values.deploy.withStorageController }}
```
mowangdk, line 307: Is this environment enabled in the ACK cluster? I don't remember adding this configuration.

tydra-wang: Yes, current template of ACK csi-provisioner addon contains this too.

PR #1081, opened 2024-06-07
CGroup V2 support

huww98:
#### What type of PR is this?

/kind feature

Fixes #

```release-note
IO limits features now supports CGroup V2
```

```docs

```

### DIFF
--- a/pkg/utils/cgroup/pod.go
+++ b/pkg/utils/cgroup/pod.go
@@ -1,0 +1,72 @@
+package cgroup
+
+import (
+	"errors"
+	"fmt"
+	"io/fs"
+	"os"
+	"path/filepath"
+	"strings"
+
+	utilsio "github.com/kubernetes-sigs/alibaba-cloud-csi-driver/pkg/utils/io"
+	log "github.com/sirupsen/logrus"
+	"golang.org/x/sys/unix"
+)
+
+type PodCGroupSlice struct {
+	KubepodsSlicePath string
+}
+
+func (cg PodCGroupSlice) FindPodDir(podUID string) (int, error) {
+	if cg.KubepodsSlicePath == "" {
+		return -1, errors.New("no cgroup detected")
+	}
+
+	// /sys/fs/cgroup/blkio/kubepods.slice/kubepods-besteffort.slice/kubepods-besteffort-podaadcc749_6776_4933_990d_d50f260f5d46.slice/blkio.throttle.write_bps_device
+	podUID = strings.ReplaceAll(podUID, "-", "_")
+	for _, p := range []string{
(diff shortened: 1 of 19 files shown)


huww98:
```
MountVolume.SetUp failed for volume "cgroupv2" : mkdir /sys/fs/cgroup/kubepods.slice: read-only file system
```

OK, So we cannot mount this path on cgroup v1 system. I think the only choice is to mount the whole `/sys/fs/cgroup`.

pkg/utils/io.go
```diff
@@ -3,65 +3,93 @@ package utils
...
 
-const PodBlkIOCgroupPath = "/host/sys/fs/cgroup/blkio/kubepods.slice"
+type PodCGroup struct {
+	slice cgroup.PodCGroupSlice
+	io    cgroup.IO
+	dev   utilsio.DevTmpFS
+}
+
+func NewPodCGroup() (*PodCGroup, error) {
+	slice, io, err := cgroup.DetectCgroupVersion("/host/sys/fs/cgroup")
```
mowangdk, line 25: constantize cgroup root path?

pkg/utils/cgroup/pod.go
```diff
@@ -0,0 +1,72 @@
...
+	_, err := os.Stat(v2Path + "/io.max")
+	if err == nil {
+		log.Infof("Detected cgroup v2 at %s", v2Path)
+		return PodCGroupSlice{v2Path}, V2{}, nil
+	}
+	if !errors.Is(err, os.ErrNotExist) {
+		return PodCGroupSlice{}, nil, err
+	}
+
+	v1Path := filepath.Join(cgroupMountPath, "blkio/kubepods.slice")
```
mowangdk, line 60: Do we need to create a directory in the codes? The v1 cgroup path was previously established in the YAML.

huww98: No. Each directory in cgroupfs corresponds to a cgroup. kubelet is responsible for creating the cgroup. Even we create the cgroup, we cannot put processes into this cgroup.

mowangdk: okay,  you had tested in new cluster right?

huww98: I've checked the log of our CI, which creates new cluster at each run.
```
2024-06-11T13:15:22.367483041+08:00 stderr F time="2024-06-11T13:15:22+08:00" level=info msg="Detected cgroup v1 at /host/sys/fs/cgroup/blkio/kubepods.slice"
```
And I think we don't need to worry about this, because csi-plugin itself is a pod, it is also in the cgroup kubepods.slice. So it must exist when csi-plugin is initializing.

mowangdk: okay

PR #1083, opened 2024-06-12
disk: refactor create disk

huww98:
#### What type of PR is this?

/kind cleanup

**disk: aggregate category metadata into one file**

We will not forget something when adding new category.

**disk: refactor create disk**

Use hash of diskName as client token to ensure only one disk is created for one PVC.
If we see IdempotentParameterMismatch, fall back to DescribeDisks to fetch existing disk.

When CreateDisk fails, it records the client token only for some case, but we don't know which.
If we use a new token for retry, we are risking creating duplicated disks.
If we reuse the same token. the retry may fail for IdempotentParameterMismatch.
So fetching the existing disk and checking it ourselves seems to be the only choice.

Fixes #

Replaces #876

```release-note
The client token generation algorithm changed. There is a very small chance that a duplicated disk will be created for a PVC.
```

```docs

```

### DIFF
--- a/pkg/cloud/ecsinterface.go
+++ b/pkg/cloud/ecsinterface.go
@@ -3,6 +3,7 @@ package cloud
 import "github.com/aliyun/alibaba-cloud-sdk-go/services/ecs"
 
 type ECSInterface interface {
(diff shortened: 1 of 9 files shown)


huww98:
I will try to add unit test later

PR #993, opened 2024-03-07
unify node/controller service enabling

huww98:
#### What type of PR is this?

/kind cleanup
/kind feature
/kind deprecation

#### What this PR does / why we need it:

Every driver supports enable/disable node or controller service, with unified interface.

deprecate SERVICE_TYPE env
deprecate -run-as-controller flags
introduce -run-controller-service and -run-node-service, allowing controller and node service to be enabled/disabled independently.

#### Which issue(s) this PR fixes:

Fixes #

#### Does this PR introduce a user-facing change?

```release-note
new -run-controller-service and -run-node-service flags are introduced, -run-as-controller and SERVICE_TYPE env are deprecated.
```

#### Additional documentation e.g., KEPs (Kubernetes Enhancement Proposals), usage docs, etc.:

```docs

```

PR #1086, opened 2024-06-17
bugfix: Fix IsVFNode check error in provisioner

mowangdk:
#### What type of PR is this?
/kind bug

Fix IsVFNode check error in provisioner

None

```release-note

```

```docs

```

### DIFF
--- a/pkg/disk/cloud.go
+++ b/pkg/disk/cloud.go
@@ -222,7 +222,7 @@ func attachDisk(ctx context.Context, tenantUserUID, diskID, nodeID string, isSha
 		devicePaths := calcNewDevices(before, after)
 
 		// BDF Disk Logical
-		if IsVFNode() && len(devicePaths) == 0 {
+		if !GlobalConfigVar.ControllerService && IsVFNode() && len(devicePaths) == 0 {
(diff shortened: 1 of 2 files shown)


mowangdk:
@huww98 ptal

huww98 (commented):
Not ideal. But we can fix it later.
/lgtm
/hold
please unhold if you still want to proceed given the comments.

pkg/disk/cloud.go
```diff
@@ -222,7 +222,7 @@ func attachDisk(ctx context.Context, tenantUserUID, diskID, nodeID string, isSha
 		devicePaths := calcNewDevices(before, after)
 
 		// BDF Disk Logical
-		if IsVFNode() && len(devicePaths) == 0 {
+		if !GlobalConfigVar.ControllerService && IsVFNode() && len(devicePaths) == 0 {
```
huww98, line 225: After #993 , We can actually enable controller and node service in the same process. And using global variable is not good. Maybe we should introduce a new parameter to attachDisk(). We should replace all ADControllerEnable related branch with the new parameter.

PR #1102, opened 2024-07-01
disk: remove the call to NeedResize

huww98:

It is unnecessary, the Resize itself is idempotent. Get rid of the dependency on blockdev executable.

Fixes #

```release-note
NONE
```

```docs

```

### DIFF
--- a/pkg/disk/nodeserver.go
+++ b/pkg/disk/nodeserver.go
@@ -704,16 +704,9 @@ func (ns *nodeServer) NodeStageVolume(ctx context.Context, req *csi.NodeStageVol
 	klog.Infof("NodeStageVolume: Mount Successful: volumeId: %s target %v, device: %s, mkfsOptions: %v, options: %v", req.VolumeId, targetPath, device, mkfsOptions, mountOptions)
 
 	r := k8smount.NewResizeFs(diskMounter.Exec)
-	needResize, err := r.NeedResize(device, targetPath)
-	if err != nil {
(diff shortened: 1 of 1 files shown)


mowangdk:
This is refer to the [aws ebs driver](https://github.com/kubernetes-sigs/aws-ebs-csi-driver/blob/2c4937ac08bb783c9d2f18337e1888802d78475b/pkg/driver/node.go#L282) .

huww98:
That does not mean it is good.

huww98:
rebased to resolve conflict

mowangdk:
There might be a reason behind it, let's proceed straightforwardly.

huww98:
https://github.com/kubernetes-sigs/gcp-compute-persistent-disk-csi-driver/blob/27ac6ca1261e71082f1d2e982d345cdf9b06b58b/pkg/gce-pd-csi-driver/node.go#L381 Google does not check.

We also not checking at other place. https://github.com/kubernetes-sigs/alibaba-cloud-csi-driver/blob/5d9fc1d3b93a9bde8900ba892c53c607c3fb1ea5/pkg/disk/nodeserver.go#L965

So, not checking should be more straightforward.

huww98:
/reopen
/remove-lifecycle rotten
@mowangdk As discussed before, removing this is actually more straightforward. The code is simpler, and the dependency is reduced. PTAL again.

huww98:
related https://github.com/kubernetes/kubernetes/pull/131568

mowangdk:
Await community change effect, block until the next release

PR #1111, opened 2024-07-23
feat(OSS): support read token from secret

AlbeeSo:
#### What type of PR is this?
/kind feature

support read temporary token from a specified secret in cluster. users should keep this secret well and refresh the token if need.

Fixes #

```release-note

```

```docs

```

### DIFF
--- a/pkg/mounter/ossfs_utils.go
+++ b/pkg/mounter/ossfs_utils.go
@@ -1,0 +1,127 @@
+package mounter
+
+import (
+	"context"
+	"encoding/json"
+	"fmt"
+
+	"github.com/alibabacloud-go/tea/tea"
+	log "github.com/sirupsen/logrus"
+	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/types"
+	"k8s.io/client-go/kubernetes"
+)
+
+// keep consitent with RAM response
+var secretRefKeysToParse []string = []string{
+	"AccessKeyId",
+	"AccessKeySecret",
+	"Expiration",
+	"SecurityToken",
+}
+
+func getPasswdSecretVolume(secretRef, nodeName, volumeId string) (secret *corev1.SecretVolumeSource) {
+	passwdFilename := "passwd-ossfs"
+	if secretRef == "" {
+		secret = &corev1.SecretVolumeSource{
+			SecretName: OssfsCredentialSecretName,
+			Items: []corev1.KeyToPath{
+				{
+					Key:  fmt.Sprintf("%s.%s", nodeName, volumeId),
+					Path: passwdFilename,
+					Mode: tea.Int32(0600),
+				},
+			},
+			Optional: tea.Bool(true),
+		}
+	} else {
+		items := []corev1.KeyToPath{}
+		for _, key := range secretRefKeysToParse {
+			item := corev1.KeyToPath{
+				Key:  key,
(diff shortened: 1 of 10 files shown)


pkg/mounter/ossfs_utils.go
```diff
@@ -0,0 +1,127 @@
...
+			},
+			Optional: tea.Bool(true),
+		}
+	} else {
+		items := []corev1.KeyToPath{}
+		for _, key := range secretRefKeysToParse {
+			item := corev1.KeyToPath{
+				Key:  key,
+				Path: fmt.Sprintf("%s/%s", passwdFilename, key),
+				Mode: tea.Int32(0600),
```
huww98, line 45: Use defaultMode for this? Also. I don't expect importing `tea` here.

Do we really need to append `passwd-ossfs/` to the Path? If we drop the prefix, we don't need all these `KeyToPath`. And we can set Optional to false, to ensure the secret do exist as expected before ossfs start.

pkg/oss/errors.go
```diff
@@ -0,0 +1,15 @@
...
+)
+
+var ParamError = errors.New("Oss parameters error")
+var AuthError = errors.New("Oss authorization error")
+var EncryptError = errors.New("Oss encrypted error")
+var PathError = errors.New("Oss path error")
+
+func WrapOssError(errType error, message string) error {
+	return fmt.Errorf("%w: %s", errType, message)
+}
```
huww98, line 15: Why not just use `fmt.Errorf` at each location? The code length is similar.

pkg/oss/errors.go
```diff
@@ -0,0 +1,15 @@
+package oss
+
+import (
+	"errors"
+	"fmt"
+)
+
+var ParamError = errors.New("Oss parameters error")
```
huww98, line 8:
```suggestion
var ParamError = errors.New("OSS parameters error")
```
It is strange. but maybe fine because the old version looks like this.

pkg/oss/nodeserver_test.go
```diff
@@ -16,58 +16,185 @@ limitations under the License.
...
+			},
+			errType: nil,
+		},
+	}
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			err := checkOssOptions(tt.opts)
+			if !isErrorType(err, tt.errType) {
+				t.Errorf("checkOssOptions() error = %v, wantErrType %v", err, tt.errType)
+			}
```
huww98, line 197:
```suggestion
			assert.ErrorIs(t, err, tt.errType)
```

pkg/oss/utils.go
```diff
@@ -70,10 +70,10 @@ func checkRRSAParams(opt *Options) error {
 		return nil
 	}
 	if opt.RoleArn != "" || opt.OidcProviderArn != "" {
-		return errors.New("Oss parameters error: use RRSA but one of the ARNs is empty, roleArn:" + opt.RoleArn + ", oidcProviderArn:" + opt.OidcProviderArn)
+		return errors.New("use RRSA but one of the ARNs is empty, roleArn:" + opt.RoleArn + ", oidcProviderArn:" + opt.OidcProviderArn)
```
huww98, line 73: use fmt.Errorf instead of string concat.

huww98 (changes requested):

pkg/mounter/ossfs_utils.go
AlbeeSo: Ossfs restricts passwd-file mode without any others permission, 600 is a normal mode as official document.
I removed the secret brought from secretRef, and for csi-ossfs-creds secret maintained by CSI @tydra-wang ptal

PR #1117, opened 2024-07-31
feature: add rund csi protocol3 for oss

mowangdk:
#### What type of PR is this?
/kind feature

Support rund-csi 3.0 protocol for oss

Fixes #

Compatible with coco protocol

```release-note

```

```docs

```

### DIFF
--- a/pkg/oss/nodeserver.go
+++ b/pkg/oss/nodeserver.go
@@ -18,9 +18,11 @@ package oss
 
 import (
 	"context"
+	"encoding/json"
 	"errors"
 	"fmt"
 	"os"
+	"path/filepath"
 	"strconv"
 	"strings"
 	"time"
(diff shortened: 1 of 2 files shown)


mowangdk:
@mozillazg please take a look as well~

pkg/oss/nodeserver.go
```diff
@@ -239,6 +242,28 @@ func (ns *nodeServer) NodePublishVolume(ctx context.Context, req *csi.NodePublis
...
+		opts, err := json.Marshal(opt)
+		if err != nil {
+			log.Errorf("NodePublishVolume:: marshal opt error: %v", err)
+			return nil, status.Error(codes.Internal, fmt.Sprintf("OSS: failed to mount volume in rund-csi 3.0, err: %s", err.Error()))
+		}
+		mountInfo := directvolume.MountInfo{
+			Source:     opt.Bucket,
+			DeviceType: directvolume.DeviceTypeOSS,
+			FSType:     "",
+			MountOpts:  []string{string(opts)},
```
huww98, line 255: Are you sure we need to put a json object in this MountOpts? This looks extremely strange. I'd expect to pass a slice of command line args here.
FYI, disk driver passes `req.GetVolumeCapability().GetMount().GetMountFlags()` to it.

mowangdk: oss doesn't have standard fixed parameters like disk, so I prefer to pass all parameters to the calling side of ossfs and let the calling side make its own judgement. And we don't want to have additional protocols to interact with the ossfs container. So we keep the same key value as csi-plugin.

huww98: I don't agree. You can mount ossfs with /etc/fstab, so obviously it as standard method to specify the parameters.
I would suggust:
DeviceType: ""
FSType: "fuse.ossfs"
MountOpts: built `options` from: https://github.com/kubernetes-sigs/alibaba-cloud-csi-driver/blob/c708e0132b6a366f9948046acab784bbba58a4ce/pkg/mounter/ossfs.go#L202

PR #1119, opened 2024-08-01
introduce new alibabacloud-disk storage class 

huww98:
#### What type of PR is this?

/kind feature

It does not pass any parameters and let CSI decide which type to use. So that we may change the behavior in the future.

Fixes #

```release-note
A new storage class alibabacloud-disk is introduced for users who just want a PV. 
```

```docs

```


huww98:
We should recommend this new storage class to all new users. If one does want a specific type, he should use VolumeAttributeClass to specify the type, either at creation or after creation.

PR #1122, opened 2024-08-02
Upgrade cluster tag key for new created disk and snapshot

mowangdk:
#### What type of PR is this?

/kind api-change
/kind cleanup

csi-provisioner now supports multiple types CO, rendering the "ack" tag name unsuitable for identifying resources it creates.

Fixes #
None

```release-note

```

```docs

```

### DIFF
--- a/pkg/common/constants.go
+++ b/pkg/common/constants.go
@@ -11,6 +11,9 @@ const (
 	// PVNameKey contains name of the final PV that will be used for the dynamically
 	// provisioned volume
 	PVNameKey = "csi.storage.k8s.io/pv/name"
+
+	// ClusterIDKey contains the cluster id key of the PV cities in
+	ClusterIDKey = "csi.storage.k8s.io/clusterid"
 )
 
 // constants of keys in volume snapshot parameters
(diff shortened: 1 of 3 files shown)


pkg/disk/cloud.go
```diff
@@ -1220,6 +1221,7 @@ func listSnapshots(ecsClient cloud.ECSInterface, diskID, clusterID, nextToken st
 	if clusterID != "" {
 		describeRequest.Tag = &[]ecs.DescribeSnapshotsTag{
 			{Key: DISKTAGKEY3, Value: clusterID},
+			{Key: common.ClusterIDKey, Value: clusterID},
```
huww98, line 1224: I think this will not work. It is selecting snapshots with both tags.

huww98 (changes requested):
So, this is a breaking change. Maybe we should keep both for half a year or so? To let the clients upgrade their scripts.

pkg/common/constants.go
```diff
@@ -11,6 +11,9 @@ const (
 	// PVNameKey contains name of the final PV that will be used for the dynamically
 	// provisioned volume
 	PVNameKey = "csi.storage.k8s.io/pv/name"
+
+	// ClusterIDKey contains the cluster id key of the PV cities in
+	ClusterIDKey = "csi.storage.k8s.io/clusterid"
```
huww98, line 16: This should be added after line 46.
Maybe naming it `csi.alibabacloud.com/cluster-id` ? I think we should not use k8s.io without the consensus with the community.

mowangdk:
/

PR #1126, opened 2024-08-05
Add clusterid tag to static provision volume

mowangdk:
#### What type of PR is this?
/kind feature

Add clusterid tag to static provision volume

Fixes #

```release-note

```

```docs

```


huww98 (commented):
How about just call getDefaultDiskTags here? We may also need some more tags. We also have non-ACK deployment, which does not have clusterID.

mowangdk:
> How about just call getDefaultDiskTags here? We may also need some more tags. We also have non-ACK deployment, which does not have clusterID.

getDefaultTags means default tags need to added to disk that created by csi-provisioner.  not suitable for the current context

PR #1135, opened 2024-08-16
disk: support for EED

huww98:
#### What type of PR is this?

/kind feature

disk: support for EED

Fixes #

EED only supports WaitForFirstConsumer volumeBindingMode. It will be bound to a single node on CreateVolume.
EED only requires binding after first attach, but we cannot modify the PV affinity after creation.

If VK creates an EED, we cannot prevent the next pod using it from being scheduled to ECS. So the only valid use-case with EED and VK is ephemeral volume.

```release-note
Support for Elastic Ephemeral Disk (EED)
```

```docs

```

### DIFF
--- a/pkg/disk/cloud.go
+++ b/pkg/disk/cloud.go
@@ -40,8 +40,7 @@ import (
 	log "github.com/sirupsen/logrus"
 	"google.golang.org/grpc/codes"
 	"google.golang.org/grpc/status"
-	apierrors "k8s.io/apimachinery/pkg/api/errors"
-	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	v1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/util/sets"
 	"k8s.io/apimachinery/pkg/util/wait"
 )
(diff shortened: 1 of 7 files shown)


pkg/disk/cloud.go
```diff
@@ -925,7 +924,7 @@ func clientToken(name string) string {
 // So we just assume they are not supported.
 var vaildDiskNameRegexp = regexp.MustCompile(`^[a-zA-Z][a-zA-Z0-9:_-]{1,127}$`)
 
-func createDisk(ecsClient cloud.ECSInterface, diskName, snapshotID string, diskVol *diskVolumeArgs, supportedTypes sets.Set[Category]) (string, createAttempt, error) {
+func createDisk(ecsClient cloud.ECSInterface, diskName, snapshotID string, diskVol *diskVolumeArgs, supportedTypes sets.Set[Category], selectedInstance string) (string, createAttempt, error) {
```
mowangdk, line 927: Should we distinguish EED from other cloud disk types? Perhaps we shouldn't conflate it with cloud_* disks in types field

mowangdk: For now, we support a mixed combination and have a plan in place for future distinction if needed.

mowangdk:
/lgtm
/approve

Need to add e2e test later

PR #1141, opened 2024-08-22
Add csi-agent for oss mounting in ACS rund

tydra-wang:
#### What type of PR is this?

/kind feature

Fixes #

```release-note

```

```docs

```

### DIFF
--- a/pkg/oss/nodeserver.go
+++ b/pkg/oss/nodeserver.go
@@ -132,11 +132,6 @@ func validateNodePublishVolumeRequest(req *csi.NodePublishVolumeRequest) error {
 
 func (ns *nodeServer) NodePublishVolume(ctx context.Context, req *csi.NodePublishVolumeRequest) (*csi.NodePublishVolumeResponse, error) {
 	klog.Infof("NodePublishVolume:: Starting Mount volume: %s mount with req: %+v", req.VolumeId, req)
-	if features.FunctionalMutableFeatureGate.Enabled(features.RundCSIProtocol3) {
-		log.Infof("NodePublishVolume: skip as the rund 3.0 protocol enabled")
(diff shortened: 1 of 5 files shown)


pkg/oss/nodeserver.go
```diff
@@ -129,6 +130,11 @@ func validateNodePublishVolumeRequest(req *csi.NodePublishVolumeRequest) error {
 }
 
 func (ns *nodeServer) NodePublishVolume(ctx context.Context, req *csi.NodePublishVolumeRequest) (*csi.NodePublishVolumeResponse, error) {
+	if features.FunctionalMutableFeatureGate.Enabled(features.RundCSIProtocol3) {
+		log.Infof("NodePublishVolume: skip as the rund 3.0 protocol enabled")
+		return &csi.NodePublishVolumeResponse{}, nil
+	}
```
huww98, line 136: What is the logic here? skip the whole NodePublishVolume if a feature gate is enabled?

If this will not become the default in the future, we should still use an env or other config, rather than a feature gate.

tydra-wang: Yeah it's strange here. We should check the runtimeClass of the pod and only skip mounting for rund, similar to how the feature gate works for disk and nas.

mowangdk: This feature gate determines the the kata runtime storage communication protocol with csi which will goes to 3.0 ultimately. why don't we check runtime class in pod first?

mowangdk:
RundCSIProtocol3 judgment seems anomalous,  but it allows skipping a single pod get request. 
/lgtm
/approve

huww98:
Looks like conflict with klog PR. Please rebase. @tydra-wang

PR #1142, opened 2024-08-23
feat: add volume attachment time & csi grpc execution time metrics

iltyty:
#### What type of PR is this?
/kind feature

Fixes #

```release-note

```

```docs

```


pkg/common/interceptors.go
```diff
@@ -0,0 +1,165 @@
...
+
+func getMethodName(fullName string) string {
+	i := strings.LastIndex(fullName, "/")
+	if !strings.Contains(fullName, "/") || i == len(fullName)-1 {
+		return ""
+	}
+	return fullName[i+1:]
+}
+
+func observeExecTime(time int64, method, driverType, volumeId string, err error) {
```
huww98, line 110:
```suggestion
func observeExecTime(time time.Duration, method, driverType, volumeId string, err error) {
```
Just pass in the Duration to avoid any ambiguous about unit.

pkg/common/interceptors.go
```diff
@@ -0,0 +1,165 @@
...
+		return handler(ctx, req)
+	}
+
+	volumeId := req.(reqWithVolumeId).GetVolumeId()
+	method := getMethodName(info.FullMethod)
+	switch method {
+	case ControllerPublish, ControllerUnpublish, NodeStage, NodeUnstage, NodeUnpublish:
+		start := time.Now()
+		resp, err := handler(ctx, req)
+		execTime := time.Since(start).Milliseconds()
```
huww98, line 83:
```suggestion
		execTime := time.Since(start).Seconds()
```
Use seconds as it already returns a float64.

pkg/metric/csi_grpc_exec_time_collector.go
```diff
@@ -0,0 +1,41 @@
...
+	}, csiGrpcExecTimeLabels),
+}
+
+func init() {
+	registerCollector(CsiGrpcExecTimeCollectorName, NewCsiGrpcExecTimeCollector, diskDriverName, nasDriverName, ossDriverName)
+}
+
+func NewCsiGrpcExecTimeCollector() (Collector, error) {
+	return &CsiGrpcExecTimeCollector, nil
+}
```
huww98, line 36: This is confusing. It is named New but returns an existing object.

pkg/metric/volume_stat_collector.go
```diff
@@ -0,0 +1,46 @@
+package metric
+
+import "github.com/prometheus/client_golang/prometheus"
+
+const (
+	VolumeStatsCollectorName = "volume_stats"
+	VolumeStatsLabelType     = "type"
+	VolumeStatsLabelId       = "id"
```
huww98, line 8: Remove ID as we discussed. We don't expect unbounded number of label values.

pkg/common/interceptors.go
```diff
@@ -0,0 +1,165 @@
...
+	}
+	return handler(ctx, req)
+}
+
+func getMethodName(fullName string) string {
+	i := strings.LastIndex(fullName, "/")
+	if !strings.Contains(fullName, "/") || i == len(fullName)-1 {
+		return ""
+	}
+	return fullName[i+1:]
```
huww98, line 107: I think this is not needed? If "/" is not in fullName, LastIndex will return -1, and fullName[i+1:] should become the fullName.
If i == len(fullName)-1, fullName[i+1:] will become empty string automatically.

pkg/common/interceptors.go
```diff
@@ -0,0 +1,165 @@
...
+
+func getCodeFromError(err error) codes.Code {
+	status, ok := status.FromError(err)
+	if ok {
+		return status.Code()
+	}
+	return codes.Unknown
+}
+
+func observeVolumeAttachmentTime(curTime int64, req *csi.NodePublishVolumeRequest, driverType string, err error) {
```
huww98, line 128:
```suggestion
func observeVolumeAttachmentTime(curTime time.Time, req *csi.NodePublishVolumeRequest, driverType string, err error) {
```
Again, to avoid ambiguity.

pkg/common/interceptors.go
```diff
@@ -0,0 +1,165 @@
...
+	if client == nil {
+		if client = getK8sClient(); client == nil {
```
huww98, line 130: This is not thread safe. And please try to avoid global variable.

pkg/common/interceptors.go
```diff
@@ -0,0 +1,165 @@
...
+			return
+		}
+	}
+	errCode := getCodeFromError(err).String()
+	podName, podNamespace := req.VolumeContext[podNameKey], req.VolumeContext[podNamespaceKey]
+	if podName == "" || podNamespace == "" {
+		log.Warnf("observeVolumeAttachmentTime: empty pod name/namespace: %s, %s", podName, podNamespace)
+		return
+	}
+	pod, err := client.CoreV1().Pods(podNamespace).Get(context.Background(), podName, metav1.GetOptions{})
```
huww98, line 141: Additional API server invocation. This can hinder the overall throughput of CSI. We should try to avoid this. Combine it to the runtime class detection?

pkg/common/interceptors.go
```diff
@@ -0,0 +1,165 @@
...
+	volumeId := req.(reqWithVolumeId).GetVolumeId()
+	method := getMethodName(info.FullMethod)
+	switch method {
+	case ControllerPublish, ControllerUnpublish, NodeStage, NodeUnstage, NodeUnpublish:
+		start := time.Now()
+		resp, err := handler(ctx, req)
+		execTime := time.Since(start).Milliseconds()
+		observeExecTime(execTime, method, driverType, volumeId, err)
+		return resp, err
+	case NodePublish:
```
huww98, line 86: How about add this one to a real NodePublishVolume handler? We should avoid string comparisons and prefer strongly typed handler generally.

pkg/common/interceptors.go
```diff
@@ -0,0 +1,129 @@
...
+		request := req.(*csi.NodePublishVolumeRequest)
+		observeVolumeAttachmentTime(clientset, end, request, driverType, err)
+		return resp, err
+	}
+	return handler(ctx, req)
+}
+
+func getMethodName(fullName string) string {
+	if !strings.Contains(fullName, "/") {
+		return ""
```
huww98, line 73: This can be hard to debug. Better to just return fullName?

huww98 (changes requested):
Still note the extra APIServer call to get pod, and hard-coded GRPC method name.

pkg/common/interceptors.go
huww98: Please consider this again. Wrap controllerServer again and implement NodePublishVolume method, instead of compare string. This should also reduce duplicated code.

mowangdk: it's too duplicated to implement within every grpc calls, string compareble will be fine to me, after all we do not utilize any reflection.

pkg/metric/collector.go
```diff
@@ -71,8 +71,8 @@ func newCSICollector(metricType string, driverNames []string) error {
 				collectors[reg.Name] = collector
 			}
 		}
-
 	}
+	collectors[CsiGrpcExecTimeCollectorName] = &CsiGrpcExecTimeCollector
```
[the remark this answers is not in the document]
iltyty: registered collectors are added to csiCollectorInstance only at the plugin side, see line 53, this line is for the provisioner to work as well

pkg/common/nodeserver.go
```diff
@@ -3,16 +3,92 @@ package common
...
+
+func (s *NodeServerWithMetricRecorder) NodePublishVolume(ctx context.Context, req *csi.NodePublishVolumeRequest) (*csi.NodePublishVolumeResponse, error) {
+	ctx = withPodInfo(ctx, s.client, req)
+	resp, err := s.NodeServer.NodePublishVolume(ctx, req)
+	s.recordVolumeAttachmentTime(ctx, req, err)
+	return resp, err
+}
+
+func withPodInfo(ctx context.Context, client kubernetes.Interface, req *csi.NodePublishVolumeRequest) context.Context {
+	name, namespace := req.VolumeContext[podNameKey], req.VolumeContext[podNamespaceKey]
```
mowangdk, line 46: extract duplicated codes into a new utils function

huww98: I would move `withPodInfo` to util.go. Getter and setter should be in the same place. Maybe also make `PodInfo` private, to avoid any misuse. How about also cache the error into ctx? so that we will not trigger the error twice.

And maybe `withPodInfo` can return the updated ctx and the `*v1.Pod` at the same time, so that we can directly pass `*v1.Pod` into `recordVolumeAttachmentTime`

pkg/common/nodeserver.go
```diff
@@ -3,16 +3,92 @@ package common
...
+		klog.Errorf("recordVolumeAttachmentTime: volume %s: error getting scheduled time for pod %s/%s", pod.GetNamespace(), pod.GetName(), req.GetVolumeId())
+		return
+	}
+
+	labels := prometheus.Labels{
+		metric.VolumeStatsLabelType: s.driverType,
+		metric.VolumeStatsLabelCode: status.Code(err).String(),
+	}
+	metric.VolumeStatCollector.AttachmentCountMetric.With(labels).Inc()
+	metric.VolumeStatCollector.AttachmentTimeTotalMetric.With(labels).Add(float64(time.Since(podStartTime).Seconds()))
```
huww98, line 81:
```suggestion
	metric.VolumeStatCollector.AttachmentTimeTotalMetric.With(labels).Add(time.Since(podStartTime).Seconds())
```
redundant

PR #1147, opened 2024-08-28
OSS: set default ossfs default log level to err

AlbeeSo:
Ossfs's default dbglevel is crit, which means users will not get any error message during mounting.
As Ossfs is containerized, I suggest setting dbglevel to err by default by CSI. 

Fixes #

```release-note

```

```docs

```

### DIFF
--- a/pkg/mounter/ossfs.go
+++ b/pkg/mounter/ossfs.go
@@ -185,21 +185,23 @@ func (f *fuseOssfs) buildPodSpec(c *FusePodContext, target string) (spec corev1.
 }
 
 func (f *fuseOssfs) AddDefaultMountOptions(options []string) []string {
+	alreadySet := false
+	for _, option := range options {
+		if strings.Contains(option, "dbglevel") {
+			alreadySet = true
+			break
+		}
+	}
+	setDbgLevel := "err"
 	switch dbglevel := f.config.Extra["dbglevel"]; dbglevel {
 	case "":
 	case "debug", "info", "warn", "err", "crit":
-		alreadySet := false
-		for _, option := range options {
-			if strings.Contains(option, "dbglevel") {
-				alreadySet = true
-				break
-			}
-		}
-		if !alreadySet {
-			options = append(options, "dbglevel="+dbglevel)
-		}
+		setDbgLevel = dbglevel
 	default:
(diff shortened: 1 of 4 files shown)


tydra-wang (approved):

pkg/mounter/ossfs.go
```diff
@@ -185,22 +185,24 @@ func (f *fuseOssfs) buildPodSpec(c *FusePodContext, target string) (spec corev1.
...
-		if !alreadySet {
-			options = append(options, "dbglevel="+dbglevel)
-		}
+		setDbgLevel = dbglevel
 	default:
 		log.Warnf("invalid ossfs dbglevel: %q", dbglevel)
 	}
+	if !alreadySet {
+		options = append(options, fmt.Sprintf("dbglevel=%s", setDbgLevel))
+	}
```
huww98, line 205: I sugguest this implementation:
```golang
	alreadySet := slices.ContainsFunc(options, func(option string) bool {
		return strings.Contains(option, "dbglevel")
	})
	if !alreadySet {
		switch dbglevel := f.config.Extra["dbglevel"]; dbglevel {
		case "":
		case "debug", "info", "warn", "err", "crit":
			options = append(options, "dbglevel="+dbglevel)
		default:
			log.Warnf("invalid ossfs dbglevel: %q", dbglevel)
		}
	}
```
skip whole switch case if alreadySet. use slices.ContainsFunc to replace the manual loop for clearer semantic.

huww98: BTW, I think we should verify the dbglevel at `extractFuseContainerConfig`, not here. Maybe refuse to start the CSI driver if it is invalid.

pkg/mounter/ossfs_test.go
```diff
@@ -52,3 +53,80 @@ func Test_buildAuthSpec(t *testing.T) {
...
+					},
+				},
+			}
+			got := f.AddDefaultMountOptions(tt.options)
+			if len(got) != len(tt.want) {
+				t.Errorf("AddDefaultMountOptions() got = %v, want %v", got, tt.want)
+				return
+			}
+			for i := range got {
+				if !strings.EqualFold(got[i], tt.want[i]) {
```
huww98, line 125: Why `EqualFold`? e.g. file path should be case sensitive?

PR #1162, opened 2024-09-06
Support set nfs readahead in PV attributes

tydra-wang:
#### What type of PR is this?

/kind feature

Fixes #

```release-note
1. Support setting nfs readahead in the volumeAttributes of NAS PV like 'sysConfig: bdi/read_ahead_kb=2048'.
2. Restrict disk sysConfig keys to those with the prefix 'queue/'.

```

```docs

```

### DIFF
--- a/pkg/disk/nodeserver.go
+++ b/pkg/disk/nodeserver.go
@@ -499,6 +499,11 @@ func (ns *nodeServer) NodeStageVolume(ctx context.Context, req *csi.NodeStageVol
 	}
 	defer ns.locks.Release(req.VolumeId)
 
+	sysConfigs, err := utilsio.ParseSysConfigs(req.VolumeContext[SysConfigTag], allowSysConfigKey)
+	if err != nil {
+		return nil, status.Error(codes.InvalidArgument, err.Error())
+	}
+
 	targetPath := req.StagingTargetPath
 	// targetPath format: /var/lib/kubelet/plugins/kubernetes.io/csi/pv/pv-disk-1e7001e0-c54a-11e9-8f89-00163e0e78a0/globalmount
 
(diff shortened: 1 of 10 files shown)


pkg/utils/io/sysconfig.go
```diff
@@ -0,0 +1,139 @@
...
+	return &SysConfigManager{
+		sysfsPath: defaultSysfsPath,
+		major:     major,
+		minor:     minor,
+	}, nil
+}
+
+func NewSysConfigManagerForNFS(mountPath string) (*SysConfigManager, error) {
+	var stat unix.Statx_t
+	err := unix.Statx(0, mountPath, 0, unix.STATX_BASIC_STATS, &stat)
```
huww98, line 64: Statx is introduced in 4.11, that means we cannot support this on CentOS 7. (Maybe OK?)

pkg/utils/io/sysconfig.go
```diff
@@ -0,0 +1,139 @@
...
+	if err != nil {
+		return nil, err
+	}
+
+	// check the filesystem is NFS
+	major, minor := stat.Dev_major, stat.Dev_minor
```
huww98, line 70: This should also work for the disk driver. Maybe we don't need a separate `NewSysConfigManagerForDisk`?

pkg/utils/io/sysconfig.go
```diff
@@ -0,0 +1,139 @@
...
+
+	data, err := os.ReadFile(path)
+	if err != nil {
+		return err
+	}
+	currentValue := strings.TrimSpace(string(data))
+
+	if currentValue == value {
+		return nil
+	}
```
huww98, line 117: Not all config can be read. Not all config ignores space. I think we should just drop this check and set the config anyway.

pkg/utils/io/sysconfig.go
```diff
@@ -0,0 +1,139 @@
...
+		return err
+	}
+
+	logrus.Infof("Set %s from %s to %s", path, currentValue, value)
+	return nil
+}
+
+func (m *SysConfigManager) sysfsPathForKey(key string) (string, error) {
+	devNum := fmt.Sprintf("%d:%d", m.major, m.minor)
+	if m.nonBlockFs {
```
huww98, line 130: Block device based fs also have `/sys/class/bdi/`. I think we can use this path in all cases and drop this branch.

pkg/disk/nodeserver.go
```diff
@@ -611,18 +611,22 @@ func (ns *nodeServer) NodeStageVolume(ctx context.Context, req *csi.NodeStageVol
...
-		for _, configStr := range configList {
-			key, value, found := strings.Cut(configStr, "=")
-			if !found {
-				log.Errorf("NodeStageVolume: Volume Block System Config with format error: %s", configStr)
-				return nil, status.Error(defaultErrCode, "NodeStageVolume: Volume Block System Config with format error "+configStr)
+		sysConfigs, err := utilsio.ParseSysConfigs(value, func(key string) bool {
+			return strings.HasPrefix(key, "queue/")
+		})
+		if err != nil {
+			return nil, status.Error(codes.InvalidArgument, err.Error())
```
huww98, line 618: Should we also add this validation to CreateVolume? I think we should report errors as early as possible.

pkg/utils/io/sysconfig.go
```diff
@@ -0,0 +1,139 @@
...
+	err := unix.Statx(0, mountPath, 0, unix.STATX_BASIC_STATS, &stat)
+	if err != nil {
+		return nil, err
+	}
+
+	// check the filesystem is NFS
+	major, minor := stat.Dev_major, stat.Dev_minor
+	if major != 0 {
+		return nil, errors.New("unexpected nfs dev major")
+	}
```
huww98, line 73: Any docs state that NFS must have major==0? I don't think this is a safe assumption. e.g.
>  144 block	Expansion Area # 1 for more non-device (e.g. NFS) mounts
> 		  0 = mounted device 256
> 		255 = mounted device 511

https://www.kernel.org/doc/Documentation/admin-guide/devices.txt

PR #1027, opened 2024-04-18
disk: cancel waiting for attach/detach when GRPC cancelled

huww98:
#### What type of PR is this?

/kind cleanup

#### What this PR does / why we need it:

The default timeout of external-attacher is only 15s. Timeout is common. We can save some OpenAPI calls after timeout.

#### Which issue(s) this PR fixes:

Fixes #

#### Does this PR introduce a user-facing change?

```release-note
NONE
```

#### Additional documentation e.g., KEPs (Kubernetes Enhancement Proposals), usage docs, etc.:

```docs

```

PR #1170, opened 2024-09-19
disk: avoid concurrent operation on the same disk

huww98:
#### What type of PR is this?

/kind bug

If the umount takes extremely long time and the NodeUnstageVolume RPC times out, kubelet may invoke NodeStageVolume when it retries and we may mistake an unmounting disk as a good one. This lock only protects us in one process. But fortunately, when restarting CSI, Kubernetes will wait for the previous process to exit before starting the new one, even if it stucks in uninterruptible syscall (D state). So, we will not have multiple processes on the same node and the lock should be safe enough.

Note that there is still small chance that the previous attach/detach OpenAPI is invoked by the terminated CSI process, but the state returned by DescribeDisks is not updated yet. We should fix this by switch attach/detach to controller.

Fixes #

```release-note
Fix unintended disk detach when the unmounting is extremely slow
```

```docs

```

### DIFF
--- a/pkg/disk/attachdetach_slot.go
+++ b/pkg/disk/attachdetach_slot.go
@@ -64,6 +64,9 @@ type serialAD_DetachSlot struct{ *serialADSlot }
 type serialAD_AttachSlot struct{ *serialADSlot }
 
 func (s serialAD_DetachSlot) Aquire(ctx context.Context) error {
+	if ctx.Err() != nil {
+		return ctx.Err()
+	}
 	select {
 	case s.highPriorityChannel <- struct{}{}:
 		return nil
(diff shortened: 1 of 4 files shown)


huww98:
#1027 is better to have since it should speed up volume lock release in some case.

huww98:
/hold
This could slow down some operations. I will review this again tomorrow.

huww98:
/unhold
No better way I think. We can use more fine-grained lock, but I don't see any benefits. So let's go the simplest way

mowangdk:
How often does this problem occur, and do we need to slow down the mounting/unmounting process of all scenarios for an occasional bug?

huww98:
> How often does this problem occur, and do we need to slow down the mounting/unmounting process of all scenarios for an occasional bug?

I guess this can be common for certain workloads, e.g. massive dirty page cache just before umount. I think we should make every effort to avoid unintended detach.

This will not slow down mounting/unmounting process of common scenarios, only some edge cases can be affected. If #1027 is also merged, I cannot think of any scenario that would be slowed down by this.

mowangdk:
gotcha, 
/lgtm
/approve

PR #1173, opened 2024-09-25
oss: translate OSS bucket domain name to endpoint automatically

AlbeeSo:
translate oss domain name like: `bucket.oss-cn-beijing.aliyuncs.com` to endpoint that ossfs need like `oss-cn-beijing.aliyuncs.com` to avoid 400 error response in ossfs.

Fixes #

```release-note

```

```docs

```

### DIFF
--- a/pkg/oss/utils.go
+++ b/pkg/oss/utils.go
@@ -145,3 +145,29 @@ func setTransmissionProtocol(originURL string) (URL string, modified bool) {
 	}
 	return "https://" + URL, true
 }
+
+// translateDomainToEndpoint will trim the prefix with "bucket." to
+//
+//	translate the bucket domain name like: bucket.oss-cn-beijing.aliyuncs.com
+//	to endpoint like: oss-cn-beijing.aliyuncs.com
+func translateDomainToEndpoint(originURL, bucket string) (URL string, modified bool) {
+	URL = originURL
+	if utils.IsPrivateCloud() {
+		return
+	}
+	var protocol string
+	t := URL
+	switch {
+	case strings.HasPrefix(URL, "http://"):
+		t = strings.TrimPrefix(URL, "http://")
+		protocol = "http://"
+	case strings.HasPrefix(URL, "https://"):
+		t = strings.TrimPrefix(URL, "https://")
+		protocol = "https://"
+	}
+	if strings.HasPrefix(t, bucket+".") {
+		URL = protocol + strings.TrimPrefix(t, bucket+".")
+		modified = true
+	}
+	return
+}
--- a/pkg/oss/utils_test.go
+++ b/pkg/oss/utils_test.go
@@ -17,6 +17,7 @@ limitations under the License.
 package oss
 
 import (
+	"os"
 	"reflect"
 	"testing"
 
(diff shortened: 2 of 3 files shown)


mowangdk:
Modifying users' incorrect configurations could give them the impression it's valid universally. Instead, i think issuing a warning and a specific error message would be more appropriate.

pkg/oss/utils_test.go
```diff
@@ -279,3 +280,85 @@ func Test_setTransmissionProtocol(t *testing.T) {
...
+			bucket:    "my-bucket",
+			isPrivate: true,
+			expected:  "my-bucket.oss-cn-beijing.aliyuncs.com",
+			modified:  false,
+		},
+	}
+
+	for _, tt := range testCases {
+		t.Run(tt.name, func(t *testing.T) {
+			setPrivateCloud(tt.isPrivate)
```
huww98, line 353: please use `t.Setenv`

pkg/oss/utils.go
```diff
@@ -145,3 +145,29 @@ func setTransmissionProtocol(originURL string) (URL string, modified bool) {
 	}
 	return "https://" + URL, true
 }
+
+// translateDomainToEndpoint will trim the prefix with "bucket." to
+//
+//	translate the bucket domain name like: bucket.oss-cn-beijing.aliyuncs.com
+//	to endpoint like: oss-cn-beijing.aliyuncs.com
+func translateDomainToEndpoint(originURL, bucket string) (URL string, modified bool) {
```
huww98, line 153: We have many URL manipulation now. How about using URL parsing (https://pkg.go.dev/net/url#Parse) instead of string manipulation?

pkg/oss/utils.go
```diff
@@ -145,3 +145,29 @@ func setTransmissionProtocol(originURL string) (URL string, modified bool) {
...
+	t := URL
+	switch {
+	case strings.HasPrefix(URL, "http://"):
+		t = strings.TrimPrefix(URL, "http://")
+		protocol = "http://"
+	case strings.HasPrefix(URL, "https://"):
+		t = strings.TrimPrefix(URL, "https://")
+		protocol = "https://"
+	}
+	if strings.HasPrefix(t, bucket+".") {
```
huww98, line 168: Can oss-cn-beijing be a valid bucket name? We should not trim oss-cn-beijing.aliyuncs.com to aliyuncs.com

pkg/oss/utils.go
mowangdk: Existing URL manipulations seem optimized. Correcting wrong URLs isn't our responsibility, I believe. besides, if we did, maybe this is more correcting things in the future.

AlbeeSo: We have similar discussion before, the point is, the 'url' here is OSS endpoint in fact... not a real url, and OSS has so many deprecated endpoints.

AlbeeSo:
> Modifying users' incorrect configurations could give them the impression it's valid universally. Instead, i think issuing a warning and a specific error message would be more appropriate.

moved to `checkOptions` and throw error in this case instead.

PR #1174, opened 2024-09-25
oss: support get sts endpoint from env for private cloud

AlbeeSo:

Fixes #

```release-note

```

```docs

```

### DIFF
--- a/deploy/chart/templates/controller.yaml
+++ b/deploy/chart/templates/controller.yaml
@@ -314,6 +314,10 @@ spec:
             - name: ECS_ENDPOINT
               value: {{ .Values.deploy.ecsEndpoint | quote }}
 {{- end }}
+{{- if .Values.deploy.stsEndpoint }}
+            - name: STS_ENDPOINT
+              value: {{ .Values.deploy.stsEndpoint | quote }}
+{{- end }}
 {{- if .Values.deploy.privateCloud }}
             - name: ALICLOUD_CLIENT_SCHEME
               value: HTTP
@@ -389,6 +393,10 @@ spec:
 {{- if .Values.deploy.ecsEndpoint }}
             - name: ECS_ENDPOINT
               value: {{ .Values.deploy.ecsEndpoint | quote }}
+{{- end }}
+{{- if .Values.deploy.stsEndpoint }}
+            - name: STS_ENDPOINT
+              value: {{ .Values.deploy.stsEndpoint | quote }}
 {{- end }}
 {{- if .Values.deploy.privateCloud }}
             - name: ALICLOUD_CLIENT_SCHEME
--- a/pkg/oss/utils.go
+++ b/pkg/oss/utils.go
@@ -68,6 +68,21 @@ func getRRSAConfig(opt *Options, m metadata.MetadataProvider) (rrsaCfg *mounter.
 	return &mounter.RrsaConfig{ServiceAccountName: saName, OidcProviderArn: oidcProviderArn, RoleArn: roleArn}, nil
 }
 
+// getSTSEndpoint get STS endpoint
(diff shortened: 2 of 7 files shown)


deploy/chart/templates/controller.yaml
```diff
@@ -315,6 +315,10 @@ spec:
             - name: ECS_ENDPOINT
               value: {{ .Values.deploy.ecsEndpoint | quote }}
 {{- end }}
+{{- if .Values.deploy.stsEndpoint }}
+            - name: STS_ENDPOINT
```
mowangdk, line 319: STS_ENDPOINT should be renamed to RRSA_STS_ENDPOINT, as it's currently only used in RRSA

huww98: But if we invoke other APIs of STS in the future, we should still use this endpoint.
BTW, RRSA is a concept for ACK, it should be OIDC for STS.

pkg/oss/utils.go
```diff
@@ -68,6 +68,21 @@ func getRRSAConfig(opt *Options, m metadata.MetadataProvider) (rrsaCfg *mounter.
 	return &mounter.RrsaConfig{ServiceAccountName: saName, OidcProviderArn: oidcProviderArn, RoleArn: roleArn}, nil
 }
 
+// getSTSEndpoint get STS endpoint
+func getSTSEndpoint(region string) (endpoint string) {
```
huww98, line 72: Don't use named return value here.

pkg/oss/utils_test.go
```diff
@@ -279,3 +280,47 @@ func Test_setTransmissionProtocol(t *testing.T) {
...
+			envEndpoint: "sts-vpc.cn-beijing.aliyuncs.com",
+			expected:    "sts-vpc.cn-beijing.aliyuncs.com",
+		},
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			if tt.envEndpoint != "" {
+				os.Setenv("STS_ENDPOINT", tt.envEndpoint)
+				defer os.Unsetenv("STS_ENDPOINT")
```
huww98, line 315:
```suggestion
				t.Setenv("STS_ENDPOINT", tt.envEndpoint)
```

deploy/chart/templates/controller.yaml
mowangdk: We should modify the endpoint in metadata as well to match stsclient in the metadata openapi fetcher

mowangdk: To utilize the universal STS_ENDPOINT  env name.

pkg/oss/utils.go
```diff
@@ -68,6 +68,21 @@ func getRRSAConfig(opt *Options, m metadata.MetadataProvider) (rrsaCfg *mounter.
...
+
+	// for PrivateCloud
+	if os.Getenv("STS_ENDPOINT") != "" {
+		endpoint = os.Getenv("STS_ENDPOINT")
```
mowangdk, line 76: if we don't want to validate endpoint, how about return the endpoint directly

deploy/chart/templates/controller.yaml
AlbeeSo: I submit a new commit to set endpoint when try new a stsclient.

pkg/utils/util.go
```diff
@@ -268,14 +269,18 @@ func NewEcsClient(ac AccessControl) (ecsClient *ecs.Client) {
 
 // NewStsClient create a stsClient object
 func NewStsClient(ac AccessControl) (stsClient *sts.Client) {
+	regionID, _ := GetRegionID()
```
mowangdk, line 272: ```GetRegionID()``` should be deprecated, use default region here would be fine

PR #1177, opened 2024-10-08
Check VBM node type automatically and support csi rund 3.0 protocol

mowangdk:

Fixes #

```release-note

```

```docs

```

### DIFF
--- a/pkg/disk/bdf.go
+++ b/pkg/disk/bdf.go
@@ -15,6 +15,7 @@ import (
 	"sync"
 	"syscall"
 	"time"
+	"regexp"
 
 	"github.com/aliyun/alibaba-cloud-sdk-go/services/ecs"
 	"github.com/kubernetes-sigs/alibaba-cloud-csi-driver/pkg/utils"
@@ -515,17 +516,17 @@ func (_type MachineType) BusName() string {
 	return busNames[_type]
 }
 
-func (_type MachineType) BusPrefix() string {
-	busPrefixes := [...]string{
+func (_type MachineType) BusRegex() (*regexp.Regexp, error) {
+	busRegexes := [...]*regexp.Regexp{
 		BDFTypeDevice,
 		DFBusTypeDevice,
 	}
 
 	if _type < BDF || _type > DFBus {
-		return fmt.Sprintf("Unknown(%d)", _type)
+		return nil, fmt.Errorf("Unknown(%d)", _type)
 	}
 
-	return busPrefixes[_type]
+	return busRegexes[_type], nil
 }
 
 type Driver interface {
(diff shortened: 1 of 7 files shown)


mowangdk:
I'll merge this PR since the e2e test for the business line has passed.

pkg/disk/bdf.go
```diff
@@ -570,5 +631,39 @@ func (d *driver) UnbindDriver() error {
...
+
+func (d *driver) CheckVFIOUsage() error {
+	actualPath, err := filepath.EvalSymlinks(filepath.Join(sysPrefix, "sys/bus", d.machineType.BusName(), "devices", d.deviceNumber, "iommu_group"))
+	if err != nil {
+		return err
+	}
+	klog.V(5).InfoS("CheckVFIOUsage: eval symlink success", "path", actualPath)
+	groupNumber := filepath.Base(actualPath)
+	// the command returns -1 if nothing is returned
+	output, _ := exec.Command("lsof", filepath.Join("/dev/vfio", groupNumber)).CombinedOutput()
```
huww98, line 664: This is expensive, I would strongly recommend not using this. Instead, ask the user to lock this file, so that we can use lock to detect usage. Or ask the user to write a PID file, and we can just check if the process is alive.

pkg/disk/bdf.go
```diff
@@ -570,5 +631,39 @@ func (d *driver) UnbindDriver() error {
...
+	if strings.TrimSpace(string(output)) != "" {
+		return errors.Errorf("CheckVFIOUsage: device: %s is still be in used, output: %s", d.deviceNumber, output)
```
huww98, line 666: convert output to string before logging. Or it will output a list of number.

pkg/disk/bdf.go
```diff
@@ -78,6 +79,13 @@ func IsNoSuchDeviceErr(err error) bool {
 	return strings.Contains(strings.ToLower(err.Error()), "no such device")
 }
 
+func IsNoSuchFileErr(err error) bool {
+	if err == nil {
+		return false
+	}
+	return strings.Contains(strings.ToLower(err.Error()), "no such file or directory")
```
huww98, line 86: don't check error string. It is subject to change. Use errors.Is instead.

pkg/disk/nodeserver.go
```diff
@@ -187,10 +196,15 @@ func NewNodeServer(m metadata.MetadataProvider) csi.NodeServer {
 	}
 
 	kataBMIOType := BDF
-	if bmType := os.Getenv("KATA_BM_IO_TYPE"); bmType == DFBus.BusName() {
+	value, err := m.Get(metadata.VmocType)
+	if err != nil {
+		klog.Errorf("get vmoc failed: %+v", err)
+	}
+	if err == nil && value == "true" {
```
huww98, line 203: else if value == "true". And we should not ignore all errors?

pkg/disk/bdf.go
```diff
@@ -504,40 +516,89 @@ func (_type MachineType) BusName() string {
...
+				continue
+			}
+			if infos[1] == strings.TrimPrefix(volumeId, "d-") {
+				DFNumber := filepath.Base(filepath.Dir(path))
+				d.deviceNumber = DFNumber
+				return d, nil
+			}
+		}
+	} else {
+		output, err := utils.CommandOnNode("xdragon-bdf", "--nvme", "-id=%s", volumeId).CombinedOutput()
```
huww98, line 597: Note: new usage of nsenter. We should try to remove nsenter usage.

pkg/disk/nodeserver.go
```diff
@@ -309,7 +323,8 @@ func (ns *nodeServer) NodePublishVolume(ctx context.Context, req *csi.NodePublis
 		}
 		klog.Infof("NodePublishVolume: TargetPath: %s is umounted, start mount in kata mode", req.TargetPath)
 		mountFlags := req.GetVolumeCapability().GetMount().GetMountFlags()
-		returned, err := ns.mountRunDVolumes(req.VolumeId, sourcePath, req.TargetPath, fsType, mkfsOptions, isBlock, mountFlags)
+		pvName := utils.GetPvNameFormPodMnt(targetPath)
```
huww98, line 326: This is fragile, target path format can change in the future.

PR #1190, opened 2024-10-17
disk: block for 1s when conflicting

huww98:
#### What type of PR is this?

/kind cleanup

Avoid too much useless retry when conflicting.

Fixes #

```release-note
NONE
```

```docs

```

### DIFF
--- a/pkg/disk/attachdetach_slot.go
+++ b/pkg/disk/attachdetach_slot.go
@@ -4,6 +4,8 @@ import (
 	"context"
 	"errors"
 	"sync"
+	"sync/atomic"
+	"time"
 )
 
 type AttachDetachSlots interface {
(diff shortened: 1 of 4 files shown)


huww98:
Delay the response is not ideal, maybe we should find another way.

huww98:
based on #1216 , so merge after it.
/hold

huww98:
/unhold
#1216 has been merged

huww98:
I should think again since we now have concurrent detach.

huww98:
Should be fine. Added some more comments.

mowangdk (commented):
## Automated Code Review

**Summary:** Adds a `blockableSlot` mechanism to the attach/detach slot system, allowing temporary blocking of attach/detach operations on a per-node basis when the node is in an unusual state (e.g., conflicting operations).

**Issues Found:**
- The `blockable.Acquire()` method uses `time.After(-time.Since(*until))` — when the block time has already passed, `-time.Since(*until)` will be negative, and `time.After` with a negative/zero duration fires immediately. This is correct behavior but could be confusing to readers.
- The CAS loop in `Block()` could theoretically spin indefinitely under extreme contention, though in practice this is unlikely with the expected concurrency levels.
- `independentBlockableSlot` acquires the underlying slot FIRST, then waits for the block. This means a goroutine holds the slot while waiting — if the block is long (e.g., 1s as the PR title suggests), this could delay other operations. Consider: should block checking happen BEFORE acquiring the slot?

**Suggestions:**
- Add a comment explaining the ordering choice (acquire slot, then wait for block) vs the alternative (wait for block, then acquire slot), since both have different tradeoff implications.
- Consider adding a metric or log when a block is active, so operators can observe when nodes are in degraded state.

**Overall:** Clean concurrency primitive with good use of atomic operations. The `independentBlockableSlot` generic wrapper is well-designed. The slot-then-block ordering decision should be documented.

pkg/disk/attachdetach_slot.go
```diff
@@ -42,15 +44,81 @@ func (a *PerNodeSlots) GetSlotFor(node string) adSlot {
...
+}
+
+func (s *blockable) Block(until time.Time) {
+	for {
+		old := s.until.Load()
+		if old != nil && old.After(until) {
+			return
+		}
+		swapped := s.until.CompareAndSwap(old, &until)
+		if swapped {
```
mowangdk, line 75: The `Acquire` method uses `time.After(-time.Since(*until))` which creates a new timer channel each loop iteration without cleanup. In practice, since the block duration is short (1s), this is unlikely to cause issues. However, `time.NewTimer` with explicit `Stop()` would be cleaner and avoid the documented timer channel leak in older Go versions.

pkg/disk/cloud.go
```diff
@@ -307,6 +307,9 @@ func (ad *DiskAttachDetach) attachDisk(ctx context.Context, diskID, nodeID strin
 				return "", status.Errorf(codes.Internal, "%v, Disk(%s) should be \"Pay by quantity\", not be \"Annual package\", please check and modify the charge type, and refer to: https://help.aliyun.com/document_detail/134767.html", err, diskID)
 			case NotSupportDiskCategory:
 				return "", status.Errorf(codes.Internal, "%v, Disk(%s) is not supported by instance, please refer to: https://help.aliyun.com/document_detail/25378.html", err, diskID)
+			case InvalidOperation_Conflict, IncorrectDiskStatus:
```
mowangdk, line 356: The `slot.Block(...)` call uses the `slot` variable from the outer scope, but `slot` is obtained from `ad.slots.GetSlotFor(nodeID).Attach()` earlier. Verify that `slot` refers to the correct attach slot here — this looks correct but is worth confirming given the refactoring.

mowangdk (commented):
## Automated Code Review

**Summary:** Adds a blocking mechanism to attach/detach slots that delays retries by 1s when conflicting ECS operations are detected (e.g., `InvalidOperation_Conflict`, `IncorrectDiskStatus`).

**Overall:** Clean implementation of a rate-limiting/backoff pattern. The atomic CAS loop for extending block time is correct. One minor observation.

PR #1192, opened 2024-10-18
disk: call sfdisk directly to expand partition

huww98:
#### What type of PR is this?

/kind bug

growpart is a shell script, and we don't have shell now in our image.

Fixes: af7a885cb5d0c5be15e22618738cf187aea1d9a7
Fixes: #1133 

```release-note
NONE
```

```docs

```

### DIFF
--- a/pkg/disk/sfdisk/expand.go
+++ b/pkg/disk/sfdisk/expand.go
@@ -1,0 +1,53 @@
+package sfdisk
+
+import (
+	"bytes"
+	"context"
+	"fmt"
+	"os/exec"
+
+	utilsos "github.com/kubernetes-sigs/alibaba-cloud-csi-driver/pkg/utils/os"
+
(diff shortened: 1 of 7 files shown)


pkg/disk/sfdisk/expand.go
```diff
@@ -0,0 +1,53 @@
...
+	if err != nil {
+		return fmt.Errorf("failed to lock %s exclusively: %v", disk, err)
+	}
+	defer func() {
+		if err := unix.Flock(fd, unix.LOCK_UN); err != nil {
+			logger.Error(err, "failed to unlock", "fd", fd)
+		}
+	}()
+
+	dump, err := exec.CommandContext(ctx, "sfdisk", "--dump", disk).Output()
```
mowangdk, line 37: why do we need dump before partition expansion

PR #1176, opened 2024-09-26
disk-snapshot: support volume group snapshot APIs

AlbeeSo:
This PR will:
1. bump snapshot client version from v4 to v8 and add default volumegroupsnapshotclass `alibabacloud-disk-group-snapshot`
2. add groupcontrollerserver for disk driver
3. implement volumegroupsnapshot APIs
4. allow users use volumegroupsnapshot by set  feature-gate `EnableVolumeGroupSnapshots=true`

pls test with yaml like:
```
apiVersion: groupsnapshot.storage.k8s.io/v1alpha1
kind: VolumeGroupSnapshot
metadata:
  name: new-group-snapshot-demo
spec:
  volumeGroupSnapshotClassName: alibabacloud-disk-group-snapshot
  source:
    selector:
      matchLabels:
        app: mysql
```

#### What type of PR is this?
/kind feature
/kind api-change

#### Which issue(s) this PR fixes:

Fixes #

#### Special notes for your reviewer:
VolumeGroupSnapshot need K8s >= 1.27
#### Does this PR introduce a user-facing change?

```release-note

```

#### Additional documentation e.g., KEPs (Kubernetes Enhancement Proposals), usage docs, etc.:

```docs

```

PR #1194, opened 2024-10-21
cleanup: remove snapshot events recorder

AlbeeSo:
#### What type of PR is this?
/kind cleanup

external-snapshotter will create event on `volumesnapshotcontent` according to the response from CSI.
so we can remove the event recorder for snapshot.

I test with:
1. create a disk PVC but do not attach to any node.
2. create a volumesnapshot and failed with event like `The specified disk has never been attached to any instance`
3. create an app and mount with this PVC
4. succeed creating a ECS snapshot.

Fixes #

I test with external-snapshotter v8 so it is better
/hold
merged after #1176 

```release-note

```

```docs

```

### DIFF
--- a/pkg/disk/controllerserver.go
+++ b/pkg/disk/controllerserver.go
@@ -518,7 +518,6 @@ func (cs *controllerServer) CreateSnapshot(ctx context.Context, req *csi.CreateS
 
 	params, err := getVolumeSnapshotConfig(req)
 	if err != nil {
-		klog.Errorf("CreateSnapshot:: get volumesnapshot config failed: %v", err)
 		return nil, status.Errorf(codes.InvalidArgument, "get volumesnapshot config failed: %v", err)
 	}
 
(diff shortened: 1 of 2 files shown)


pkg/disk/controllerserver.go
```diff
@@ -573,32 +561,26 @@ func (cs *controllerServer) CreateSnapshot(ctx context.Context, req *csi.CreateS
 		if existsSnapshot.SourceDiskId == req.GetSourceVolumeId() {
 			csiSnapshot, err := formatCSISnapshot(&existsSnapshot)
 			if err != nil {
-				return nil, err
+				klog.Errorf("CreateSnapshot:: format snapshot failed: %v", err)
+				return nil, status.Errorf(codes.Internal, "format snapshot failed: %v", err)
```
huww98, line 565: Since you have modified this, can you also remove all the logs immediately preceding return error? Interceptor will log the error automatically

pkg/disk/controllerserver.go
```diff
@@ -823,7 +779,8 @@ func newListSnapshotsResponse(snapshots []ecs.Snapshot, nextToken string) (*csi.
 	for _, snapshot := range snapshots {
 		csiSnapshot, err := formatCSISnapshot(&snapshot)
 		if err != nil {
-			return nil, err
+			klog.Errorf("ListSnapshots:: format snapshot failed: %v", err)
```
huww98, line 782:
```suggestion
```

huww98 (commented):
Looks good. Remove the final leftover?

huww98:
/unhold
since #1197 has been merged
/lgtm

PR #1203, opened 2024-11-06
Get available disk count from node annotation

iltyty:

```release-note

```

```docs

```

### DIFF
--- a/pkg/disk/utils.go
+++ b/pkg/disk/utils.go
@@ -66,7 +66,10 @@ var (
 	SupportedFilesystemTypes   = sets.New(EXT4_FSTYPE, EXT3_FSTYPE, XFS_FSTYPE, NTFS_FSTYPE)
 )
 
-const DISK_TAG_PREFIX = "diskTags/"
+const (
+	DISK_TAG_PREFIX            = "diskTags/"
+	instanceTypeInfoAnnotation = "alibabacloud.com/instance-type-info"
+)
 
 // DefaultOptions is the struct for access key
 type DefaultOptions struct {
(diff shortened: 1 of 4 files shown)


pkg/disk/utils.go
```diff
@@ -1255,7 +1263,24 @@ func getAttachedCloudDisks(ecsClient cloud.ECSInterface, m metadata.MetadataProv
...
+}
+
+func getAvailableDiskCountFromAnnotation(getNode func() (*v1.Node, error)) (int, error) {
+	node, err := getNode()
+	if err != nil {
+		return 0, err
+	}
+	var typeInfo InstanceTypeInfo
+	err = json.Unmarshal([]byte(node.Annotations[instanceTypeInfoAnnotation]), &typeInfo)
+	return typeInfo.DiskQuantity, err
```
huww98, line 1280: Please add logs about the error, since it is otherwise ignored.

huww98 (approved):
/lgtm
/ok-to-test

huww98:
/feature

Better to add UT for this feature.

pkg/disk/utils_test.go
```diff
@@ -300,6 +300,58 @@ func (m *MockDisks) AddDisk(t testing.TB, path string, diskID []byte) {
...
+func TestGetAvailableDiskCountFromAnnotation(t *testing.T) {
+	tests := []struct {
+		name     string
+		node     *corev1.Node
+		expected int
+		wantErr  bool
+	}{
+		{
+			name:     "Empty node",
+			node:     nil,
```
huww98, line 312: I think we don't expect nil node? How about testing for nodes that don't have this annotation?

BTW, test for getAvailableDiskCount ? So we also cover the fallback behavior.

PR #1207, opened 2024-11-08
fix: limit role-session-name for oss rrsa to 64 chars

AlbeeSo:
#### What type of PR is this?
/kind bug

role-session-name has modified to be no longer than 64 chars.

Fixes #

```release-note

```

```docs

```

### DIFF
--- a/pkg/mounter/ossfs.go
+++ b/pkg/mounter/ossfs.go
@@ -359,7 +359,7 @@ func CleanupOssfsCredentialSecret(ctx context.Context, clientset kubernetes.Inte
 func getRoleSessionName(volumeId, target string) string {
 	name := fmt.Sprintf("ossfs.%s.%s", volumeId, computeMountPathHash(target))
 	if len(name) > MaxRoleSessionNameLimit {
-		name = fmt.Sprintf("ossfs.%s", computeMountPathHash(target))
+		name = name[:MaxRoleSessionNameLimit]
 	}
 	return name
 }
--- a/pkg/mounter/ossfs.go
+++ b/pkg/mounter/ossfs.go
@@ -43,6 +43,8 @@ const (
 
 	PasswdMountDir = "/etc/ossfs"
 	PasswdFilename = "passwd-ossfs"
+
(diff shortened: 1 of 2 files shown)


pkg/mounter/ossfs.go
```diff
@@ -355,7 +357,11 @@ func CleanupOssfsCredentialSecret(ctx context.Context, clientset kubernetes.Inte
 }
 
 func getRoleSessionName(volumeId, target string) string {
-	return fmt.Sprintf("ossfs.%s.%s", volumeId, computeMountPathHash(target))
+	name := fmt.Sprintf("ossfs.%s.%s", volumeId, computeMountPathHash(target))
```
mowangdk, line 360: So the whole role name length depends on the length of volumeId? btw i don't think this is a good idea that we have two pattern names.

pkg/mounter/ossfs_test.go
```diff
@@ -185,3 +185,19 @@ func Test_AddDefaultMountOptions(t *testing.T) {
...
+		target   string
+		wantName string
+	}{
+		{"vol1", "/mnt/target1", "ossfs.vol1." + computeMountPathHash("/mnt/target1")},
+		{"hereisalonglongpvnamethatisalreadylongerthan64ibeleive", "/mnt/target2", "ossfs." + computeMountPathHash("/mnt/target2")},
+	}
+
+	for _, test := range tests {
+		sessionName := getRoleSessionName(test.volumeId, test.target)
+		assert.Equal(t, test.wantName, sessionName)
```
huww98, line 201: Also add an assertion about the length of session name? In case we change the hash algorithm in the future.

pkg/mounter/ossfs.go
huww98: I also think truncating is better.

huww98:
```suggestion
	name := fmt.Sprintf("ossfs.%s.%s", computeMountPathHash(target), volumeId)
```
OK. But how about this? So that we always have a complete hash at least. The hash maybe used as label selector for diagnose.

PR #1209, opened 2024-11-10
oss: allow setup mime.types by configmap

AlbeeSo:
#### What type of PR is this?
/kind feature

allow users setup mime.types with volume scoped, and modify it after CSI startup

Fixes #

```release-note

```

```docs

```

### DIFF
--- a/deploy/chart/templates/rbac.yaml
+++ b/deploy/chart/templates/rbac.yaml
@@ -228,6 +228,10 @@ rules:
   - apiGroups: [""]
     resources: ["nodes/stats"]
     verbs: ["get"]
+  - apiGroups: [""]
+    resources: ["configmaps"]
+    resourceNames: ["ossfs-mime-config"]
+    verbs: ["get"]
 ---
 apiVersion: rbac.authorization.k8s.io/v1
 kind: RoleBinding
--- a/deploy/chart/templates/rbac.yaml
+++ b/deploy/chart/templates/rbac.yaml
@@ -63,6 +63,10 @@ rules:
   - apiGroups: ["storage.alibabacloud.com"]
     resources: ["containernetworkfilesystems"]
     verbs: ["get","list", "watch"]
+  - apiGroups: [""]
+    resources: ["configmaps"]
+    resourceNames: ["ossfs-mime-config"]
+    verbs: ["get"]
 ---
 kind: ClusterRoleBinding
 apiVersion: rbac.authorization.k8s.io/v1
--- a/pkg/mounter/ossfs.go
+++ b/pkg/mounter/ossfs.go
@@ -171,7 +171,7 @@ func (f *fuseOssfs) buildPodSpec(c *FusePodContext, target string) (spec corev1.
 	}
 	*mimeFileVolume.ConfigMap.Optional = true
 
-	spec.Volumes = []corev1.Volume{targetDirVolume, metricsDirVolume, etcDirVolume, mimeFileVolume}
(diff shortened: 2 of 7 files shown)


pkg/mounter/ossfs.go
```diff
@@ -396,3 +443,14 @@ func getPasswdSecretVolume(secretRef string) (secret *corev1.SecretVolumeSource)
 	}
 	return
 }
+
+func hasSetMimeConfigMap(volumeId string) bool {
+	cfg := options.MustGetRestConfig()
+	clientset := kubernetes.NewForConfigOrDie(cfg)
```
huww98, line 449: Please reuse clientset, for rate-limiter and others to work properly.

pkg/mounter/ossfs_test.go
```diff
@@ -111,7 +114,44 @@ func Test_getPasswdSecretVolume(t *testing.T) {
...
+	}
+	responder := httpmock.NewStringResponder(200, mockConfigMapJson)
+	responder = responder.HeaderSet(map[string][]string{
+		"Content-Type": {"application/json"},
+	})
+	httpmock.RegisterResponder("GET", url, responder)
+}
+
 func Test_AddDefaultMountOptions(t *testing.T) {
+	httpmock.Activate()
```
huww98, line 151: Please try to inject a fake clientset into the tested instance, and avoid modifying global state, to enable parallel test in the future.
And, why not use the fake clientset implemented in client-go?

deploy/chart/templates/rbac.yaml
```diff
@@ -228,6 +228,10 @@ rules:
   - apiGroups: [""]
     resources: ["nodes/stats"]
     verbs: ["get"]
+  - apiGroups: [""]
+    resources: ["configmaps"]
+    resourceNames: ["ossfs-mime-config"]
+    verbs: ["get"]
```
huww98, line 234: Will this be called from node or controller? I think it should be controller, because the fuse pod is created there, right? If it is really called from node, it can be blocked by node-restriction if enabled, despite you adding RBAC permission for it.

huww98 (changes requested):
I would design the API like this:
Add two per-PV options:
- mimeVolume: we just pass it through to fuse pod spec.volumes. Should support both hostPath and configMap type volume. The name of configmap should be configurable.
- mimeSubpath: use this path to access the actual mime.types file in the volume

The current design is not desiable because:
- If user have too many volumes, the `ossfs-mime-config` can be very large. It is not scalable.
- The user cannot reuse the same config for multiple volume, making `ossfs-mime-config` even larger. And hard to operate.
- It is hard to authorize config for different volumes to different users (e.g. by RBAC) if we put all of them in the same configMap.

AlbeeSo:
> I would design the API like this: Add two per-PV options:
> 
> * mimeVolume: we just pass it through to fuse pod spec.volumes. Should support both hostPath and configMap type volume. The name of configmap should be configurable.
> * mimeSubpath: use this path to access the actual mime.types file in the volume
> 
> The current design is not desiable because:
> 
> * If user have too many volumes, the `ossfs-mime-config` can be very large. It is not scalable.
> * The user cannot reuse the same config for multiple volume, making `ossfs-mime-config` even larger. And hard to operate.
> * It is hard to authorize config for different volumes to different users (e.g. by RBAC) if we put all of them in the same configMap.

sounds great. but i think a entire volume config is a little bit hard to config? maybe we can just support the configmap way to set up the mime.types. users can just specify the name of cm, then CSI convert the map of data to standard  format for mime.types.

PR #1213, opened 2024-11-13
disk: remove expand auto snapshot

huww98:
#### What type of PR is this?

/kind bug
/kind deprecation

Following up #1182 , remove this function since it is not working and will block resizing.

preserve the function of deleting snapshot on successful resize for one more version, to avoid leaking snapshots.

Fixes #

```release-note
DisableExpandAutoSnapshot feature gate is now locked to true, auto snapshot on resizing is no longer supported
```

```docs

```

### DIFF
--- a/pkg/features/features.go
+++ b/pkg/features/features.go
@@ -48,7 +48,7 @@ var (
 		DiskADController:           {Default: false, PreRelease: featuregate.Alpha},
 		DiskParallelAttach:         {Default: false, PreRelease: featuregate.Alpha},
 		DiskParallelDetach:         {Default: false, PreRelease: featuregate.Alpha},
-		DisableExpandAutoSnapshots: {Default: true, PreRelease: featuregate.Beta},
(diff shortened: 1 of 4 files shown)


pkg/features/features.go
```diff
@@ -46,7 +46,7 @@ var (
 		DiskADController:           {Default: false, PreRelease: featuregate.Alpha},
 		DiskParallelAttach:         {Default: false, PreRelease: featuregate.Alpha},
 		DiskParallelDetach:         {Default: false, PreRelease: featuregate.Alpha},
-		DisableExpandAutoSnapshots: {Default: false, PreRelease: featuregate.Alpha},
```
mowangdk, line 49: Set the featuregate to default first? code remove in next version

huww98: please see #1220

huww98:
Do you want to unhold this?

PR #1218, opened 2024-11-17
disk/ad_slot: return better GRPC error message

huww98:
#### What type of PR is this?

/kind cleanup

expected new message shown in Kubernetes events:
failed to reserve node i-xxxx for attach: still waiting for other disk(s) to finish attach/detach

Fixes #

```release-note
NONE
```

```docs

```


huww98:
I've verified that I can see the expected Kubernetes event with this patch. However, this path is hard to trigger, because requests are not enqueued at the same time. when the currently attaching request is cancelled, the next one kicks in and issue AttachDisk call immediately, and likely receive Conflict response. So, these conflicting AttachDisk calls can be avoided.

Maybe something like #1190 can be helpful, but still not ideal. Because we don't want to delay the GRPC response to avoid potential timeout.

PR #1221, opened 2024-11-19
Make the OSSFS pod name visible in VolumeAttachment

tydra-wang:
…atus

#### What type of PR is this?

/kind feature

Make troubleshooting easiler.

Fixes #

```release-note

```

```docs

```


huww98:
Can replace #954 
But when will you need this? For a certain volume, currently there will be at most one Pod on each node, so we can filter based on node and volume easily.

PR #1222, opened 2024-11-19
Add a feature gate for kubelet stat/summary metrics collector

tydra-wang:
#### What type of PR is this?

/kind cleanup

Fixes #

```release-note

```

```docs

```

### DIFF
--- a/pkg/metric/kubelet_stats_summary_collector.go
+++ b/pkg/metric/kubelet_stats_summary_collector.go
@@ -2,7 +2,6 @@ package metric
 
 import (
 	"encoding/json"
-	"fmt"
 	"io"
 	"net/http"
 	"time"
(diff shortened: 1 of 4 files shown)


pkg/metric/kubelet_stats_summary_collector.go
```diff
@@ -90,7 +90,8 @@ func (c *kubeletStatsSummaryCollector) Update(ch chan<- prometheus.Metric) error
 	defer resp.Body.Close()
 	if resp.StatusCode != http.StatusOK {
 		body, _ := io.ReadAll(resp.Body)
-		return fmt.Errorf("failed to get kubelet stats summary: %d: %q", resp.StatusCode, string(body))
+		klog.V(4).InfoS("Failed to get kubelet stats summary", "status_code", resp.StatusCode, "body", string(body))
```
mowangdk, line 93: Should the klog level be set to 2 to prevent excessive logging of infos?

huww98: v=2 is the default level set in helm.

I think this is not the correct way to hide the error. Maybe we should add a flag to disable this collector completely?

tydra-wang: Added a featuregate for the collector

PR #1225, opened 2024-11-20
disk/snapshot: wait for cut

huww98:
#### What type of PR is this?

/kind bug

CSI requires us to ensure the snapshot is cut before returning. Implementing this by describe the just created snapshot and ensure its CreationTime is set.

Fixes #

```release-note
We now properly ensure the snapshot is cut before creating the VolumeSnapshotContent
```

```docs

```

### DIFF
--- a/pkg/disk/controllerserver.go
+++ b/pkg/disk/controllerserver.go
@@ -577,7 +577,7 @@ func (cs *controllerServer) CreateSnapshot(ctx context.Context, req *csi.CreateS
 	snapshotResponse, err := requestAndCreateSnapshot(ecsClient, params)
 
 	if err != nil {
-		return nil, status.Errorf(codes.Internal, "create snapshot[%s] with sourceId[%s] failed with error: %v", req.Name, req.GetSourceVolumeId(), err)
+		return nil, err
 	}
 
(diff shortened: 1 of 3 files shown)


pkg/disk/controllerserver.go
```diff
@@ -81,10 +85,38 @@ var veasp = struct {
...
 var delVolumeSnap sync.Map
 
+func newSnapshotStatusWaiter() waitstatus.StatusWaiter[ecs.Snapshot] {
+	client := desc.Snapshots{
+		Client: GlobalConfigVar.EcsClient,
+	}
+	if GlobalConfigVar.DiskMultiTenantEnable {
+		waiter := waitstatus.NewSimple(client, clock.RealClock{})
+		waiter.ClientFactory = func(ctx context.Context) (desc.Client[ecs.Snapshot], error) {
+			tenantUserUID := GetTenantUserUID(ctx)
```
mowangdk, line 95: This block is no longer needed since whole multi-tenant feature has been removed

huww98: year, will rebase

huww98: done

PR #1245, opened 2024-12-03
Fix nvme serial id check

mowangdk:
#### What type of PR is this?
/kind bug

Finding series-id of disk logics we have did not match new Nvme rules.  And udev rules may no longer needed.

Fixes #

```release-note

```

```docs

```


huww98:
Please also update the comment above `GetVolumeIDByDevice`

PR #1248, opened 2024-12-04
disk: re-implement StagingTargetPath cleanup logic

huww98:
#### What type of PR is this?

/kind bug

We should unlink the file we created for volumeDevice.

We should check targetPath is not mounted before we check for volumeDevice, in case the user has a file with the same name.

And the overall complexity is greatly reduced. We do not call stat() on the mounted path now, so no need to worry about EIO or similar.

Fixes #

This is necessary for csi-sanity test case to pass.

```release-note
NONE
```

```docs

```

### DIFF
--- a/pkg/disk/nodeserver.go
+++ b/pkg/disk/nodeserver.go
@@ -92,8 +92,6 @@ const (
 	RundSocketDir = "/host/etc/kubernetes/volumes/rund/"
 	// VolumeDirRemove volume dir remove
 	VolumeDirRemove = "/host/etc/kubernetes/volumes/disk/remove"
-	// InputOutputErr tag
-	InputOutputErr = "input/output error"
 	// DiskMultiTenantEnable Enable disk multi-tenant mode
 	DiskMultiTenantEnable = "DISK_MULTI_TENANT_ENABLE"
 	// TenantUserUID tag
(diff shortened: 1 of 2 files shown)


pkg/disk/nodeserver.go
```diff
@@ -682,9 +680,44 @@ func addDiskXattr(diskID string) (err error) {
...
+		case errors.Is(err, unix.ENOENT):
+			return nil
+		case errors.Is(err, unix.EINVAL):
+			// Maybe not mounted, proceed to remove it. If not, unlink will report error.
+		default:
+			return err
+		}
+	}
+
+	errUnlink := unix.Unlink(path)
```
mowangdk, line 707: i didn't think we have ever link the volumedevice, why do we need to unlink it?

huww98: `unix.Unlink` is just another way to `os.Remove` a file. I'm using unix package to be consistent with above 'unix.Umount'

huww98:
Use the just merged `CleanupSimpleMount`

pkg/disk/nodeserver.go
```diff
@@ -681,77 +681,62 @@ func (ns *nodeServer) addDiskXattr(logger klog.Logger, diskID string) (err error
...
 }
 
-// target format: /var/lib/kubelet/plugins/kubernetes.io/csi/pv/pv-disk-1e7001e0-c54a-11e9-8f89-00163e0e78a0/globalmount
-func (ns *nodeServer) NodeUnstageVolume(ctx context.Context, req *csi.NodeUnstageVolumeRequest) (*csi.NodeUnstageVolumeResponse, error) {
-	logger := klog.FromContext(ctx)
-	klog.Infof("NodeUnstageVolume:: Starting to Unmount volume, volumeId: %s, target: %v", req.VolumeId, req.StagingTargetPath)
-
-	if !ns.locks.TryAcquire(req.VolumeId) {
-		return nil, status.Errorf(codes.Aborted, "There is already an operation for %s", req.VolumeId)
+func ensureUnmounted(mounter k8smount.Interface, target string) error {
```
mowangdk: Is here exactly same as ```k8smounter.Unmount```?

pkg/disk/nodeserver.go
```diff
@@ -681,77 +681,62 @@ func (ns *nodeServer) addDiskXattr(logger klog.Logger, diskID string) (err error
...
-				return nil, status.Error(codes.Internal, err.Error())
+func (ns *nodeServer) unmountTargetPath(logger klog.Logger, targetPath, volumeID string) error {
+	err := ns.unixMounter.Unmount(targetPath)
+	if err != nil {
+		switch {
+		case errors.Is(err, unix.ENOENT):
+			logger.V(1).Info("targetPath not exist")
+		case errors.Is(err, unix.EINVAL):
+			// Maybe unmounted, lets check
+			if errCheck := ensureUnmounted(ns.k8smounter, targetPath); errCheck != nil {
```
mowangdk, line 696: The block device mountpoint detection sequence has been changed. Will this cause any issues? Have you tried on blockdevice mode

huww98: Yes, both unit-test and external-storage e2e test.

pkg/disk/mounter/unix_mounter.go
```diff
@@ -0,0 +1,46 @@
...
+		default:
+			return err
+		}
+	}
+
+	errRm := os.Remove(path)
+	if errRm == nil {
+		return nil
+	}
+	if err != nil {
```
mowangdk: Why is this error check necessary?

huww98: Because we ignored `errors.Is(err, unix.EINVAL):` case above. But we still want to log this error if rm failed.

huww98: The condition was inverted. Fixed now.

PR #1251, opened 2024-12-05
helm: parse featureGates as map

AlbeeSo:
CSI includes kinds of feature-gates, which may lead different configuration on helm templates. For example, to minimize rbac rules, and to set flags for sidecars only when related feature-gates enabled.
This PR involves a `parseFeatureGates` function in helm to help this and apply to `EnableVolumeGroupSnapshots` feature. 
#### What type of PR is this?
/kind feature

Fixes #

```release-note

```

```docs

```

### DIFF
--- a/deploy/chart/templates/_utils.tpl
+++ b/deploy/chart/templates/_utils.tpl
@@ -57,3 +57,18 @@
   value: {{ $d }}
 {{- end -}}
 {{- end -}}
+
+{{- define "parseFeatureGates" -}}
+{{- $featureGates := list -}}
+{{- range $index, $featureGate := splitList "," .  }}
+  {{- $featureGateKV := splitList "=" $featureGate }}
+  {{- if eq (len $featureGateKV) 2 -}}
+    {{- $key := trim (index $featureGateKV 0) -}}
+    {{- $value := lower (trim (index $featureGateKV 1) ) -}}
+    {{- if or (eq $value "true") (eq $value "t") -}}
+      {{- $featureGates = append $featureGates $key -}}
+    {{- end -}}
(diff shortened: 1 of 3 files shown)


huww98:
This will pose extra challenge to our ACK release tools, which cannot handle such complex.

And I doubt if this is necessary. Some extra RBAC for new CRs should be fine. And feature gates are expected to be true for everyone in the future, so all necessary permissions should be acceptable for everyone. Feature gate is just for cluster admin to acknowledge that this feature is not stable yet.

huww98:
Error: template: alibaba-cloud-csi-driver/templates/rbac.yaml:47:45: executing "alibaba-cloud-csi-driver/templates/rbac.yaml" at <$featureGates>: wrong type for value; expected string; got []interface {}

AlbeeSo:
> This will pose extra challenge to our ACK release tools, which cannot handle such complex logic.
> 
> And I doubt if this is necessary. Some extra RBAC for new CRs should be fine. And feature gates are expected to be true for everyone in the future, so all necessary permissions should be acceptable for everyone. Feature gate is just for cluster admin to acknowledge that this feature is not stable yet.

@huww98 Flags on sidecars by default may cause unexpected warning logs. Let users set feature-gates by hand according to the document is a way to announce them some extra RBACs are added. Even permissions are all acceptable for everyone, they are unnecessary.

AlbeeSo:
> Error: template: alibaba-cloud-csi-driver/templates/rbac.yaml:47:45: executing "alibaba-cloud-csi-driver/templates/rbac.yaml" at <$featureGates>: wrong type for value; expected string; got []interface {}

Done.

PR #1262, opened 2024-12-11
feat(nas): automatic CNFS switching when mounting NAS volumes

iltyty:
#### What type of PR is this?
/kind feature

Fixes #

```release-note

```

```docs

```

### DIFF
--- a/pkg/nas/nodeserver.go
+++ b/pkg/nas/nodeserver.go
@@ -39,14 +39,17 @@ import (
 	"github.com/kubernetes-sigs/alibaba-cloud-csi-driver/pkg/utils/rund/directvolume"
 	"google.golang.org/grpc/codes"
 	"google.golang.org/grpc/status"
+	v1 "k8s.io/api/core/v1"
+	"k8s.io/client-go/tools/record"
 	"k8s.io/klog/v2"
 	mountutils "k8s.io/mount-utils"
 )
 
 type nodeServer struct {
-	config  *internal.NodeConfig
-	mounter mountutils.Interface
-	locks   *utils.VolumeLocks
+	config   *internal.NodeConfig
+	mounter  mountutils.Interface
+	locks    *utils.VolumeLocks
+	recorder record.EventRecorder
 	common.GenericNodeServer
 }
 
@@ -60,8 +63,9 @@ func newNodeServer(config *internal.NodeConfig) *nodeServer {
 	}
 
 	ns := &nodeServer{
-		config: config,
-		locks:  utils.NewVolumeLocks(),
+		config:   config,
+		locks:    utils.NewVolumeLocks(),
+		recorder: utils.NewEventRecorder(),
 		GenericNodeServer: common.GenericNodeServer{
 			NodeID: config.NodeName,
 		},
@@ -130,7 +134,10 @@ const (
 	//NFSClient
 	NFSClient = "nfsclient"
 	//NativeClient
-	NativeClient = "nativeclient"
+	NativeClient                                = "nativeclient"
+	cnfsAlwaysFallbackEventTmpl                 = "CNFS automatically switched from %s to %s."
+	cnfsIfConnectFailedFallbackEventTmpl        = "Due to network issues, CNFS automatically switched from %s to %s."
+	cnfsIfMountTargetUnhealthyFallbackEventTmpl = "Due to mount target inactive, CNFS automatically switched from %s to %s."
 )
 
 func validateNodePublishVolumeRequest(req *csi.NodePublishVolumeRequest) error {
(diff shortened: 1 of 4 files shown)


pkg/utils/util.go
```diff
@@ -478,10 +478,10 @@ func GetPodFromContextOrK8s(ctx context.Context, client kubernetes.Interface, re
 	if ok {
 		return pod, nil
 	}
-	return getPodFromK8s(ctx, client, req)
+	return getPodFromK8s(client, req)
 }
 
-func getPodFromK8s(ctx context.Context, client kubernetes.Interface, req *csi.NodePublishVolumeRequest) (*v1.Pod, error) {
```
huww98, line 484: How about passing the ctx to the `Get()` below, instead of removing it?

pkg/nas/nodeserver.go
```diff
@@ -455,6 +462,57 @@ func (ns *nodeServer) NodePublishVolume(ctx context.Context, req *csi.NodePublis
...
+	if cnfs == nil {
+		return false
+	}
+	switch cnfs.Spec.Fallback.Strategy {
+	case v1beta1.FallbackStrategyAlways:
+		return true
+	case v1beta1.FallbackStrategyIfConnectFailed:
+		server := cnfs.Spec.Parameters.Server
+		dialer := net.Dialer{Timeout: 5 * time.Second}
+		_, err := dialer.Dial("tcp", server+":2049")
```
huww98, line 486: Better to use `DialContext` and explicitly close the returned conn object.

pkg/nas/nodeserver.go
```diff
@@ -455,6 +462,57 @@ func (ns *nodeServer) NodePublishVolume(ctx context.Context, req *csi.NodePublis
...
+	case v1beta1.FallbackStrategyIfMountTargetUnhealthy:
+		return cnfs.Status.Status == v1beta1.StatusUnavailable
+	}
+	return false
+}
+
+func (ns *nodeServer) fallbackCNFSAndRecord(ctx context.Context, req *csi.NodePublishVolumeRequest, cnfs *v1beta1.ContainerNetworkFileSystem) (*v1beta1.ContainerNetworkFileSystem, error) {
+	oldName, newName := cnfs.Name, cnfs.Spec.Fallback.Name
+	pod, err := utils.GetPodFromContextOrK8s(ctx, ns.config.KubeClient, req)
+	if err != nil {
```
huww98, line 497: I think we can just skip the event if this fails?

pkg/nas/nodeserver_test.go
```diff
@@ -0,0 +1,202 @@
...
+		panic(err)
+	}
+	defer listener.Close()
+
+	for {
+		conn, _ := listener.Accept()
+		if conn != nil {
+			conn.Close()
+		}
+	}
```
huww98, line 53: This will not terminate after the test case finishes. How about making use of `t.Cleanup()`?

pkg/nas/nodeserver_test.go
```diff
@@ -0,0 +1,202 @@
...
+					oldCNFSName, newCNFSName := tt.primaryCNFS.Name, tt.fallbackCNFS.Name
+					msg := <-eventRecorder.Events
+					switch tt.primaryCNFS.Spec.Fallback.Strategy {
+					case v1beta1.FallbackStrategyAlways:
+						assert.Contains(t, msg, fmt.Sprintf(cnfsAlwaysFallbackEventTmpl, oldCNFSName, newCNFSName))
+					case v1beta1.FallbackStrategyIfConnectFailed:
+						assert.Contains(t, msg, fmt.Sprintf(cnfsIfConnectFailedFallbackEventTmpl, oldCNFSName, newCNFSName))
+					case v1beta1.FallbackStrategyIfMountTargetUnhealthy:
+						assert.Contains(t, msg, fmt.Sprintf(cnfsIfMountTargetUnhealthyFallbackEventTmpl, oldCNFSName, newCNFSName))
+					}
```
huww98, line 195: How about add a expectedEvent field to `tests`, instead of copy pasting a large segment of code?

mowangdk:
@iltyty ut test failed, ptal

pkg/nas/nodeserver_test.go
```diff
@@ -0,0 +1,224 @@
...
+
+func fakeCNFS(name, status, server, fallbackName string, fallbackStrategy v1beta1.FallbackStrategy) *v1beta1.ContainerNetworkFileSystem {
+	return &v1beta1.ContainerNetworkFileSystem{
+		ObjectMeta: metav1.ObjectMeta{
+			Name: name,
+		},
+		Status: v1beta1.ContainerNetworkFileSystemStatus{
+			Status: status,
+		},
+		Spec: v1beta1.ContainerNetworkFileSystemSpec{
```
mowangdk, line 66: I think we need to add a test case that fallback is empty struct

pkg/nas/nodeserver_test.go
```diff
@@ -0,0 +1,231 @@
...
+		select {
+		case <-ctx.Done():
+			wg.Done()
+			return
+		default:
+			l, ok := listener.(*net.TCPListener)
+			if !ok {
+				return
+			}
+			l.SetDeadline(time.Now().Add(100 * time.Millisecond))
```
huww98: Do we still need this one? I think it is enough to cancel the `Accept` by calling `Close` on listener

PR #1272, opened 2024-12-21
Read configmap from volume mount

huww98:
#### What type of PR is this?

/kind cleanup

Avoid RBAC permission to configmap and save some APIServer call. The permission to get ack-cluster-profile is moved to a separate role, not assuming we are installed in kube-system.

Fixes #

```release-note
NONE
```

```docs

```


huww98:
try to trigger golangci-lint again

PR #1283, opened 2024-12-25
disk: increase MaxIdleConns for OpenAPI from 2 to 100

huww98:
#### What type of PR is this?

/kind cleanup

We are only talking to very few hosts, so the per-host config can be large. This should reduce the number of sockets in TIME_WAIT caused by frequently re-creating TCP connection.

Fixes #

```release-note
NONE
```

```docs

```

### DIFF
--- a/pkg/disk/utils.go
+++ b/pkg/disk/utils.go
@@ -93,6 +93,22 @@ type InstanceTypeInfo struct {
 	DiskQuantity int
 }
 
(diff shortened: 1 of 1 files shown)


mowangdk:
Have you tested it under the stress environment to ensure it meets our requirements?

huww98:
Not yet. Queued for the next test.

PR #1287, opened 2024-12-27
Modified use newToken configuration name

mowangdk:
#### What type of PR is this?
/kind cleanup

Modified use newToken configuration name

Fixes #

```release-note

```

```docs

```

### DIFF
--- a/deploy/chart/values.yaml
+++ b/deploy/chart/values.yaml
@@ -32,8 +32,6 @@ nodePools:
           operator: NotIn
           values:
             - virtual-kubelet
-    deploy:
-      newPluginToken: true
 
 defaultStorageClass:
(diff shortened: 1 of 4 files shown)


deploy/chart/values.yaml
```diff
@@ -54,7 +52,7 @@ deploy:
   # and the storage-controller will handle the cleanup of the subpath in NAS filesystem.
   withStorageController: false
 
-  newProvisionerToken: true
+  ramTokenV2: true
```
huww98, line 55:
```suggestion
  ramToken: v2
```
if we ever have a v3 in the future, we want to prevent user from enable v2 and v3 simultaneously.

PR #1289, opened 2024-12-27
disk: fix sfdisk for in-use disk

huww98:
#### What type of PR is this?

/kind bug

We are getting the error now:
> This disk is currently in use - repartitioning is probably a bad idea.
> Umount all file systems, and swapoff all swap partitions on this disk.
> Use the --no-reread flag to suppress this check.

Fixes: 1191e248654f7a0add1ae2b8aee474e623a5c719
Fixes #1192 

```release-note
Fixed expanding of disk with partition
```

```docs

```

### DIFF
--- a/pkg/disk/sfdisk/expand_test.go
+++ b/pkg/disk/sfdisk/expand_test.go
@@ -6,36 +6,70 @@ import (
 	"errors"
 	"os"
 	"os/exec"
+	"strconv"
 	"strings"
 	"testing"
 
 	utilsos "github.com/kubernetes-sigs/alibaba-cloud-csi-driver/pkg/utils/os"
 	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
 )
+
+func runCommand(t *testing.T, name string, arg ...string) string {
+	cmd := exec.Command(name, arg...)
+	output, err := cmd.Output()
+	require.NoError(t, utilsos.ErrWithStderr(err))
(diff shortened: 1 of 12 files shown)


huww98:
The unit test TestExpandPartition is hard to run in prow. I've tested it manually on ECS.

pkg/disk/sfdisk/expand_test.go
```diff
@@ -6,36 +6,70 @@ import (
...
 	cmd := exec.Command("sfdisk", testImage)
 	cmd.Stdin = bytes.NewReader([]byte("label: gpt\n,\n")) // create a single partition
 	result, err := cmd.Output()
-	assert.NoError(t, utilsos.ErrWithStderr(err))
+	require.NoError(t, utilsos.ErrWithStderr(err))
 	t.Logf("create partition success: %s", string(result))
 
-	assert.NoError(t, os.Truncate(testImage, 1<<24)) // expand to 16MB
-	assert.NoError(t, ExpandPartition(context.Background(), testImage, "1"))
+	result, err = exec.Command("losetup", "--find", testImage, "--show", "--partscan").Output()
```
mowangdk, line 45: Can we add a concurrency test to verify the effectiveness of the lock parameters?

huww98: `--lock` is actually intended to protect us from concurrent access with udev.

PR #1307, opened 2025-01-24
disk: wait for device to appear on attach

huww98:
#### What type of PR is this?

/kind cleanup
/kind flake

Even after OpenAPI returns Attached status, the device may not be available in OS immediately. We wait for 10s to reduce noisy error events.

Fixes #
Reduce flakes in csi-sanity, which will not retry.

```release-note
NONE
```

```docs

```

### DIFF
--- a/pkg/disk/cloud.go
+++ b/pkg/disk/cloud.go
@@ -42,6 +42,7 @@ import (
 	"google.golang.org/grpc/codes"
 	"google.golang.org/grpc/status"
 	v1 "k8s.io/api/core/v1"
+	utilerrors "k8s.io/apimachinery/pkg/util/errors"
 	"k8s.io/apimachinery/pkg/util/sets"
 	"k8s.io/apimachinery/pkg/util/wait"
 	"k8s.io/klog/v2"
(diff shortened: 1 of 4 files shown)


pkg/disk/cloud.go
```diff
@@ -234,7 +234,7 @@ func (ad *DiskAttachDetach) attachDisk(ctx context.Context, diskID, nodeID strin
 
 	// step 5: diff device with previous files under /dev
 	if fromNode {
-		device, err := DefaultDeviceManager.GetDeviceByVolumeID(diskID)
+		device, err := DefaultDeviceManager.WaitDevice(ctx, diskID)
```
mowangdk, line 237: Wait 10 seconds for each disk without a series id?

mowangdk: Better to add logs to record err message

huww98: > Wait 10 seconds for each disk without a series id?

Currently yes. This can be resolved by #1199 , which explicitly distinguishes disks with or without serial number.

huww98: Added logs for all abnormal code path.

PR #1314, opened 2025-02-05
Add typo check

mowangdk:
#### What type of PR is this?
/kind cleanup

Add typo check in github action

Fixes #

```release-note

```

```docs

```

### DIFF
--- a/.github/workflows/typos.yaml
+++ b/.github/workflows/typos.yaml
@@ -14,8 +14,3 @@ jobs:
       with:
         files: ./
         config: ./_typos.toml
-
-    - name: Writes changes in the local checkout
-      uses: crate-ci/typos@v1.29.5
-      with:
-        write_changes: true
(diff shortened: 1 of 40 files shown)


huww98:
How about using [golangci-lint](https://golangci-lint.run/usage/linters/#misspell)? It is also used by Kubernetes.

mowangdk:
> How about using [golangci-lint](https://golangci-lint.run/usage/linters/#misspell)? It is also used by Kubernetes.

I've found that golint-ci is limited in the number of errors it can scan, and I scan the kubernetes repository for a lot of issues using typos. for example:

```
error: `Permssion` should be `Permission`
  --> ./pkg/volume/volume_linux_test.go:87:8
   |
87 |    skipPermssion:       false,
   |        ^^^^^^^^^
```

huww98:
OK, golangci-lint seems using a hard coded dictionary that records both correct and incorrect word, and it is not updated https://github.com/golangci/misspell/blob/master/words.go

PR #1319, opened 2025-02-13
Fix golangci lint error

mowangdk:
#### What type of PR is this?
/kind cleanup

Fix golangci-lint error

Fixes #

```release-note

```

```docs

```


.golangci.yaml
```diff
@@ -0,0 +1,55 @@
...
+  uniq-by-line: false
+
+linters-settings:
+  errcheck:
+    # exclude defer functions
+    exclude-functions:
+      - os.Remove
+      - os.(*File).Close
+      - syscall.Munmap
+      - io.Closer.Close<sub index="8" url="https://golangci-lint.run/usage/linters/" title="Linters | golangci-lint" snippet="errcheck. Errcheck is a program for checking for unchecked errors in Go code. These unchecked errors can be critical bugs in some cases."></sub>
```
huww98, line 54: This looks strange. Copy-paste error?

pkg/utils/util.go
```diff
@@ -788,7 +788,7 @@ func FormatNewDisk(readOnly bool, source, fstype, target string, mkfsOptions, mo
 	_, err := diskMounter.Exec.Command("mkfs."+fstype, args...).CombinedOutput()
 	if err == nil {
 		// the disk has been formatted successfully try to mount it again.
-		klog.Infof("Disk format succeeded, pvName: %s elapsedTime: %+v ms", pvName, time.Now().Sub(startT).Milliseconds())
+		klog.Infof("Disk format succeeded, pvName: %s elapsedTime: %+v ms", pvName, time.Since(startT).Microseconds())
```
huww98, line 791: I suppose this change is not intended? because the unit in the message is not changed.

```suggestion
		klog.Infof("Disk format succeeded, pvName: %s elapsedTime: %+v ms", pvName, time.Since(startT).Milliseconds())
```

huww98:
Many errors are explicitly ignored now. This can make the errors even harder to find. If the error handling is not obvious, maybe just leave them for future works? Just turn on the incremental scan.

I think every time we ignore an error, there should be a really good reason, and have a comment explaining that.

mowangdk:
> Many errors are explicitly ignored now. This can make the errors even harder to find. If the error handling is not obvious, maybe just leave them for future works? Just turn on the incremental scan.
> 
> I think every time we ignore an error, there should be a really good reason, and have a comment explaining that.

I've reviewed every omit error and believe they can be omitted; feel free to double-check. @huww98

PR #1325, opened 2025-02-18
[v1.26] disk: avoid concurrent operation on the same disk

huww98:
#### What type of PR is this?

/kind bug

If the umount takes extremely long time and the NodeUnstageVolume RPC times out, kubelet may invoke NodeStageVolume when it retries and we may mistake an unmounting path as a good one. This lock only protects us in one process. But fortunately, when restarting CSI, Kubernetes will wait for the previous process to exit before starting the new one, even if it stucks in uninterruptible syscall (D state). So, we will not have multiple processes on the same node and the lock should be safe enough.

Note that there is still small chance that the previous attach/detach OpenAPI is invoked by the terminated CSI process, but the state returned by DescribeDisks is not updated yet. We should fix this by switch attach/detach to controller.

This is a backport commit. We backport this to workaround another edge case in QEMU/Linux. If we call DetachDisk while unmounting, the disk will stuck at Detaching state, then QEMU will issue multiple interrupts to retry detaching. After unmount finished, Linux kernel will eject the PCI slot multiple times in response to the interrupts. But that may eject the disk that was just inserted to the same slot.  This commit will ensure we don't call DetachDisk while unmounting in UnstageVolume.

Note that we may still call DetachDisk in DeleteVolume, which may be triggered by old KCM if unmounting takes over 6min. And this is still possible to trigger the above edge case.

Fixes #

backport of #1170 (partial)

```release-note
Fix unintended disk detach or failed attach when the unmounting is extremely slow
```

```docs

```

### DIFF
--- a/pkg/disk/nodeserver.go
+++ b/pkg/disk/nodeserver.go
@@ -45,6 +45,7 @@ type nodeServer struct {
 	k8smounter        k8smount.Interface
 	clientSet         *kubernetes.Clientset
 	ad                DiskAttachDetach
+	locks             *utils.VolumeLocks
 	*csicommon.DefaultNodeServer
 }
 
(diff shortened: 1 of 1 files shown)


mowangdk:
/lgtm

Please provide the customer with a testing version, I will merge it upon it has been verifed.

PR #1329, opened 2025-02-23
disk: create block mount file after attach

huww98:
#### What type of PR is this?

/kind bug

We should not leave it over if the attach failed

Fixes #

kubelet will do `os.RemoveAll()`, so this will not affect Kubernetes users. But we'd better not do it.

On old CSI version, this will block NodeUnstageVolume, but that is already fixed by #912 .

```release-note
NONE
```

```docs

```

### DIFF
--- a/pkg/disk/nodeserver.go
+++ b/pkg/disk/nodeserver.go
@@ -510,6 +510,11 @@ func (ns *nodeServer) NodeStageVolume(ctx context.Context, req *csi.NodeStageVol
 	targetPath := req.StagingTargetPath
 	// targetPath format: /var/lib/kubelet/plugins/kubernetes.io/csi/pv/pv-disk-1e7001e0-c54a-11e9-8f89-00163e0e78a0/globalmount
 
+	if err := os.MkdirAll(targetPath, 0755); err != nil {
+		klog.Errorf("NodeStageVolume: create volume %s path %s error: %v", req.VolumeId, targetPath, err)
+		return nil, status.Error(codes.Internal, err.Error())
+	}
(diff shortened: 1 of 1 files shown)


pkg/disk/nodeserver.go
```diff
@@ -625,6 +619,10 @@ func (ns *nodeServer) NodeStageVolume(ctx context.Context, req *csi.NodeStageVol
 
 	// Block volume not need to format
 	if isBlock {
+		if err := ns.mounter.EnsureBlock(targetPath); err != nil {
```
mowangdk, line 622: We must initialize targetPath before using it as a parameter in CheckDeviceAvailable.

pkg/disk/nodeserver.go
```diff
@@ -510,6 +510,11 @@ func (ns *nodeServer) NodeStageVolume(ctx context.Context, req *csi.NodeStageVol
 	targetPath := req.StagingTargetPath
 	// targetPath format: /var/lib/kubelet/plugins/kubernetes.io/csi/pv/pv-disk-1e7001e0-c54a-11e9-8f89-00163e0e78a0/globalmount
 
+	if err := os.MkdirAll(targetPath, 0755); err != nil {
```
mowangdk, line 513: I remember we can't put ensureDir before the runD check, is there any particular reason why we have to put it there?

pkg/disk/nodeserver.go
huww98: Why? I don't see anything that requires that path to exist on disk.

PR #1330, opened 2025-02-24
disk: only use diff for disks without serial number

huww98:
#### What type of PR is this?

/kind cleanup

Eliminate the risk of finding the wrong disk. It may take a while for the disk device to appear after the OpenAPI returns in_use. e.g. when the previous detach is blocked, the attach will also be blocked.

Fixes #

```release-note
Only use diff to find disks without serial number
```

```docs

```


pkg/disk/controllerserver.go
```diff
@@ -395,13 +395,17 @@ func (cs *controllerServer) ControllerPublishVolume(ctx context.Context, req *cs
...
 	if err != nil {
 		klog.Errorf("ControllerPublishVolume: attach disk: %s to node: %s with error: %s", req.VolumeId, req.NodeId, err.Error())
 		return nil, err
 	}
-	klog.Infof("ControllerPublishVolume: Successful attach disk: %s to node: %s", req.VolumeId, req.NodeId)
-	return &csi.ControllerPublishVolumeResponse{}, nil
+	klog.Infof("ControllerPublishVolume: successfully attached disk: %s (serial %s) to node: %s", req.VolumeId, serial, req.NodeId)
+	return &csi.ControllerPublishVolumeResponse{
+		PublishContext: map[string]string{
+			PUBLISH_CONTEXT_SERIAL: serial,
```
mowangdk, line 406: Can we ensure that serial id returned means we have symlink mapping exists in node?

huww98: I think so. I've asked ESC colleagues to create some disks without serial number and tested locally.

PR #1335, opened 2025-02-28
metadata: if value in cluster profile is empty, return error

AlbeeSo:
In Private Cloud, cluster profile only has clusterId and resourceGroupId key.
ack-cluster-profile provider should return error to avoid a nil string, and allow the following providers to keep trying retrieve the values.
#### What type of PR is this?
/kind bug

Fixes #

```release-note

```

```docs

```

### DIFF
--- a/pkg/cloud/metadata/ack_cluster_profile.go
+++ b/pkg/cloud/metadata/ack_cluster_profile.go
@@ -27,7 +27,9 @@ func NewProfileMetadata(client kubernetes.Interface) (*ProfileMetadata, error) {
 
 func (m *ProfileMetadata) Get(key MetadataKey) (string, error) {
 	if key, ok := MetadataProfileDataKeys[key]; ok {
-		return m.profile.Data[key], nil
+		if m.profile.Data[key] != "" {
+			return m.profile.Data[key], nil
+		}
 	}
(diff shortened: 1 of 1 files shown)


pkg/cloud/metadata/ack_cluster_profile.go
```diff
@@ -27,7 +27,9 @@ func NewProfileMetadata(client kubernetes.Interface) (*ProfileMetadata, error) {
 
 func (m *ProfileMetadata) Get(key MetadataKey) (string, error) {
 	if key, ok := MetadataProfileDataKeys[key]; ok {
-		return m.profile.Data[key], nil
+		if m.profile.Data[key] != "" {
```
mowangdk, line 30: It's unreasonable to return ErrUnknownMetadataKey for an empty value; please use a new error code instead.

huww98: I think this is appropriate. Golang returns empty string if the key does not exist in the map. In this case,  ErrUnknownMetadataKey will trigger fallback to another source, which is expected.

AlbeeSo: Just "unknown" for this source as keys are not found in cluster-profile.

PR #1336, opened 2025-03-03
fix: use long form for flags when create ossfs pod to keep conpatible with pflag

AlbeeSo:
Different with `flag`, `pflag` will see a flag like `-socket` as  `-s`.
Use flags in long form to keep compatible with the `pflag`, it still works with the older  mount-proxy(ossfs) version.
#### What type of PR is this?
/kind bug

Fixes #

```release-note

```

```docs

```


huww98:
/lgtm
fixes: 447d95557c1e28230f24bcc3a70492b9e1788607
-v can still use the short form, though

PR #1347, opened 2025-03-21
utils: fix serviceType log

huww98:
Fixes: 5defa961b54a444feef153e87cf6706e4f4a81b9

#### What type of PR is this?

/kind bug

Fixes #

```release-note
NONE
```

```docs

```

### DIFF
--- a/pkg/utils/util.go
+++ b/pkg/utils/util.go
@@ -137,10 +137,10 @@ func GetServiceType(runAsController, runControllerService, runNodeService bool) 
 	if serviceType == 0 {
 		klog.Warning("no service type activated, this configuration may not be useful")
 	}
-	if serviceType&Controller == 0 {
+	if serviceType&Controller != 0 {
 		klog.Infof("activate CSI controller service")
 	}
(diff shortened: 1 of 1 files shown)


pkg/utils/util.go
```diff
@@ -137,10 +137,10 @@ func GetServiceType(runAsController, runControllerService, runNodeService bool)
 	if serviceType == 0 {
 		klog.Warning("no service type activated, this configuration may not be useful")
 	}
-	if serviceType&Controller == 0 {
+	if serviceType&Controller != 0 {
```
mowangdk, line 140: Just print the serviceType; the log isn't valuable enough for two conditions.

huww98: serviceType is a number. We need to implement String() for it to get meaningful output. And we still need branch there, maybe more complex, since we can enable controller and node service at the same time.

Constant log message is also friendly when searching for the message in code.

mowangdk: all right, I thought it was simplicity, the possibilities are solely 1, 2, and 3.

PR #1355, opened 2025-04-01
[bug]Forbid empty zoneId request for disk

mowangdk:
#### What type of PR is this?
/kind bug

Forbid empty zoneId request for disk

Fixes #

```release-note

```

```docs

```

### DIFF
--- a/pkg/disk/utils.go
+++ b/pkg/disk/utils.go
@@ -27,6 +27,7 @@ import (
 	"os"
 	"path"
 	"path/filepath"
+	"slices"
 	"strconv"
 	"strings"
 	"time"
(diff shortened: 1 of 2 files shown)


huww98:
How about regional ESSD? which does not require zoneID

pkg/disk/utils.go
```diff
@@ -563,6 +563,12 @@ func getDiskVolumeOptions(req *csi.CreateVolumeRequest) (*diskVolumeArgs, error)
 	if err != nil {
 		return nil, fmt.Errorf("illegal required parameter type: %s", volOptions["type"])
 	}
+
+	if len(diskType) != 1 || diskType[0] != DiskRegional {
```
huww98, line 567: It is better to check CategoryDesc.Regional, in case we have more regional disk category in the future.

```suggestion
	if slices.ContainsFunc(diskType, func(t Category) bool {return !AllCategories[t].Regional}) {
```

PR #1357, opened 2025-04-03
Add cnfs-nas-daemon support in mount-proxy

tydra-wang:
#### What type of PR is this?

/kind feature

Fixes #

```release-note

```

```docs

```

### DIFF
--- a/build/mount-proxy/Dockerfile
+++ b/build/mount-proxy/Dockerfile
@@ -9,6 +9,8 @@ RUN --mount=type=bind,target=. \
     go build -o /out/csi-mount-proxy-server ./cmd/mount-proxy-server && \
     go build -o /out/csi-mount-proxy-client ./cmd/mount-proxy-client
 
+#################### OSSFS 1.91 MOUNT-PROXY IMAGE ####################
+
 FROM registry-cn-hangzhou.ack.aliyuncs.com/dev/alinux:3-update as oss
 ARG TARGETPLATFORM
 ARG OSSFS_VERSION=v1.91.5.ack.1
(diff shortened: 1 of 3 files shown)


build/mount-proxy/Dockerfile
```diff
@@ -60,3 +66,28 @@ RUN ln -s /host/etc/mime.types /etc/mime.types
...
+    case "${TARGETPLATFORM}" in \
+        linux/amd64) ARCH="x86_64" ;; \
+        # linux/arm64) ARCH="aarch_64" ;; \
+        *) echo "unsupported platform"; exit 1 ;; \
+    esac; \
+    yum install -y util-linux procps tini; \
+    yum install -y https://aliyun-alinas-eac.oss-cn-beijing.aliyuncs.com/aliyun-alinas-utils-${ALINAS_UTILS_VERSION}.noarch.rpm; \
+    yum install -y https://aliyun-alinas-eac.oss-cn-beijing.aliyuncs.com/alinas-efc-${EFC_VERSION}.${ARCH}.rpm; \
+    yum clean all
+RUN cp -r /etc/aliyun /etc/aliyun-defaults
```
mowangdk, line 88: What is the purpose of copying the binary?

deploy/chart/templates/plugin.yaml
```diff
@@ -205,6 +205,8 @@ spec:
 {{- end }}
             - mountPath: /host/run/csi-tool
               name: run-csi
+            - mountPath: /run/cnfs
```
mowangdk, line 208: Adding host prefix path(/host) to  align with other host paths.

build/mount-proxy/Dockerfile
mowangdk: Add some explaination

deploy/chart/templates/plugin.yaml
tydra-wang: You can think of it as similar to /var/lib/kubelet, and it won’t cause conflicts.

If we change this, we’ll also need to update the constant `defaultAlinasMountProxySocket = "/run/cnfs/alinas-mounter.sock"` in the Go code.

build/mount-proxy/Dockerfile
tydra-wang: Added to comments

PR #1360, opened 2025-04-10
re-enable metric for controller

huww98:
this seems to be an error while resolving conflict

Fixes: 5defa96
Fixes: #993 

#### What type of PR is this?

/kind bug

Fixes #

```release-note
NONE
```

```docs

```

### DIFF
--- a/main.go
+++ b/main.go
@@ -254,7 +254,7 @@ func main() {
 	csiMux := http.NewServeMux()
 	csiMux.HandleFunc("/healthz", healthHandler)
 	klog.Infof("Metric listening on address: /healthz")
-	if enableMetric && serviceType&utils.Node != 0 {
+	if enableMetric {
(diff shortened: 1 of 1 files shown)


main.go
```diff
@@ -254,7 +254,7 @@ func main() {
 	csiMux := http.NewServeMux()
 	csiMux.HandleFunc("/healthz", healthHandler)
 	klog.Infof("Metric listening on address: /healthz")
-	if enableMetric && serviceType&utils.Node != 0 {
```
mowangdk, line 257: So we only enabled metrics for node before?

huww98: Metric for controller is disabled when I write #993, but enabled afterwards. I fail to remove this when rebasing.

huww98:
#1142 is merged after #993 actually, so we should never have metrics enabled for controller before

/remove-kind bug
/kind feature

PR #1359, opened 2025-04-07
oss: Skip tls verify by default for private cloud environments

iawia002:
#### What type of PR is this?

/kind bug

#### What this PR does / why we need it:

The certificate for the OSS service in the private cloud environment is self-signed, which prevents the mounting of the OSS PV in the private cloud environment:

Screenshot 2025-04-07 at 14 40 52

#### Which issue(s) this PR fixes:

Fixes #

#### Does this PR introduce a user-facing change?

```release-note
Skip tls verify by default for private cloud environments.
```

#### Additional documentation e.g., KEPs (Kubernetes Enhancement Proposals), usage docs, etc.:

```docs

```

PR #1362, opened 2025-04-11
oss: support passing mount options through environment variables

iawia002:
#### What type of PR is this?

/kind feature

Another solution to the problem described in #1359.

close #1359 

Fixes #

```release-note
oss: support passing mount options through environment variables.
```

```docs

```

### DIFF
--- a/pkg/mounter/ossfs.go
+++ b/pkg/mounter/ossfs.go
@@ -12,9 +12,6 @@ import (
 	"strings"
 
 	"github.com/alibabacloud-go/tea/tea"
-	"github.com/kubernetes-sigs/alibaba-cloud-csi-driver/pkg/cloud/metadata"
-	"github.com/kubernetes-sigs/alibaba-cloud-csi-driver/pkg/features"
-	"github.com/kubernetes-sigs/alibaba-cloud-csi-driver/pkg/utils"
 	corev1 "k8s.io/api/core/v1"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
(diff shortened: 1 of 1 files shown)


pkg/oss/nodeserver.go
```diff
@@ -392,6 +394,9 @@ func (o *Options) MakeMountOptionsAndAuthConfig(m metadata.MetadataProvider, vol
 	}
 	mountOptions = append(mountOptions, authOptions...)
 
+	defaultOSSFSArgs := os.Getenv("DEFAULT_OSSFS_ARGS")
+	mountOptions = append(mountOptions, strings.Split(defaultOSSFSArgs, ",")...)
```
huww98, line 398: Maybe this should belongs to `func (f *fuseOssfs) AddDefaultMountOptions(options []string) []string`?

Split by comma can be problematic as we cannot include comma in the option itself. But maybe fine, since we already doing this in `parseOtherOpts`.

pkg/oss/nodeserver.go
```diff
@@ -392,6 +394,9 @@ func (o *Options) MakeMountOptionsAndAuthConfig(m metadata.MetadataProvider, vol
 	}
 	mountOptions = append(mountOptions, authOptions...)
 
+	defaultOSSFSArgs := os.Getenv("DEFAULT_OSSFS_ARGS")
```
huww98, line 397:
```suggestion
	defaultOSSFSOptions := os.Getenv("DEFAULT_OSSFS_OPTIONS")
```

PR #1363, opened 2025-04-14
correctly verify fsType

huww98:
#### What type of PR is this?

/kind bug

We should check for VolumeCapabilities, since that is what actually used in mount.

We currently check for "fsType" but not "fstype", while the latter is what in the doc. We check check for both (and every variants recognized by external provisioner).

Fixes #

```release-note
Check for validity of fstype (along with fsType) parameter.
```

```docs

```

### DIFF
--- a/pkg/disk/constants.go
+++ b/pkg/disk/constants.go
@@ -132,12 +132,10 @@ const (
 	PROVISIONED_IOPS_KEY = "provisionedIops"
 	BURSTING_ENABLED_KEY = "burstingEnabled"
 
-	CSI_DEFAULT_FS_TYPE = "csi.storage.k8s.io/fstype"
-	FS_TYPE             = "fsType"
-	EXT4_FSTYPE         = "ext4"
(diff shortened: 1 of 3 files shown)


mowangdk:
Will this be a breaking change for current users?

huww98:
> Will this be a breaking change for current users?

I think no. We currently only pack ext4 and xfs mkfs tools, so if user actually using other fstype, they will get event on Pod:
```
MountVolume.MountDevice failed for volume "d-xxxxxxxxxxxxxx" : rpc error: code = Aborted desc = executable file not found in $PATH
```

With this change, User will get error on PVC:
```
failed to provision volume with StorageClass "wrong-fstype": rpc error: code = InvalidArgument desc = Invalid parameters from input: disk-8cb9cb23-006f-449a-b76a-0c6c7767430f, with error: fsType notsupported is not supported, please use [ext3 xfs ntfs ext4]
```

So we just make the error appears earlier.

PR #1370, opened 2025-04-24
bump external snapshotter to v8.2.1

huww98:
#### What type of PR is this?

/kind cleanup

Update VGS related CRDs to v1beta1

Fixes #

```release-note
Update VGS related CRDs to v1beta1
```

```docs

```


huww98:
/cc @AlbeeSo 
Maybe we need to consider what will happen when upgrading existing cluster that already used v1alpha1

huww98:
/hold
conversion webhook

huww98:
We are not supporting v1beta1 version, and go to v1beta2 directly.

Also removing the conversion webhook, which converts between v1beta1 and v1beta2.

huww98:
/unhold

v1alpha1 cannot be migrated to v1beta2 (will crash sidecar due to unmarshal failure, time fields changed from int to string). But considering v1alpha1 is alpha, and nobody actually using this on ACK, I think it is OK to ignore this issue.

Not removing v1alpha1 because it would make upgrade hard. ready to merge now.

huww98:
port 9080 is reserved for POV controller internally. so use 10080 instead.

mowangdk:
@AlbeeSo ptal

/lgtm

PR #1386, opened 2025-04-30
disk: don't get VGS to check for an unsupported parameter

huww98:
#### What type of PR is this?

/kind cleanup

We should not support this in the first place. This will affect user cost, and should not be allowed to set on PVC.

Fixes #

```release-note
NONE
```

```docs

```

### DIFF
--- a/pkg/disk/group_volume_snapshot_utils.go
+++ b/pkg/disk/group_volume_snapshot_utils.go
@@ -1,7 +1,6 @@
 package disk
 
 import (
-	"context"
 	"fmt"
 	"slices"
 	"strings"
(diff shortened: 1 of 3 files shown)


mowangdk:
@AlbeeSo ptal

pkg/disk/group_volume_snapshot_utils.go
```diff
@@ -81,19 +48,7 @@ func parseGroupSnapshotParameters(params map[string]string, ecsParams *createGro
 			})
 		}
 	}
-	return nil
-}
-
-// if volumesnapshot have Annotations, use it first.
-// storage.alibabacloud.com/snapshot-ttl
-func parseGroupSnapshotAnnotations(anno map[string]string, ecsParams *createGroupSnapshotParams) error {
```
AlbeeSo, line 89: BackupCenter may rely on this annotation. As each backup has different  expected TTL. It's not considerable to create a new VGSclass everytime.

huww98: If we do support setting snapshot-ttl in the future, we should pass that in from external-snapshotter. Get VGS in CSI is not correct. User may have deleted the VGS when we get it. User may even have created another VGS with the same name and we may get wrong info.

For the TTL case. Can we just implement a controller to delete the VS/VGS from k8s, instead of rely on ECS to delete the underlying snapshot?

AlbeeSo: For backup case, VGS/VS will be deleted after it's done and its templated has uploaded to OSS. And to ensure the date security, they will all be set to Retain. And backupcenter controller do not and should not rely on the version of CSI. It's means the similar logic of VS cannot be removed or else TTL will not work for backup.

PR #1394, opened 2025-05-06
ossfs2: set default logdir to /dev/stdout

AlbeeSo:
#### What type of PR is this?
/kind bug

use /dev/stdout as default log_dir for ossfs 2.0.

Fixes #

```release-note

```

```docs

```

### DIFF
--- a/pkg/mounter/oss/ossfs.go
+++ b/pkg/mounter/oss/ossfs.go
@@ -303,10 +303,7 @@ func (f *fuseOssfs) getAuthOptions(o *Options, region string) (mountOptions []st
 func (f *fuseOssfs) AddDefaultMountOptions(options []string) []string {
 	defaultOSSFSOptions := os.Getenv("DEFAULT_OSSFS_OPTIONS")
 	if defaultOSSFSOptions != "" {
-		optList := strings.Split(defaultOSSFSOptions, ",")
-		for _, opt := range optList {
-			options = append(options, strings.TrimSpace(opt))
-		}
+		options = append(options, strings.Split(defaultOSSFSOptions, ",")...)
 	}
 
 	alreadySet := false
--- a/pkg/mounter/oss/ossfs.go
+++ b/pkg/mounter/oss/ossfs.go
@@ -301,6 +301,14 @@ func (f *fuseOssfs) getAuthOptions(o *Options, region string) (mountOptions []st
 }
 
 func (f *fuseOssfs) AddDefaultMountOptions(options []string) []string {
+	defaultOSSFSOptions := os.Getenv("DEFAULT_OSSFS_OPTIONS")
+	if defaultOSSFSOptions != "" {
+		optList := strings.Split(defaultOSSFSOptions, ",")
+		for _, opt := range optList {
+			options = append(options, strings.TrimSpace(opt))
+		}
+	}
+
 	alreadySet := false
 	for _, option := range options {
 		if strings.Contains(option, "dbglevel") {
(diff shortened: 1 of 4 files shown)


pkg/mounter/oss/ossfs.go
```diff
@@ -301,6 +301,14 @@ func (o *Options) getAuthOptions(region string) (mountOptions []string) {
 }
 
 func (f *fuseOssfs) AddDefaultMountOptions(options []string) []string {
+	defaultOSSFSOptions := os.Getenv("DEFAULT_OSSFS_OPTIONS")
+	if defaultOSSFSOptions != "" {
+		optList := strings.Split(defaultOSSFSOptions, ",")
+		for _, opt := range optList {
+			options = append(options, strings.TrimSpace(opt))
```
huww98, line 308: Are you 100% sure that space is not valid in an option? e.g., I think `passwd_file` can take a path ending with space? We should take extra care here because this is an API. It's behavior may not change theoretically.

pkg/mounter/oss/ossfs2.go
```diff
@@ -156,15 +156,35 @@ func (f *fuseOssfs2) buildPodSpec(c *utils.FusePodContext, target string) (spec
...
-			alreadySet = true
-			break
+		if option == "" {
+			continue
+		}
+		parts := strings.SplitN(option, "=", 2)
+		if len(parts) == 1 {
+			parts = append(parts, "")
 		}
+		tm[parts[0]] = parts[1]
```
huww98, line 183:
```suggestion
		k, v, _ := strings.Cut(option, "=")
		tm[k] = v
```
But I think we should have a unified place to do the parsing. Maybe `AddDefaultMountOptions` can just return all the defaults, and have another func `DeduplicateOptions` to parse the options and remove duplications? Mount options supports escaping with `\` actually. It is complicated.

pkg/mounter/oss/ossfs2.go
```diff
@@ -176,9 +196,9 @@ func (f *fuseOssfs2) AddDefaultMountOptions(options []string) []string {
 		}
 	}
 
-	defaultOSSFSOptions := os.Getenv("DEFAULT_OSSFS2_OPTIONS")
-	if defaultOSSFSOptions != "" {
-		options = append(options, strings.Split(defaultOSSFSOptions, ",")...)
+	// set default log dir
+	if _, ok := tm[KeyLogDir]; !ok {
+		options = append(options, fmt.Sprintf("log_dir=%s", "/dev/stdout"))
```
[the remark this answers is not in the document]
AlbeeSo: I'm sure it works.

pkg/mounter/oss/ossfs.go
AlbeeSo: I don't know any, but i's just for private cloud now.  
Maybe @iawia002 can consider here, I think it depend on if this ENV is decided to configured by users.
I will not trimspace here in this PR.

pkg/mounter/oss/ossfs2.go
AlbeeSo: I'm not sure if another func is needed to split the defaultMountOptions retrieving and duplications removing. It together works as `AddDefaultMountOptions ` and it's meaningful. Maybe we can consider it if more default options need to be handled.

mowangdk: We need to list the default options in comments and docs

mowangdk:
/lgtm
/approve

@AlbeeSo TODO: Add default oss parameters to docs

PR #1399, opened 2025-05-20
Add bmcpfs driver

tydra-wang:
#### What type of PR is this?

/kind feature

Fixes #

```release-note

```

```docs

```


pkg/mounter/proxy/server/alinas/driver.go
```diff
@@ -76,17 +76,25 @@ func runCommandForever(command string, args ...string) {
 
 // addAutoFallbackNFSMountOptions adds auto_fallback_nfs mount option when using efc
```
mowangdk, line 77: Please update comments, and add UT.

pkg/bmcpfs/internal/vsc.go
```diff
@@ -0,0 +1,458 @@
...
+	err error
+}
+
+type PrimaryVscManagerWithCache struct {
+	VscManager
+	retryTimes int
+
+	cond *sync.Cond
+	// Instance ID to VSC
+	cache map[string]vscWithErr
```
mowangdk, line 131: Why not just use concurrent map here? maybe we can optimize in the future

PR #1375, opened 2025-04-25
fix possible race in storageClassZonePos

huww98:
#### What type of PR is this?

/kind bug

#### Which issue(s) this PR fixes:

Fixes #

#### Does this PR introduce a user-facing change?

```release-note
NONE
```

#### Additional documentation e.g., KEPs (Kubernetes Enhancement Proposals), usage docs, etc.:

```docs

```

PR #1405, opened 2025-05-26
disk: refactor determine ZoneID

huww98:
#### What type of PR is this?

/kind bug
/kind cleanup

Pick the first from the intersection of parameter and AccessibilityRequirements.

Ensure error is returned when parameter and AccessibilityRequirements conflicts, or when unknown topology is specified.

fix possible race in storageClassZonePos, it is now removed. We will respect the order from AccessibilityRequirements, falling back to random.

Replaces #1375 

```release-note
We now returns error when zoneID in parameter and AccessibilityRequirements conflicts, or when unknown topology is specified.
```

```docs

```

### DIFF
--- a/build/csi-agent/Dockerfile
+++ b/build/csi-agent/Dockerfile
@@ -1,4 +1,4 @@
-FROM --platform=$BUILDPLATFORM registry-cn-hangzhou.ack.aliyuncs.com/dev/golang:1.22.9 as builder
+FROM --platform=$BUILDPLATFORM registry-cn-hangzhou.ack.aliyuncs.com/dev/golang:1.23.9 as builder
 WORKDIR /src
 ARG TARGETARCH
 ARG TARGETOS
--- a/pkg/disk/utils.go
+++ b/pkg/disk/utils.go
@@ -54,6 +54,7 @@ import (
 	v1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/util/sets"
+	"k8s.io/client-go/tools/record"
 	"k8s.io/klog/v2"
 	"k8s.io/mount-utils"
 	k8smount "k8s.io/mount-utils"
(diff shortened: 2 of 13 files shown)


build/csi-agent/Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM --platform=$BUILDPLATFORM registry-cn-hangzhou.ack.aliyuncs.com/dev/golang:1.22.9 as builder
+FROM --platform=$BUILDPLATFORM registry-cn-hangzhou.ack.aliyuncs.com/dev/golang:1.23.9 as builder
```
mowangdk, line 1: Why do we need to bump up golang version

huww98: Because I'm using [iter](https://pkg.go.dev/iter) from go 1.23 in `iterZone`

pkg/disk/utils.go
```diff
@@ -458,36 +459,64 @@ func parseTags(params map[string]string) (map[string]string, error) {
 	return seenTags, nil
 }
 
+var ErrUnknownZone = errors.New("unknown zone")
+
+func getZone(req *csi.CreateVolumeRequest) (string, error) {
+	volOptions := req.GetParameters()
+	zoneID, ok := volOptions[ZoneID]
```
mowangdk, line 476: I think we should add logic here, the ZoneID parameter should be invalid when pvc uses wffc, at least a warning event if we don't want to make a break change

huww98: Just tested, if scheduler selects zone A, but we really provision a disk in zone B, scheduler can still assign the Pod to a Node in zone B, and the pod can starts normally.

So yes, if we want to comply with CSI spec, there will be a breaking change here. For users who use WaitForFirstConsumer and zoneID simultaneously, they may get the error now:
```
failed to provision volume with StorageClass "with-zone": rpc error: code = InvalidArgument desc = Invalid parameters from input: disk-e7e63512-0430-4222-8e25-5a7614fb9b26, with error: conflicting zone, parameters specified: cn-beijing-i, accessibility requires: [cn-beijing-g]
```

It is OK if the scheduler select the zone that also appears in parameter.

huww98: Just discussed with @mowangdk , I will change this to just report warning event, maybe rejecting this in the future.

PR #1406, opened 2025-05-27
nas: make 'standard' the default NAS file system type

iawia002:
#### What type of PR is this?

/kind bug

In private cloud environments, NAS service domain names do not follow the `nas.aliyuncs.com` format. This leads to the file system type being incorrectly determined in these private clouds.

To resolve this for private cloud use cases, this patch makes 'standard' the default NAS file system type.

Fixes #

```release-note
none
```

```docs

```

/cc @AlbeeSo

### DIFF
--- a/pkg/nas/cloud/utils.go
+++ b/pkg/nas/cloud/utils.go
@@ -14,10 +14,9 @@ func GetFilesystemTypeByMountTargetDomain(domain string) string {
 		return FilesystemTypeExtreme
 	case strings.HasSuffix(domain, "cpfs.nas.aliyuncs.com"):
 		return FilesystemTypeCpfs
-	case strings.HasSuffix(domain, "nas.aliyuncs.com"):
-		return FilesystemTypeStandard
 	default:
-		return ""
+		// *.nas.aliyuncs.com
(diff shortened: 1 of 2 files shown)


tydra-wang (approved):
/lgtm

pkg/nas/cloud/utils.go
```diff
@@ -14,10 +14,8 @@ func GetFilesystemTypeByMountTargetDomain(domain string) string {
 		return FilesystemTypeExtreme
 	case strings.HasSuffix(domain, "cpfs.nas.aliyuncs.com"):
 		return FilesystemTypeCpfs
-	case strings.HasSuffix(domain, "nas.aliyuncs.com"):
-		return FilesystemTypeStandard
 	default:
-		return ""
+		return FilesystemTypeStandard
```
mowangdk, line 19: We need to add some comments here to explain why we chose the standard type, if not we also need to add some explainations

iawia002: Currently, there are only these three types of NAS. Returning an empty string serves no practical purpose and will cause issues with the description. I've added a comment to indicate the URL that the default case will match.

PR #1413, opened 2025-05-30
disk: try force attach if the same disk is already detaching

huww98:
#### What type of PR is this?

/kind feature

In some cases, we cannot detach a disk from ECS, maybe because of
malfunctioning guest OS, or communication issue between control and data plane.
When this happens, we can use force attach to escape to another instance, in
order to recover workload.  The old instance will see I/O error if detach
failed, and the detach will be retried by ECS.

Force attach is currently only possible for regional ESSD, when old and new
instances are in different zone. These restrictions may be lifted in the
future.

For this to work, we rely on KCM to initiate attaching after the first
detaching failed. i.e. Failed detaching will not block next attaching.

Fixes #

```release-note
Use force attach when we cannot detach the disk.
```

```docs

```

### DIFF
--- a/pkg/disk/category.go
+++ b/pkg/disk/category.go
@@ -45,6 +45,7 @@ type CategoryDesc struct {
 	SingleInstance          bool // Cannot be attached to another instance once it is attached.
 	Regional                bool
 	SnapshotConsistentGroup bool
+	ForceAttach             bool
 }
 
 var AllCategories = map[Category]CategoryDesc{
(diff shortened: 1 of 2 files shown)


huww98:
rebased and dropped the 2 already merged commits

pkg/disk/category.go
```diff
@@ -44,6 +44,7 @@ type CategoryDesc struct {
 	SingleInstance          bool // Cannot be attached to another instance once it is attached.
 	Regional                bool
 	SnapshotConsistentGroup bool
+	ForceAttach             bool
```
mowangdk, line 48: Is there any featuregate here?

huww98: Do we need one? Currently only Regional ESSD supports this. The impact should already be small enough. No breaking change is expected.

PR #1423, opened 2025-06-05
ossfs 1.0: add default option: listobjectsv2

AlbeeSo:
#### What type of PR is this?
/kind feature

https://help.aliyun.com/zh/oss/developer-reference/listobjects#section-t6d-ovx-jse
As the document of OSS, it's recommended to use listObjectsV2 by default.

Fixes #

```release-note

```

```docs

```

### DIFF
--- a/pkg/mounter/oss/ossfs.go
+++ b/pkg/mounter/oss/ossfs.go
@@ -343,6 +343,10 @@ func (f *fuseOssfs) AddDefaultMountOptions(options []string) []string {
 	}
 
 	// set listobjectsv2
+	// Note: OSS officially recommends using v2 API as the preferred version,
+	// but ossfs hasn't enabled listobjectv2 by default yet.
+	// This is a temporary workaround added by CSI driver.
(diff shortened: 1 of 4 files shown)


huww98:
> As the document of OSS, it's recommended to use listObjectsV2 by default.

Then why not build this into OSSFS itself? Will there be any breaking changes?

AlbeeSo:
> > As the document of OSS, it's recommended to use listObjectsV2 by default.
> 
> Then why not build this into OSSFS itself? Will there be any breaking changes?

Both are okay. No break change, some performance differences.

AlbeeSo:
But if the upstream ossfs do not change the default value, it's better add it in command line, it'll be clear.

huww98:
I think it is better to change this upstream, because this is not container specific.

mowangdk:
Changing the default API version in a CSI driver is really weird,  it's better to modify the upstream. I anticipate no similar future requirements.

mowangdk:
/lgtm
/approve

These codes need to be removed In the future and similar codes should not appear again

PR #1426, opened 2025-06-06
feat:support cloud essd with new disk type as xc0 and xc1

ThunderYe:
What type of PR is this?
/kind feature

What this PR does / why we need it:
In some financial cloud scenarios, we need support some special disk category as cloud_essd_xc0 and cloud_essd_xc1
Which issue(s) this PR fixes:
Fixes #

Special notes for your reviewer:
Does this PR introduce a user-facing change?

Additional documentation e.g., KEPs (Kubernetes Enhancement Proposals), usage docs, etc.:

### DIFF
--- a/pkg/disk/category.go
+++ b/pkg/disk/category.go
@@ -8,9 +8,12 @@ const (
 	DiskEfficiency Category = "cloud_efficiency"
 	DiskSSD        Category = "cloud_ssd"
 	DiskESSD       Category = "cloud_essd"
-	DiskESSDAuto   Category = "cloud_auto"
-	DiskESSDEntry  Category = "cloud_essd_entry"
-	DiskRegional   Category = "cloud_regional_disk_auto"
+	//special type as _xc* only used in financial cloud
+	DiskESSDXc0   Category = "cloud_essd_xc0"
+	DiskESSDXc1   Category = "cloud_essd_xc1"
+	DiskESSDAuto  Category = "cloud_auto"
+	DiskESSDEntry Category = "cloud_essd_entry"
+	DiskRegional  Category = "cloud_regional_disk_auto"
 
 	DiskPPerf            Category = "cloud_pperf"
 	DiskSPerf            Category = "cloud_sperf"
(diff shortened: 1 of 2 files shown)


pkg/disk/cloud_test.go
```diff
@@ -225,6 +225,44 @@ func TestBuildCreateDiskRequest(t *testing.T) {
...
+	})
+	assert.Equal(t, "cloud_essd_xc1", req2.DiskCategory)
+	// fields is copied
+	assert.Equal(t, "cn-hangzhou", req2.ZoneId)
+
+	// send req2 should not affect req
+	requests.InitParams(req2)
+	assert.Greater(t, len(req2.QueryParams), len(req.QueryParams))
+}
+
```
huww98: I think these are redundant. They don't increase any coverage.

pkg/disk/category.go
```diff
@@ -8,6 +8,8 @@ const (
 	DiskEfficiency Category = "cloud_efficiency"
 	DiskSSD        Category = "cloud_ssd"
 	DiskESSD       Category = "cloud_essd"
+	DiskESSDXc0    Category = "cloud_essd_xc0"
+	DiskESSDXc1    Category = "cloud_essd_xc1"
```
huww98: Please add some comment and/or git commit message stating these are used in financial cloud.

huww98 (changes requested):

pkg/disk/cloud_test.go
mowangdk: +1

ThunderYe: Hi @huww98 @mortent ,
     Sorry ,look alike , I hold some wrong assumption in "finalizeCreateDiskRequest"  as it should assume some pre-defination  "DiskESSDXc1" and "DiskESSDXc0"  in AllCategories;  
    But this funcion doesn't check there is a valid value when calling "cateDesc := AllCategories[attempt.Category]", maybe with a exist checking will be better as "cateDesc, ok := AllCategories[attempt.Category]"
   any way ,this test code is redundant, let's remove it.

pkg/disk/category.go
ThunderYe: yes, later I'll update this commit with more detail

ThunderYe:
Quite thanks guys, @mowangdk @huww98 and all!

PR #1429, opened 2025-06-11
Update plugin.yaml

engabduallah:
### Problem Description

Currently, when using the Helm chart with `nodePools` enabled, Helm template evaluation may fail in some environments. Specifically, when running:

```bash
helm install ... --values values.yaml
```

with this configuration present:

```yaml
nodePools:
  default:
    pluginname: csi-plugin
    nodeSelectorTerms:
      - matchExpressions:
          - key: kubernetes.io/os
            operator: In
            values:
              - linux
```

we encountered this error:

```
Error: template: alibaba-cloud-csi-driver/templates/plugin.yaml:2:36: executing "alibaba-cloud-csi-driver/templates/plugin.yaml" at <ne $nodePool nil>: error calling ne: uncomparable type map[string]interface {}: map[...]
```

This happens because Helm (Go templating engine) cannot safely compare maps to `nil`. Once `nodePools` is defined, Helm cannot process `ne $nodePool nil` safely.

---

### What we did

We modified the template logic in:

```bash
alibaba-cloud-csi-driver/deploy/chart/templates/plugin.yaml
```

**Before:**

```gotemplate
{{- if ne $nodePool nil }}
```

**After:**

```gotemplate
{{- if and $.Values.plugin.enabled (not (empty $nodePool)) }}
```

* The function `empty` safely evaluates both `nil` and empty maps.
* This change prevents Helm from failing even when `nodePools` is defined and non-empty.
* The behavior remains identical for valid nodePools configurations.

---

### Why we still use `nodePools`

* In our environment, we are using the chart with `nodePools` configuration because it provides us more flexibility to target nodes using `nodeSelectorTerms` instead of only `plugin.nodeSelector`.
* We did not remove `nodePools`. We simply fixed the template evaluation logic so it works safely in all environments, including private cloud and bare metal clusters.

---

### Result

* ✅ Helm chart now installs properly with `nodePools` enabled.
* ✅ DaemonSet is created correctly.
* ✅ CSI Plugin is running and storage provisioning works fine.
* ✅ The fix is backwards compatible and minimal.

---

### Summary

This small change makes Helm chart more robust and compatible with current Helm templating best practices while preserving existing nodePools logic for users who rely on it.

---

✅ **Tested environment:**

* Bare metal Kubernetes cluster
* Private cloud (non-ECS)
* Helm v3.13.0+

---

### DIFF
--- a/deploy/chart/templates/plugin.yaml
+++ b/deploy/chart/templates/plugin.yaml
@@ -1,5 +1,5 @@
 {{- range $key, $nodePool := .Values.nodePools -}}
-{{- if and $.Values.plugin.enabled (ne $nodePool nil) }}
+{{- if and $.Values.plugin.enabled (not (empty $nodePool)) }}
 ---
 {{- $values := deepCopy $.Values }}
 {{- $nodePool := mergeOverwrite $values $nodePool }}


mowangdk:
@engabduallah Please sign the CLA~

huww98:
Is this an issue specific to old version of Helm? We are actually testing nodepool config in CI (pull-alibaba-cloud-csi-driver-verify-helm), but with higher Helm version:

https://github.com/kubernetes-sigs/alibaba-cloud-csi-driver/blob/37bf5545a6de75746d3665b86e6448c136f85e5a/Makefile#L15

huww98:
I think it is fine. The use case is that we should support disabling the default node pool with:
```yaml
nodePools:
  default: nil
```

engabduallah:
Hi Team, 
i am using Helm version 3.17.3

helm version
version.BuildInfo{Version:"v3.17.3", GitCommit:"e4da49785aa6e6ee2b86efd5dd9e43400318262b", GitTreeState:"clean", GoVersion:"go1.23.7"}

when i set it as following:
nodePools:
  default: nil
  
the deamonset of the plugin is not deployed and volume can not be created so without this fix, i could not procced. on my local i fixed it and then used the following so it can deploy on all linux worker:
nodePools:
  default:
    pluginname: csi-plugin
    nodeSelectorTerms:
      - matchExpressions:
          - key: kubernetes.io/os
            operator: In
            values:
              - linux

mowangdk:
I couldn't reproduce the issue using following helm version and specific nodepool configuration. 
 
version.BuildInfo{Version:"v3.18.3", GitCommit:"6838ebcf265a3842d1433956e8a622e3290cf324", GitTreeState:"clean", GoVersion:"go1.24.4"}

engabduallah:
Hi @mowangdk 

thank you for your comment. 

Can you please share the value.yaml that you used  while trying to reporduce it ?

mowangdk:
```
controller:
  enabled: true
  replicas: 2

plugin:
  enabled: true

logging:
  format: text  # json is recommended if sls is enabled
  sls:
    enabled: false
    project:
      name: null  # default to k8s-log-<CLUSTER_ID>
    logstore:
      name: alibabacloud-csi-plugin

csi:
  disk:
    enabled: true
    controller:
      enabled: true
  nas:
    enabled: true
    controller:
      enabled: true
  oss:
    enabled: true
    controller:
      enabled: true
  local:
    enabled: false
    controller:
      enabled: true

nodePools:
  default:
    pluginname: csi-plugin
    nodeSelectorTerms:
      - matchExpressions:
          - key: kubernetes.io/os
            operator: In
            values:
              - linux

defaultStorageClass:
  enabled: true

volumeSnapshot:
  # These are shared by all CSI drivers that support snapshots.
  # Disable if you manage them elsewhere.
  controller:
    enabled: true

deploy:
  kubeletRootDir: /var/lib/kubelet

  # By default, the controller server does not delete NAS subpaths when DeleteVolume, that means
  # the files created in the volume will remain in the filesystem after PV deleted.
  # You can set "withStorageController" to true in ACK clusters with storage-operator addon installed, 
  # then the controller server will patch a finalizer on NAS subpath-type PV when DeleteVolume 
  # and the storage-controller will handle the cleanup of the subpath in NAS filesystem.
  withStorageController: false

  ramToken: v2
  # Which namespace holds ossfs pods
  fuseNamespace: ack-csi-fuse

  featureGates: ""   # only for alpha feature
  privateCloud: false
  ecsEndpoint: null  # most useful on private cloud
  stsEndpoint: null  # most useful on private cloud
  nasEndpoint: null  # most useful on private cloud

  ack: true  # deployed on managed Alibaba Cloud Container Service for Kubernetes (ACK)
  ecs: true  # use Alibaba Cloud Elastic Compute Service (ECS) for Node

  # Possible values:
  #   - public: has access to Internet
  #   - vpc: has access to Alibaba Cloud Virtual Private Cloud (VPC)
  #   - null: use vpc for ECS, SDK defaults otherwise
  network: null
  clusterID: null  # will be injected on installation on ACK
  regionID: null  # useful when deploying on non-ECS, but need to access Alibaba Cloud OpenAPI

  # To access Alibaba Cloud OpenAPI in self-deployed Kubernetes cluster, use a Secret with "id" and "secret" set.
  accessKey:
    enabled: false
    secretName: csi-access-key
    idKey: id # the key of AccessKey id in the k8s Secret
    secretkey: secret # the key of AccessKey secret in the k8s Secret

images:
  registry: registry-cn-hangzhou.ack.aliyuncs.com
  registryVPC: registry-cn-hangzhou-vpc.ack.aliyuncs.com
  controller:
    repo: acs/csi-plugin
    tag: "v1.31.4-9819c8b-aliyun"
  plugin:
    repo: acs/csi-plugin
    tag: "v1.31.4-9819c8b-aliyun"
  pluginInit:
    repo: acs/csi-plugin
    tag: "v1.31.4-9819c8b-aliyun-init"
  localController:
    repo: acs/local-csi-driver
    tag: "v0.2.0"
  localPlugin:
    repo: acs/local-csi-driver
    tag: "v0.2.0"
  externalAttacher:
    repo: acs/csi-attacher
    tag: "v4.6.1-aliyun"
  externalProvisioner:
    repo: acs/csi-provisioner
    tag: "v5.0.3-aliyun"
  externalResizer:
    repo: acs/csi-resizer
    tag: "v1.11.2-aliyun"
  externalSnapshotter:
    repo: acs/csi-snapshotter
    tag: "v8.1.1-aliyun"
  externalSnapshotController:
    repo: acs/snapshot-controller
    tag: "v8.1.1-aliyun"
  pluginRegistrar:
    repo: acs/csi-node-driver-registrar
    tag: "v2.9.0-d48d2e0-aliyun"
  ossfs:
    repo: acs/csi-ossfs
    tag: "v1.91.5.ack.1-ed398f6-aliyun"
```
Modified only the noodpool section in values.yaml.

PR #940, opened 2024-01-08
Refactor get device by volumeID routines

huww98:
#### What type of PR is this?

/kind cleanup

#### What this PR does / why we need it:

* Extract logic to gain testability
* Try device link first
* Support NVMe device link
* Return the disk/by-id path if found
* Support for NVMe device in getDeviceBySerial
* Remove search disk by udevadm
* Merge in the logic of GetRootSubDevicePath into adaptDevicePartition
* Add unit test

See commit message for details

#### Which issue(s) this PR fixes:

Fixes the confusing error message when AD controller enabled but disk not found:

Old: NodeStageVolume: ADController Enabled, but device can't be found:d-2ze08oskojgoeh7vceblvolumeID link path "/dev/disk/by-id/virtio-2ze08oskojgoeh7vcebl" not found

New: NodeStageVolume: ADController Enabled, but disk d-2ze08oskojgoeh7vceblvolume can't be found: [get by link "/dev/disk/by-id/virtio-2ze08oskojgoeh7vcebl" failed: file does not exist, get by link "/dev/disk/by-id/nvme-Alibaba_Cloud_Elastic_Block_Storage_2ze08oskojgoeh7vcebl" failed: file does not exist, find by serial: device not found]

#### Does this PR introduce a user-facing change?

```release-note
NONE
```

#### Additional documentation e.g., KEPs (Kubernetes Enhancement Proposals), usage docs, etc.:

```docs

```

PR #1430, opened 2025-06-16
disk: partition must be block device

huww98:
New anolis kernel introduce a new char device named /dev/vdXc0. We need to exclude it from the list of partition.

#### What type of PR is this?

/kind bug

Fixes #

In master, this is fixed by #940 

```release-note
NONE
```

```docs

```


pkg/disk/utils.go
```diff
@@ -387,7 +387,13 @@ func adaptDevicePartition(devicePath string) ([]string, error) {
 	for _, tmpDevice := range globDevices {
 		// find all device partitions
 		if result, err := regexp.MatchString(digitPattern, strings.TrimPrefix(tmpDevice, rootDevicePath)); err == nil && result == true {
-			deviceList = append(deviceList, tmpDevice)
+			info, err := os.Stat(tmpDevice)
```
mowangdk, line 390: why don't we mock stat here?

huww98: Because we currently don't have UT that cover this. There are too many things needs mock, basically everything about IO.

PR #1439, opened 2025-07-01
Add volumeid validation

mowangdk:
#### What type of PR is this?
/kind bug

The FormatLabelSelector function errors out with "\<error\>" when an invalid value is input, without specifying the invalid part.

Fixes #

```release-note

```

```docs

```

### DIFF
--- a/pkg/mounter/utils/fuse_pod_manager.go
+++ b/pkg/mounter/utils/fuse_pod_manager.go
@@ -14,6 +14,7 @@ import (
 	"k8s.io/apimachinery/pkg/fields"
 	"k8s.io/apimachinery/pkg/runtime"
 	utilerrors "k8s.io/apimachinery/pkg/util/errors"
+	"k8s.io/apimachinery/pkg/util/validation"
 	"k8s.io/apimachinery/pkg/util/wait"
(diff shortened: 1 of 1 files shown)


pkg/mounter/utils/fuse_pod_manager.go
```diff
@@ -212,7 +213,10 @@ func (fpm *FusePodManager) Create(c *FusePodContext, target string, atomic bool)
 		"volumeId", c.VolumeId,
 		"nodeName", c.NodeName,
 	)
-
+	errs := validation.IsValidLabelValue(c.VolumeId)
+	if errs != nil {
+		return nil, fmt.Errorf("invalid volumeId: %s, %v", c.VolumeId, errs)
```
huww98, line 218:
```suggestion
		return nil, fmt.Errorf("invalid volumeId: %s, %w", c.VolumeId, errs)
```

edit: never mind. It returns []string, not error.

huww98 (commented):
Should we just replace `FormatLabelSelector` with `LabelSelectorAsSelector`? So that we can return error earlier.

tydra-wang (approved):
/lgtm

PR #1440, opened 2025-07-02
oss: if the volumeId is too long for label value, use sha1

AlbeeSo:
#### What type of PR is this?
/kind bug

Allow using PV name / volumeHandle longer than 63 chars.

Fixes #

```release-note

```
New label is like 
```
csi-fuse-ossfs-nhcln   1/1     Running   0          81s   alibabacloud.com/drain-pod=skip,csi.alibabacloud.com/fuse-type=ossfs,csi.alibabacloud.com/mount-path-hash=849f9d6c5,csi.alibabacloud.com/volume-id=6836cf13bac400e9105071cd6af47084dfacad4e5e302c94bfed24e013afb73
```

```docs

```

### DIFF
--- a/pkg/mounter/utils/helper.go
+++ b/pkg/mounter/utils/helper.go
@@ -31,6 +31,15 @@ func computeMountPathHash(target string) string {
 	return rand.SafeEncodeString(fmt.Sprint(hasher.Sum32()))
 }
 
+func computeVolumeIdLabelVal(volumeId string) string {
+	errs := validation.IsValidLabelValue(volumeId)
+	if errs == nil {
+		return volumeId
+	}
+	volSha := sha256.Sum256([]byte(volumeId))
+	return hex.EncodeToString(volSha[:])[:63]
+}
+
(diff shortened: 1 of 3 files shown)


pkg/mounter/utils/helper.go
```diff
@@ -31,6 +31,15 @@ func computeMountPathHash(target string) string {
 	return rand.SafeEncodeString(fmt.Sprint(hasher.Sum32()))
 }
 
+func computeVolumeIdLabelVal(volumeId string) string {
+	errs := validation.IsValidLabelValue(volumeId)
+	if errs == nil {
+		return volumeId
+	}
+	volSha := sha256.Sum256([]byte(volumeId))
+	return hex.EncodeToString(volSha[:])[:63]
```
huww98: Since we are not caring about collision attack. Can we just use sha1, which is 40 hex chars, to avoid the somewhat confusing `[:63]`? And I also advise adding a `h1.` prefix, to clearly indicate it is a hash.
```suggestion
	volSha := sha1.Sum([]byte(volumeId))
	return "h1." + hex.EncodeToString(volSha[:])
```

mowangdk: lgtm

mowangdk:
@tydra-wang ptal

PR #1444, opened 2025-07-08
fix: use json content-type for snapshotclient, fixing autosnapshot issues

AlbeeSo:
#### What type of PR is this?
/kind bug

SnapClient does not support Protobuf ContentType, change it back to JSON 

Fixes #

```release-note

```

```docs

```


huww98 (commented):
OK. We also have several `dynamic.NewForConfig` usage. They are currently only used for read, so they should be fine now. But also fix them as a precaution?

PR #1453, opened 2025-07-15
feat(compatibility): support mounting bmcpfs vsc mount target using nas driver

tydra-wang:
#### What type of PR is this?
/kind feature

Fixes #

```release-note

```

```docs

```

### DIFF
--- a/pkg/nas/utils.go
+++ b/pkg/nas/utils.go
@@ -108,7 +108,13 @@ func doMount(mounter mountutils.Interface, opt *Options, targetPath, volumeId, p
 			mountFstype = "alinas"
 			// must enable tls when using accesspoint
 			combinedOptions = addTLSMountOptions(combinedOptions)
-		} else {
+		}
+		// Enable compatibility for BMCPFS VPC mount points using the NAS Driver, to support the same usage in ECI.
+		if strings.HasSuffix(opt.Server, "cpfs.aliyuncs.com") && opt.MountProtocol == "efc" {
+			mountFstype = "alinas"
+			combinedOptions = append(combinedOptions, "efc,protocol=efc,net=tcp,fstype=cpfs")
+		}
+		if mountFstype == "" {
 			mountFstype = opt.MountProtocol
 		}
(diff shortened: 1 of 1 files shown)


pkg/nas/utils.go
```diff
@@ -108,7 +108,13 @@ func doMount(mounter mountutils.Interface, opt *Options, targetPath, volumeId, p
 			mountFstype = "alinas"
 			// must enable tls when using accesspoint
 			combinedOptions = addTLSMountOptions(combinedOptions)
-		} else {
+		}
+		// Enable compatibility for BMCPFS VPC mount points using the NAS Driver, to support the same usage in ECI.
+		if strings.HasSuffix(opt.Server, "cpfs.aliyuncs.com") && opt.MountProtocol == "efc" {
+			mountFstype = "alinas"
+			combinedOptions = []string{"efc,protocol=efc,net=tcp,fstype=cpfs"}
```
huww98: Is it intentional to replace all the options? All other cases are `append`.

pkg/nas/utils.go
```diff
@@ -108,7 +108,13 @@ func doMount(mounter mountutils.Interface, opt *Options, targetPath, volumeId, p
 			mountFstype = "alinas"
 			// must enable tls when using accesspoint
 			combinedOptions = addTLSMountOptions(combinedOptions)
-		} else {
+		}
+		// Enable compatibility for BMCPFS VPC mount points using the NAS Driver, to support the same usage in ECI.
+		if strings.HasSuffix(opt.Server, "cpfs.aliyuncs.com") && opt.MountProtocol == "efc" {
```
mowangdk, line 113: Are there alternative methods to avoid modifying mount options based on domain suffixes, especially for custom domains?

tydra-wang: To support the custom domain, we need another fields to indicate the filesystem type. Maybe we can add a `filesystemType` field in the future, which can also fix other similar problems for nas.

mowangdk: ok

PR #1457, opened 2025-07-21
adding imagepullsecrets support in helm chart

waheedshahani:
#### What type of PR is this?
/kind feature

Fixes https://github.com/kubernetes-sigs/alibaba-cloud-csi-driver/issues/1455 

Fixes #

```release-note
NONE
```

Allow chart users to specify the imagepullsecrets used to authenticate to private registry. 

```docs

```


huww98:
How about using kustomize to further edit the yaml rendered by Helm? There are so many fields in the Pod, we definitely don't want to support customization each of them in helm.

I've not used that but PTAL at `kubectl kustomize --enable-helm`

waheedshahani:
> How about using kustomize to further edit the yaml rendered by Helm? There are so many fields in the Pod, we definitely don't want to support customization each of them in helm.
> 
> I've not used that but PTAL at `kubectl kustomize --enable-helm`

Sure it is a good idea for advanced customization. Since this is common parameter and is supported by CSI drivers from other cloud providers, it makes sense to have similar support here so that organizations can use similar approach across hyperscaler.

mowangdk:
@waheedshahani Please address the conflicts

PR #1500, opened 2025-08-28
disk: use our own topology keys

huww98:
We've got some report that some non-ACK cluster uses topology.kubernetes.io/region node label for their own purpose. So we should keep using our own topology key.

#### What type of PR is this?

/kind feature

Fixes #

```release-note
We use topology.diskplugin.csi.alibabacloud.com/region instead of topology.kubernetes.io/region for our topology key to avoid any conflicts.
```

```docs

```


huww98:
It turns out that if we don't report topology.kubernetes.io/zone topology key, user cannot use it in storage class allowedTopologies. So this is a breaking change (discovered by external-storage e2e test).

We may proceed to add a config to disable these keys.

PR #1508, opened 2025-09-10
support virtio for csi3.0

mowangdk:
#### What type of PR is this?
/kind feature

Support virtio protocol for csi3.0

Fixes #

```release-note

```

```docs

```

### DIFF
--- a/pkg/disk/nodeserver.go
+++ b/pkg/disk/nodeserver.go
@@ -1357,7 +1357,7 @@ func (ns *nodeServer) mountRunDVolumes(volumeId, pvName, sourcePath, targetPath,
 			}
 			err = unix.Setxattr(deviceName, DiskXattrVirtioBlkName, []byte("1"), 0)
 			if err != nil {
-				klog.Errorf("NodePublishVolume(rund3.0): Setxattr device: %s, err: %v",deviceName, err)
+				klog.Errorf("NodePublishVolume(rund3.0): Setxattr device: %s, err: %v", deviceName, err)
 				return true, err
 			}
 			return true, nil
--- a/pkg/bmcpfs/nodeserver.go
+++ b/pkg/bmcpfs/nodeserver.go
@@ -120,7 +120,7 @@ func (ns *nodeServer) NodePublishVolume(ctx context.Context, req *csi.NodePublis
 	if path := req.VolumeContext[_path]; path != "" {
 		source = fmt.Sprintf("%s:%s", source, path)
 	}
-	klog.InfoS("Mounting vpc mount target", "targetPath", req.TargetPath, "source", source)
+	klog.InfoS("Mounting mount target", "targetPath", req.TargetPath, "source", source)
 
 	mountOptions = append(mountOptions, "efc,protocol=efc,fstype=cpfs")
 	err = ns.mounter.Mount(source, req.TargetPath, "alinas", mountOptions)


pkg/disk/nodeserver.go
```diff
@@ -1420,6 +1454,9 @@ func (ns *nodeServer) checkMountedOfRunvAndRund(volumeId, targetPath string) boo
 		klog.ErrorS(err, "NodeStageVolume:  Failed to get current driver", "volumeId", volumeId)
 		return false
 	}
+	if ns.pvmType && defaultDrivers.Has(cDriver) {
```
huww98: This means if we run runC container on PVM machine, it will return true, is that OK? Can we just check for the file existence like the runv code path?

mowangdk: The file lies only in csi3.0 under the publish path, as stagevolume can't access to the file.

huww98: OK, so for PVM machine, we will always skip NodeStageVolume. and if it turns out to be a runC Pod, we rely on NodePublishVolume to detect and correct this (the `sourceNotMounted` branch), right?

pkg/utils/rund/directvolume/direct_volume.go
```diff
@@ -22,6 +25,12 @@ const (
 	BlockFileModeReadWrite = "432"
 	// BlockFileModeReadOnly decimal value of 440 (-r--r----)
 	BlockFileModeReadOnly = "288"
+
+	// DiskXattrPVMName xattr is applied on the block device file to indicate that it is managed by the CSI driver in PVM ways.
+	// Value is 1.
+	// Linux only support trusted namespace xattr in tmpfs before kernel v6.6,
+	// but setting trusted xaattr requires CAP_SYS_ADMIN capability, we may use user namespace instead in unit tests.
+	DiskXattrPVMName = "trusted.pvm"
```
huww98: How about `trusted.virtio-blk`? Even if PVM supports vfio in the future, we may still need this for other virtio-blk use-cases.

pkg/utils/rund/directvolume/direct_volume_test.go
```diff
@@ -0,0 +1,156 @@
...
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+	"golang.org/x/sys/unix"
+)
+
+func TestRemovePVMXattr(t *testing.T) {
+	// Create a temporary directory for testing
+	tempDir, err := os.MkdirTemp("", "directvolume_test")
```
huww98: please use `t.TempDir()`, which is automatically cleaned up.

Should we have at least one success case here?

pkg/utils/rund/directvolume/direct_volume_test.go
```diff
@@ -0,0 +1,156 @@
...
+	t.Run("device exists with pvm xattr set to 1", func(t *testing.T) {
+		deviceFile := filepath.Join(tempDir, "testdevice_pvm")
+		err := os.WriteFile(deviceFile, []byte("test"), 0644)
+		require.NoError(t, err)
+
+		// Set the xattr to "1"
+		err = unix.Setxattr(deviceFile, DiskXattrPVMName, []byte("1"), 0)
+		if err != nil {
+			// Skip test if xattrs are not supported in this environment
+			t.Skipf("xattrs not supported in this environment: %v", err)
```
huww98: We can set `DiskXattrPVMName` to other value in the test. See `TestGetVolumeCountFromOpenAPI()` for example.

pkg/utils/rund/directvolume/direct_volume_test.go
```diff
@@ -0,0 +1,156 @@
...
+		deviceFile := filepath.Join(tempDir, "testdevice_error")
+		err := os.WriteFile(deviceFile, []byte("test"), 0644)
+		require.NoError(t, err)
+
+		// Test with a special case where we simulate an error
+		// We can't easily simulate unix.Getxattr returning an error that is not ENODATA
+		// but is not nil either, so we'll test that the function properly handles
+		// the case when IsXattrNotFound returns false but err is not nil
+		// This is more of a logic test than a functional test
+		result := CheckDevicePVMMounted(deviceFile)
```
huww98: Isn't this identical to `device exists but has no xattr`? You can set the xattr to more than 8 bytes to get an error other than NotFound.

pkg/utils/rund/directvolume/direct_volume.go
```diff
@@ -75,3 +84,29 @@ func VolumeMountInfo(volumePath string) (*MountInfo, error) {
 	}
 	return &mountInfo, nil
 }
+
+func RemovePVMXattr(volumePath string) error {
+	mountInfo, err := VolumeMountInfo(volumePath)
+	if err != nil {
+		return err
+	}
+	device, err := filepath.EvalSymlinks(mountInfo.Source)
```
huww98: I think we don't need this one, Removexattr should follow symlinks by default.

pkg/disk/nodeserver.go
```diff
@@ -1323,6 +1340,26 @@ func (ns *nodeServer) mountRunDVolumes(volumeId, pvName, sourcePath, targetPath,
...
 
+		if pvmMode {
+			mountInfo.DeviceType = directvolume.DeviceTypeVirtioBlk
+			mountInfo.Source = deviceName
+			err = directvolume.AddMountInfo(directvolume.EnsureVolumeAttributesFileDir(targetPath, isRawBlock), mountInfo)
+			if err != nil {
+				klog.Errorf("NodePublishVolume(rund3.0): Adding runD mount information to DirectVolume within pvm failed: %v", err)
+				return true, err
+			}
+			actualDevice, err := filepath.EvalSymlinks(deviceName)
```
huww98: Don't need this as well.

pkg/utils/rund/directvolume/direct_volume.go
mowangdk: MountInfo.Source may be bdf,  for consistency we use EvalSymlinks to check

pkg/disk/nodeserver.go
mowangdk: Same as above

pkg/utils/rund/directvolume/direct_volume_test.go
mowangdk: I think test coverage has already been 100%
image

pkg/utils/rund/directvolume/direct_volume.go
huww98: That makes no sense to me. Removexattr can also return "file not found" error like EvalSymlinks

pkg/utils/rund/directvolume/direct_volume_test.go
huww98: `go test` only counts line coverage, not branch coverage.

And we write test to ensure the code does work as intended, so success case is important IMO.

pkg/utils/rund/directvolume/direct_volume.go
mowangdk: Removexattr is an irrevocable and potentially hazardous operation IMO

pkg/utils/rund/directvolume/direct_volume.go
mowangdk: virtio-blk may found in localdisk, but not desired in cloud storage

pkg/utils/rund/directvolume/direct_volume.go
huww98: Even if you really want to check for file existence, you should use stat().  EvalSymlinks() is complex, has high overhead, and breaks the operation atomicity.

mowangdk: Would passing the actual device to Removexattr simplify the function than symlink?

pkg/disk/bdf.go
```diff
@@ -35,9 +35,8 @@ const (
 	// BlkIDSz value
 	BlkIDSz = 20
 
-	sysPrefix        = "/host"
-	iohubSriovAction = sysPrefix + "/sys/bus/pci/drivers/iohub_sriov/"
-	virtioPciAction  = sysPrefix + "/sys/bus/pci/drivers/virtio-pci/"
+	iohubSriovAction = "/sys/bus/pci/drivers/iohub_sriov/"
+	virtioPciAction  = "/sys/bus/pci/drivers/virtio-pci/"
```
huww98, line 40: Just add a note for future viewer: /sys is automatically mounted read-write for privileged pod. So we don't need to manually mount it under /host/sys

pkg/utils/rund/directvolume/direct_volume.go
huww98: Kernel can resolve the symlink much more efficiently, while EvalSymlinks requires multiple round-trip from user-space to kernel space.

Did a benchmark to see the difference
```golang
const (
	device    = "/dev/disk/by-id/virtio-xxxxxx"
	xattrName = "trusted.virt-io"
)

func BenchmarkXAttr(b *testing.B) {
	for b.Loop() {
		require.NoError(b, unix.Setxattr(device, xattrName, []byte("1"), 0))
		require.NoError(b, unix.Removexattr(device, xattrName))
	}
}

func BenchmarkEvalXAttr(b *testing.B) {
	for b.Loop() {
		d, err := filepath.EvalSymlinks(device)
		require.NoError(b, err)
		require.NoError(b, unix.Setxattr(d, xattrName, []byte("1"), 0))
		d, err = filepath.EvalSymlinks(device)
		require.NoError(b, err)
		require.NoError(b, unix.Removexattr(d, xattrName))
	}
}
```
The result is:
```
goos: linux
goarch: amd64
pkg: github.com/kubernetes-sigs/alibaba-cloud-csi-driver/pkg/utils/rund/directvolume
cpu: Intel(R) Xeon(R) Platinum
BenchmarkXAttr-4          380840              3200 ns/op             128 B/op          4 allocs/op
BenchmarkEvalXAttr-4       76596             15169 ns/op            3008 B/op         48 allocs/op
PASS
```
As you can see, add `EvalSymlinks` makes it 5x slower.

From strace, we can see that EvalSymlinks is implement by 6 syscalls. Rather complex.
```
newfstatat(AT_FDCWD, "/dev", {st_mode=S_IFDIR|0755, st_size=3120, ...}, AT_SYMLINK_NOFOLLOW) = 0
newfstatat(AT_FDCWD, "/dev/disk", {st_mode=S_IFDIR|0755, st_size=140, ...}, AT_SYMLINK_NOFOLLOW) = 0
newfstatat(AT_FDCWD, "/dev/disk/by-id", {st_mode=S_IFDIR|0755, st_size=140, ...}, AT_SYMLINK_NOFOLLOW) = 0
newfstatat(AT_FDCWD, "/dev/disk/by-id/virtio-xxxx", {st_mode=S_IFLNK|0777, st_size=9, ...}, AT_SYMLINK_NOFOLLOW) = 0
readlinkat(AT_FDCWD, "/dev/disk/by-id/virtio-xxxx", "../../vda", 128) = 9
newfstatat(AT_FDCWD, "/dev/vda", {st_mode=S_IFBLK|0660, st_rdev=makedev(0xfd, 0), ...}, AT_SYMLINK_NOFOLLOW) = 0
```

And separate this to 2 step has the extra risk of operating on wrong device if something strange happened between these two step.

PR #1520, opened 2025-09-28
disk: fix Detaching disks treated as attached

huww98:
#### What type of PR is this?

/kind bug

We should reject to attach disk detaching from the requested node.

Fixes #

It is strange that ECS does not return error on attaching a detaching disk. But we fix it at our side first.

/hold
based on #1460, wait for that one to be merged first

```release-note
NONE
```

```docs

```


huww98:
Dropped the 2 commits already merged at #1649

PR #1525, opened 2025-10-11
ossfs: allow change ossfs pod' dnspolicy to ClusterFirstWithHostNet 

AlbeeSo:
#### What type of PR is this?
/kind bug /kind feature

Use ClusterFirstWithHostNet dnsPolicy by default for hostnetwork ossfs pod.

Fixes #

```release-note

```

```docs

```

### DIFF
--- a/pkg/mounter/oss/oss_fuse_manager.go
+++ b/pkg/mounter/oss/oss_fuse_manager.go
@@ -3,6 +3,7 @@ package oss
 import (
 	"github.com/kubernetes-sigs/alibaba-cloud-csi-driver/pkg/cloud/metadata"
 	"github.com/kubernetes-sigs/alibaba-cloud-csi-driver/pkg/mounter/utils"
+	corev1 "k8s.io/api/core/v1"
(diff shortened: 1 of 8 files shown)


pkg/mounter/oss/oss_fuse_manager.go
```diff
@@ -86,4 +87,7 @@ type Options struct {
 	AuthType      string `json:"authType"`
 	FuseType      string `json:"fuseType"`
 	ReadOnly      bool   `json:"readOnly"`
+
+	// pod template
+	DnsPolicy corev1.DNSPolicy `json:"dnsPolicy"`
```
mowangdk, line 92: I suggest exposing more abstract configuration options to users, as they may not be aware of our network policy being hostNetwork. A focus on coreDNS and nodeDNS settings would suffice.

AlbeeSo: But it means `ClusterFirstWithHostNet`, not sure to use coreDNS. In fact we just strictly reuse the definition of K8s, not defining a new action.

PR #1527, opened 2025-10-16
Fix incorrect nas mounter issue

iltyty:
#### What type of PR is this?
/kind bug

Fixes incorrect nas mounter selection: still use NewConnectorMounter when CSI is not running in agent mode.

Fixes incorrect nas mounter selection: still use NewConnectorMounter when CSI is not running in agent mode.

```release-note

```

```docs

```

### DIFF
--- a/pkg/nas/mounter.go
+++ b/pkg/nas/mounter.go
@@ -32,12 +32,16 @@ func (m *NasMounter) Mount(source string, target string, fstype string, options 
 	return err
 }
 
-func newNasMounter() mountutils.Interface {
+func newNasMounter(agentMode bool) mountutils.Interface {
 	inner := mountutils.NewWithoutSystemd("")
-	return &NasMounter{
+	m := &NasMounter{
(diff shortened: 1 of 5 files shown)


pkg/nas/mounter.go
```diff
@@ -32,12 +32,17 @@ func (m *NasMounter) Mount(source string, target string, fstype string, options
...
-func newNasMounter() mountutils.Interface {
+func newNasMounter(agentMode bool) mountutils.Interface {
 	inner := mountutils.NewWithoutSystemd("")
-	return &NasMounter{
-		Interface:     inner,
-		alinasMounter: inner,
+	m := &NasMounter{
+		Interface: inner,
+	}
+	if agentMode {
```
mowangdk: We can set ```alinasMounter``` as inner by default

mowangdk: Optimally, set AlinasMounter during initialization and make it unmodifiable.

PR #1528, opened 2025-10-17
[WIP]Support fileset of bmcpfs

mowangdk:
#### What type of PR is this?
/kind feature

Create fileset upon PV creation, delete it upon PVC deletion.
A temporary method for deleting bmcpfs filesets involves using the fileset ID within the volumeId(modified in external-provisioner). Once a solution is found, we should adopt a new approach.

Fixes #

```release-note

```

```docs

```

### DIFF
--- a/pkg/bmcpfs/controllerserver.go
+++ b/pkg/bmcpfs/controllerserver.go
@@ -23,6 +23,7 @@ import (
 	"path/filepath"
 	"strconv"
 	"strings"
+	"time"
 
 	openapi "github.com/alibabacloud-go/darabonba-openapi/v2/client"
 	efloclient "github.com/alibabacloud-go/eflo-controller-20221215/v3/client"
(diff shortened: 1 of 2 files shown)


pkg/bmcpfs/controllerserver.go
```diff
@@ -61,18 +113,153 @@ func newControllerServer(region string) (*controllerServer, error) {
...
+		klog.Errorf("CreateVolume: error parameters from input: %v, with error: %v", req.Name, err)
+		return nil, status.Errorf(codes.InvalidArgument, "Invalid parameters from input: %v, with error: %v", req.Name, err)
+	}
+
+	// Extract parameters
+	params := req.GetParameters()
+	bmcpfsID := params["bmcpfsId"]
+	fullPath := req.Name
+	if rootPath, ok := params["path"]; ok && rootPath != "" {
+		fullPath = filepath.Join(rootPath, req.Name)
```
AlbeeSo, line 157: Is it possible to update the path field with the final fullpath?  BMCPFS VolumePopulator needs to extract the target Path from the PV.

AlbeeSo: So do the FsetId.

PR #1529, opened 2025-10-20
NAS: validate server in NodePublishVolume

iltyty:
#### What type of PR is this?
/kind cleanup

Prevent ambiguous mount errors by validating NAS/CPFS server format (no colons allowed).

Fixes #

```release-note

```

```docs

```

### DIFF
--- a/pkg/nas/nodeserver.go
+++ b/pkg/nas/nodeserver.go
@@ -314,6 +314,10 @@ func (ns *nodeServer) NodePublishVolume(ctx context.Context, req *csi.NodePublis
 		return &csi.NodePublishVolumeResponse{}, nil
 	}
 
+	if !isValidServer(opt.Server) {
+		return nil, errors.New("invalid server address: mount path must be specified in the 'path' field")
+	}
+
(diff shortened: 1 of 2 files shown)


pkg/nas/nodeserver.go
```diff
@@ -314,6 +314,10 @@ func (ns *nodeServer) NodePublishVolume(ctx context.Context, req *csi.NodePublis
 		return &csi.NodePublishVolumeResponse{}, nil
 	}
 
+	if strings.Contains(opt.Server, ":") {
```
huww98: Maybe someone would specify an IPv6 address in the `Server` field?

mowangdk:
Is there any user requirments about this issue?

pkg/nas/nodeserver.go
huww98: https://github.com/stefanha/nfs-utils/blob/708a963f5eb1892e39db1470d46429d8d8e17e38/utils/mount/parse_dev.c#L210

Here is how mount.nfs parse it. Maybe we can mirror its behavior.

PR #1535, opened 2025-10-29
feat: support erofs filesystem when create disk from a snapshot

AlbeeSo:
#### What type of PR is this?
/kind feature

For OSS VP to Disk.

Fixes #

```release-note

```

```docs

```

### DIFF
--- a/pkg/disk/utils.go
+++ b/pkg/disk/utils.go
@@ -499,11 +499,11 @@ func getDiskVolumeOptions(
 	for _, cap := range req.GetVolumeCapabilities() {
 		mnt := cap.GetMount()
 		if mnt != nil && mnt.FsType != "" {
-			if SupportedFilesystemTypes.Has(mnt.FsType) {
+			// Note: skip filesystem type validation when CreateDisk from a snapshot.
+			if snapshotId != "" {
 				continue
 			}
-			// CreateDisk from a snapshot allows to use erofs
-			if mnt.FsType == EROFS_FSTYPE && snapshotId != "" {
+			if SupportedFilesystemTypes.Has(mnt.FsType) {
 				continue
 			}
 			return nil, fmt.Errorf("fsType %s is not supported, please use %v", mnt.FsType, SupportedFilesystemTypes.UnsortedList())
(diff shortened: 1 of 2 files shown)


pkg/disk/utils.go
```diff
@@ -697,6 +709,9 @@ func validateDiskPerformanceLevel(opts map[string]string) ([]PerformanceLevel, e
 func validateCapabilities(capabilities []*csi.VolumeCapability) (bool, error) {
 	multiAttachRequired := false
 	for _, cap := range capabilities {
+		if cap.GetAccessMode() == nil {
```
mowangdk, line 712: Why is this check necessary? it's not related to the PR?

AlbeeSo: just affect the related UTs. But to avoid nillptr error, adding a check is necessary (or at least there's no any side effect?)

mowangdk:
/lgtm
/approve

cc @hsiangkao

hsiangkao:
> /lgtm /approve
> 
> cc @hsiangkao

That is really a good news!

pkg/disk/constants.go
```diff
@@ -107,10 +107,11 @@ const (
...
 
-	EXT4_FSTYPE = "ext4"
-	EXT3_FSTYPE = "ext3"
-	XFS_FSTYPE  = "xfs"
-	NTFS_FSTYPE = "ntfs"
+	EXT4_FSTYPE  = "ext4"
+	EXT3_FSTYPE  = "ext3"
+	XFS_FSTYPE   = "xfs"
+	NTFS_FSTYPE  = "ntfs"
+	EROFS_FSTYPE = "erofs"
```
huww98, line 114: I think we should remove this. It is only used in tests now.

hsiangkao: it seems
``` go
SupportedFilesystemTypes   = sets.New(EXT4_FSTYPE, EXT3_FSTYPE, XFS_FSTYPE, NTFS_FSTYPE)
```
doesn't change.

huww98: Yes, this is intended. We now skip fstype validation when create from snapshot. We still don't have mkfs.erofs, so SupportedFilesystemTypes is not changed.

PR #1538, opened 2025-10-31
Copy alinas-utils conf files to csi-agent /etc/aliyun-defaults

iltyty:
#### What type of PR is this?
/kind bug

The mount proxy server copies the default alinas-utils configuration files from /etc/aliyun-defaults on startup and will panic if any error occurs during this process.
However, the csi-agent Dockerfile does not create the /etc/aliyun-defaults directory, causing the container to crash in environments where this directory is absent.

Fixes #

```release-note

```

```docs

```

### DIFF
--- a/build/csi-agent/Dockerfile
+++ b/build/csi-agent/Dockerfile
@@ -35,6 +35,8 @@ RUN set -ex; \
     fi; \
     yum clean all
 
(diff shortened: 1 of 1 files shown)


huww98:
/lgtm

Note: when deploying, we need to mount a volume at /etc/aliyun, rather than /etc/aliyun-defaults

PR #1544, opened 2025-11-01
fuse/metrics: add mountpoint metrics for ossfs

AlbeeSo:
#### What type of PR is this?
/kind feature

Add mountpoint status metrics for ossfs:
```
### mount_retry_count
Description: The total number of retry attempts made when mounting fails due to timeout or execution errors.

### mount_point_status
Description: The health status of the mount point, where 1 indicates healthy (successfully mounted and linked), and 0 indicates unhealthy (disconnected or not created).

### mount_point_failover_count
Description: The number of failover attempts performed to remount the mount point after a FUSE client crash or disconnection.

### last_fuse_client_exit_reason
Description: The error code or reason for the last unexpected exit of the FUSE client process.
```

Fixes #

```release-note

```

```docs

```

### DIFF
--- a/pkg/metric/fuse_stat_collector.go
+++ b/pkg/metric/fuse_stat_collector.go
@@ -6,6 +6,7 @@ import (
 	"strconv"
 	"strings"
 
+	"github.com/kubernetes-sigs/alibaba-cloud-csi-driver/pkg/mounter/proxy/server"
 	"github.com/kubernetes-sigs/alibaba-cloud-csi-driver/pkg/utils"
 	"github.com/prometheus/client_golang/prometheus"
 	"k8s.io/klog/v2"
@@ -19,7 +20,12 @@ var (
 	counterTypeArray        = []string{"capacity_counter", "inodes_counter", "throughput_counter", "iops_counter", "latency_counter", "posix_counter", "oss_object_counter"}
 	backendCounterTypeArray = []string{"backend_throughput_counter", "backend_iops_counter", "backend_latency_counter", "backend_meta_qps_ounter"}
 	hotSpotArray            = []string{"hot_spot_read_file_top", "hot_spot_write_file_top", "hot_spot_head_file_top"}
-	mountPointStatusArray   = []string{"mount_retry_count", "mount_point_status", "mount_point_failover_count", "last_fuse_client_exit_reason"}
+	mountPointStatusArray   = []string{
+		server.MetricsMountRetryCount,
+		server.MetricsMountPointStatus,
+		server.MetricsMountPointFailoverCount,
+		server.MetricsLastFuseClientExitReason,
+	}
 )
 
 var (
@@ -310,22 +316,22 @@ var (
 		usFsStatLabelNames, nil,
 	)
 	mountRetryTotalCounterDesc = prometheus.NewDesc(
-		prometheus.BuildFQName(nodeNamespace, volumeSubsystem, "mount_retry_total_counter"),
+		prometheus.BuildFQName(nodeNamespace, volumeSubsystem, server.MetricsMountRetryCount),
 		".",
 		usFsStatLabelNames, nil,
 	)
 	mountPointStatusDesc = prometheus.NewDesc(
-		prometheus.BuildFQName(nodeNamespace, volumeSubsystem, "mount_point_status"),
+		prometheus.BuildFQName(nodeNamespace, volumeSubsystem, server.MetricsMountPointStatus),
 		".",
 		usFsStatLabelNames, nil,
 	)
 	mountPointFailoverTotalCountDesc = prometheus.NewDesc(
-		prometheus.BuildFQName(nodeNamespace, volumeSubsystem, "mount_point_failover_total_counter"),
+		prometheus.BuildFQName(nodeNamespace, volumeSubsystem, server.MetricsMountPointFailoverCount),
 		".",
 		usFsStatLabelNames, nil,
 	)
 	lastFuseClientExitReasonDesc = prometheus.NewDesc(
-		prometheus.BuildFQName(nodeNamespace, volumeSubsystem, "last_fuse_client_exit_reason"),
+		prometheus.BuildFQName(nodeNamespace, volumeSubsystem, server.MetricsLastFuseClientExitReason),
 		".",
 		usFsStatLabelNames, nil,
 	)
@@ -668,7 +674,7 @@ func (p *usFsStatCollector) postBackendCounterMetrics(counterType string, fsClie
 func (p *usFsStatCollector) postMountPointStatusMetrics(statusType string, fsClientInfo *fuseInfo, metricsArray []string, ch chan<- prometheus.Metric) {
 	var err error
 	for _, value := range metricsArray {
-		if statusType == "last_fuse_client_exit_reason" {
+		if statusType == server.MetricsLastFuseClientExitReason {
 			labels := p.getCommonLabels(fsClientInfo, "", value)
 			ch <- p.lastFuseClientExitReason.mustNewConstMetric(1, labels...)
 			continue
@@ -683,11 +689,11 @@ func (p *usFsStatCollector) postMountPointStatusMetrics(statusType string, fsCli
 
 		labels := p.getCommonLabels(fsClientInfo, "", "")
 		switch statusType {
-		case "mount_retry_count":
+		case server.MetricsMountRetryCount:
 			ch <- p.mountRetryTotalCounter.mustNewConstMetric(valueFloat64, labels...)
-		case "mount_point_status":
+		case server.MetricsMountPointStatus:
 			ch <- p.mountPointStatus.mustNewConstMetric(valueFloat64, labels...)
-		case "mount_point_failover_count":
+		case server.MetricsMountPointFailoverCount:
 			ch <- p.mountPointFailoverTotalCounter.mustNewConstMetric(valueFloat64, labels...)
 		}
 	}
@@ -790,7 +796,7 @@ func (p *usFsStatCollector) postVolMetrics(volPath string, fsClientInfo *fuseInf
 	// foreach mountpoint status related metrics
 	for _, mountPointStatus := range mountPointStatusArray {
 		var metricsArray []string
-		if mountPointStatus == "last_fuse_client_exit_reason" {
+		if mountPointStatus == server.MetricsLastFuseClientExitReason {
 			metrics, err := readAllContent(filepath.Join(volPath, mountPointStatus))
 			if err != nil {
 				continue
--- a/pkg/metric/utils.go
+++ b/pkg/metric/utils.go
@@ -46,12 +46,6 @@ func readFirstLines(path string) ([]string, error) {
 
 // readAllContent reads all content from a file and replaces newlines with spaces
 func readAllContent(path string) (string, error) {
-	file, err := os.Open(path)
-	if err != nil {
-		return "", err
-	}
-	defer file.Close()
-
 	content, err := os.ReadFile(path)
 	if err != nil {
 		return "", err
--- a/pkg/mounter/mounter.go
+++ b/pkg/mounter/mounter.go
@@ -11,5 +11,5 @@ type ExtendedMountParams struct {
 
 type Mounter interface {
 	mountutils.Interface
-	ExtendedMount(source, target, fstype string, options []string, params ExtendedMountParams) error
+	ExtendedMount(source, target, fstype string, options []string, params *ExtendedMountParams) error
 }
--- a/pkg/mounter/proxy/server/metrics.go
+++ b/pkg/mounter/proxy/server/metrics.go
@@ -1,7 +1,6 @@
 package server
 
 import (
-	"context"
 	"fmt"
 	"os"
 	"os/exec"
@@ -10,11 +9,9 @@ import (
 	"sync"
 	"time"
 
-	"github.com/kubernetes-sigs/alibaba-cloud-csi-driver/pkg/utils"
 	"k8s.io/klog/v2"
 	"k8s.io/mount-utils"
 	mountutils "k8s.io/mount-utils"
-	"k8s.io/utils/ptr"
 )
 
 const (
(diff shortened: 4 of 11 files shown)


pkg/mounter/proxy/server/metrics.go
```diff
@@ -0,0 +1,328 @@
...
+func (m *MountMonitor) handleMountResult(process *exec.Cmd, err error) {
+	m.mu.Lock()
+	defer m.mu.Unlock()
+
+	if err != nil {
+		// Mount failed
+		m.updateMountPointMetrics(ptr.To(m.retryCount), nil, ptr.To(true), err)
+		m.State = MonitorStateInitialized // Back to initialized for retry
+
+		klog.V(2).InfoS("Mount failed, updated retry count",
```
mowangdk: Use klog.Error in this case.

pkg/mounter/proxy/server/metrics.go
```diff
@@ -0,0 +1,328 @@
...
+func (m *MountMonitor) updateMountPointMetrics(
+	retryCont *int,
+	failoverCont *int,
+	notHealthy *bool,
+	lastExistError error) {
+	if retryCont != nil {
+		// Update mount_retry_count
+		retryFile := filepath.Join(m.MetricsPath, MetricsMountRetryCount)
+		if err := utils.WriteAndSyncFile(retryFile, []byte(strconv.Itoa(*retryCont)), 0644); err != nil {
+			klog.Errorf("Failed to update %s: %v", MetricsMountRetryCount, err)
```
mowangdk: Redefine all klog calls with klog.ErrorS for consistency.

pkg/mounter/proxy/server/metrics.go
```diff
@@ -0,0 +1,328 @@
...
+	MetricsLastFuseClientExitReason = "last_fuse_client_exit_reason"
+)
+
+var maxCountRecord int = 999
+
+// MonitorState represents the state of a monitor
+type MonitorState int
+
+const (
+	MonitorStateInitialized MonitorState = iota
```
mowangdk, line 30: I think we lack the necessary monitor state here. At least we also have a terminating monitor state.

pkg/mounter/proxy/server/metrics.go
```diff
@@ -0,0 +1,328 @@
...
+
+// GetMountMonitor creates a new monitor or returns existing one
+func (manager *MountMonitorManager) GetMountMonitor(target, metricsPath string, raw mount.Interface, createIfNotExist bool) *MountMonitor {
+	// Check if monitor already exists
+	if existingMonitor, exists := manager.monitors.Load(target); exists {
+		monitor := existingMonitor.(*MountMonitor)
+		// Not allow to update info like metrics path currently
+		monitor.mu.Lock()
+		defer monitor.mu.Unlock()
+		if monitor.retryCount < maxCountRecord {
```
mowangdk: Why do we add retryCount when getting monitor objects? it's really confusing

pkg/metric/fuse_stat_collector.go
```diff
@@ -582,33 +637,62 @@ func (p *usFsStatCollector) postBackendCounterMetrics(counterType string, fsClie
...
+		var valueFloat64 float64
+		valueFloat64, err = strconv.ParseFloat(value, 64)
+		if err != nil {
+			klog.Errorf("Convert value %s to float64 is failed, err:%s, counterType:%s, metricsArray:%+v", value, err, statusType, metricsArray)
+			continue
+		}
+
+		labels := p.getCommonLabels(fsClientInfo, "", "")
+		switch statusType {
+		case "mount_retry_count":
```
mowangdk: Use MetricsMountRetryCount instead?

pkg/metric/fuse_stat_collector.go
```diff
@@ -703,4 +787,21 @@ func (p *usFsStatCollector) postVolMetrics(volPath string, fsClientInfo *fuseInf
 		}
 		p.postBackendCounterMetrics(backendCounterType, fsClientInfo, metricsArray, ch)
 	}
+	// foreach mountpoint status related metrics
+	for _, mountPointStatus := range mountPointStatusArray {
+		var metricsArray []string
+		if mountPointStatus == "last_fuse_client_exit_reason" {
```
mowangdk: use MetricsLastFuseClientExitReason

pkg/mounter/mounter.go
```diff
@@ -4,7 +4,12 @@ import (
...
 
+type ExtendedMountParams struct {
+	Secrets     map[string]string
+	MetricsPath string
+}
+
 type Mounter interface {
 	mountutils.Interface
-	MountWithSecrets(source, target, fstype string, options []string, secrets map[string]string) error
+	ExtendedMount(source, target, fstype string, options []string, params ExtendedMountParams) error
```
mowangdk: I'd prefer adding a new function over altering the existing interface completely

pkg/mounter/proxy/server/ossfs/driver.go
```diff
@@ -136,6 +152,10 @@ func (h *Driver) Mount(ctx context.Context, req *proxy.MountRequest) error {
 	})
 
 	if err == nil {
+		// Handle mount result
+		server.SafeHandleMountResult(monitor, cmd, err)
```
mowangdk: Should the statement be placed outside the if condition?

pkg/mounter/proxy/server/ossfs2/driver.go
```diff
@@ -142,6 +157,10 @@ func (h *Driver) Mount(ctx context.Context, req *proxy.MountRequest) error {
 	})
 
 	if err == nil {
+		// Handle mount result
+		server.SafeHandleMountResult(monitor, cmd, err)
```
mowangdk: Quite confuse about handling mount results when err is nil versus when it's not.

pkg/utils/util.go
```diff
@@ -610,8 +612,12 @@ func GetFuseMetricsMountDir(metricsPathPrefix, volumeId string) string {
 	return filepath.Join(metricsPathPrefix, fmt.Sprintf("%x", sha256.Sum256([]byte(volumeId))))
 }
 
-func WriteSharedMetricsInfo(metricsPathPrefix string, req *csi.NodePublishVolumeRequest, clientName string, storageBackendName string, fsName string, sharedPath string) {
-	mountPointPath := GetFuseMetricsMountDir(metricsPathPrefix, req.GetVolumeId())
+func WriteSharedMetricsInfo(metricsPathPrefix string,
```
mowangdk, line 615: Add unit tests for independent functions.

pkg/mounter/proxy/server/metrics.go
AlbeeSo: Monitor will be terminated only when the mount-proxy mounter get a SIGTERM, so I ignore the terminating state.  When the ossfs exits, `handleMountResult` func will lock and set the state to MonitorStateInitialized and stop the regularly `checkAndUpdateMountStatus` overwriting the last exit reason.

pkg/mounter/mounter.go
AlbeeSo: And the number of the mount functions will be increasing?

mowangdk: Yeah, please add one new function, it'will be safer

pkg/mounter/proxy/server/metrics.go
```diff
@@ -0,0 +1,332 @@
...
+		failoverCount: 0,
+	}
+
+	manager.monitors.Store(target, monitor)
+
+	klog.InfoS("Created new monitor", "target", target)
+	return monitor, false
+}
+
+func (m *MountMonitor) IncreaseMountRetryCount() {
```
mowangdk, line 98: ReUse Count is a more appropriate choice.

pkg/metric/utils.go
```diff
@@ -44,6 +44,24 @@ func readFirstLines(path string) ([]string, error) {
 	return lineStrArray, scanner.Err()
 }
 
+// readAllContent reads all content from a file and replaces newlines with spaces
+func readAllContent(path string) (string, error) {
+	file, err := os.Open(path)
+	if err != nil {
+		return "", err
+	}
+	defer file.Close()
```
mowangdk: Simply call os.ReadFile instead of opening twice within this function.

pkg/mounter/proxy/server/ossfs/driver.go
```diff
@@ -50,10 +53,20 @@ func (h *Driver) Fstypes() []string {
 
 func (h *Driver) Mount(ctx context.Context, req *proxy.MountRequest) error {
 	options := req.Options
+	target := req.Target
+
+	// Get or create monitor for this target
+	monitor, found := h.monitorManager.GetMountMonitor(target, req.MetricsPath, h.raw, true)
```
mowangdk: Is a monitor required for the Mount function in the future?  If so, promptly verify if it's nil after Get.

pkg/mounter/proxy/server/ossfs2/driver.go
```diff
@@ -50,10 +52,20 @@ func (h *Driver) Fstypes() []string {
 
 func (h *Driver) Mount(ctx context.Context, req *proxy.MountRequest) error {
 	options := req.Options
+	target := req.Target
+
+	// Get or create monitor for this target
+	monitor, found := h.monitorManager.GetMountMonitor(target, req.MetricsPath, h.raw, true)
```
mowangdk: Same question with ossfs

pkg/mounter/proxy/server/metrics.go
```diff
@@ -0,0 +1,332 @@
...
+// initializeMetrics initializes metrics files to empty/zero state
+func (m *MountMonitor) updateMountPointMetrics(
+	retryCont *int,
+	failoverCont *int,
+	notHealthy *bool,
+	lastExistError error) {
+	if retryCont != nil {
+		// Update mount_retry_count
+		retryFile := filepath.Join(m.MetricsPath, MetricsMountRetryCount)
+		if err := utils.WriteAndSyncFile(retryFile, []byte(strconv.Itoa(*retryCont)), 0644); err != nil {
```
huww98: Do we need to sync? If the host goes down unexpectly, I think we can accept the metrics being lost.

pkg/mounter/proxy/server/metrics.go
```diff
@@ -0,0 +1,332 @@
...
+	if newUnhealthy != oldUnhealthy {
+		klog.Warningf("Mount point status changed. but the detailed reason was missing")
+		// Update status
+		m.updateMountPointMetrics(nil, nil, ptr.To(newUnhealthy), err)
+	}
+}
+
+func boolToBinaryString(b bool) string {
+	boolMap := map[bool]string{true: "1", false: "0"}
+	return boolMap[b]
```
huww98:
```suggestion
	if b {
	    return "1"
	}
	return "0"
```
Using a map is too heavy.

pkg/mounter/proxy/server/metrics.go
```diff
@@ -0,0 +1,332 @@
...
+	}
+}
+
+// HandleMountFailure handles the case when mount operation fails
+func (m *MountMonitor) HandleMountFailure(err error) {
+	m.mu.Lock()
+	defer m.mu.Unlock()
+
+	// Update metrics for mount failure
+	m.updateMountPointMetrics(ptr.To(m.retryCount), nil, ptr.To(true), err)
```
huww98:
```suggestion
	m.updateMountPointMetrics(&m.retryCount, nil, ptr.To(true), err)
```
Please only use ptr.To for literal

pkg/mounter/proxy/server/metrics_test.go
```diff
@@ -0,0 +1,73 @@
...
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			// Clean up any existing files
+			for filename := range tt.wantFiles {
+				file := filepath.Join(tempDir, filename)
+				os.Remove(file)
+			}
+
+			// Call the function
+			m.updateMountPointMetrics(nil, nil, ptr.To(tt.handleErr != nil), tt.handleErr)
```
huww98:
```suggestion
			m.updateMountPointMetrics(nil, nil, tt.handleErr)
```
So should non-nil err already indicates unhealthy? do we need the third parameter?

pkg/mounter/proxy/server/metrics.go
```diff
@@ -0,0 +1,332 @@
...
+	monitorCancel context.CancelFunc
+	raw           mount.Interface
+	// Mount retry count (persistent across mount attempts)
+	retryCount    int
+	failoverCount int
+}
+
+// MountMonitorManager manages multiple mount monitors
+type MountMonitorManager struct {
+	monitors *sync.Map // map[string]*MountMonitor - key is target path
```
huww98:
```suggestion
	monitors sync.Map // map[string]*MountMonitor - key is target path
```

pkg/mounter/proxy/server/metrics.go
huww98: That means if user constantly create and mount new volumes on a node, we will have unbounded number of monitors?

pkg/mounter/proxy/server/metrics.go
```diff
@@ -0,0 +1,332 @@
...
+
+// StartMonitoring starts the monitoring goroutine for this monitor
+func (manager *MountMonitorManager) StartMonitoring(target string) {
+	manager.wg.Add(1)
+	monitor, _ := manager.GetMountMonitor(target, "", nil, false)
+	if monitor == nil {
+		// No monitor found; nothing to start
+		manager.wg.Done()
+		err := fmt.Errorf("monitor not found")
+		klog.ErrorS(err, "StartMonitoring failed", "target", target)
```
huww98:
```suggestion
		klog.ErrorS(nil, "monitor not found", "target", target)
```
You can pass nil to ErrorS

pkg/mounter/proxy/server/ossfs2/driver.go
```diff
@@ -142,6 +164,11 @@ func (h *Driver) Mount(ctx context.Context, req *proxy.MountRequest) error {
 	})
 
 	if err == nil {
+		if monitor != nil {
+			monitor.HandleMountSuccess(cmd)
+			// Start monitoring goroutine (ticker based only)
+			h.monitorManager.StartMonitoring(target)
```
huww98, line 176: Can we merge `StartMonitoring` into `HandleMountSuccess`? They seems always be called together.

pkg/mounter/proxy/server/metrics.go
```diff
@@ -0,0 +1,332 @@
...
+	m.State = MonitorStateMonitoring
+	m.Pid = process.Process.Pid
+
+	klog.InfoS("Mount succeeded", "target", m.Target, "pid", m.Pid)
+}
+
+// Stop stops the monitoring
+func (m *MountMonitor) Stop() {
+	close(m.stopCh)
+	m.monitorCancel()
```
huww98: Why do we need both stopChannel and context? One of them should be enough?

pkg/mounter/proxy/server/ossfs2/driver.go
AlbeeSo: it works, and HandleMountSucess will be a function of monitor-manager, and HandleMountFailure is a function of monitor (or else find it in the sync.map everytime), it seems a little strange...

PR #1554, opened 2025-11-07
Disable additional cgroup of efc container

mowangdk:
#### What type of PR is this?
/kind bug

This options filter invalid OS kernel versions for bmcpfs, hence their deletion is not possible."

Disable additional cgroup of containerized EFC

Fixes #

```release-note

```

```docs

```

### DIFF
--- a/build/csi-agent/Dockerfile
+++ b/build/csi-agent/Dockerfile
@@ -18,7 +18,7 @@ ARG TARGETPLATFORM
 # The versions specified here are for RPM packages, which may not necessarily align with the image version
 ARG OSSFS_VERSION=v1.91.8.ack.3
 ARG OSSFS2_VERSION=v2.0.4.ack.1
-ARG ALINAS_UTILS_VERSION=1.9-0.20250805171424.2871b7.al7
+ARG ALINAS_UTILS_VERSION=1.9-0.20251120211108.7e6a18.al7
 ARG EFC_VERSION=1.9-20250822095129.f34347.release
 # v2.1 is exclusive to a specific storage cluster version and the first to support ARM arch. 
 # It's temporarily unavailable for all ACK clusters, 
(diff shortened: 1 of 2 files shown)


huww98:
This already have `isEFC && !isVSC` guard. Is it still not enough? What does this option means?

mowangdk:
> auto_fallback_nfs is exclusive to NAS, integrating it into bmcpfs causes misleading error messages

As description said, this option is exclusive for NAS, not compatible with BMCPFS. EFC is available for both NAS and BMCPFS.

huww98:
By investigating the git log, current logic is already tuned for for bmcpfs (a3f5c05c3adc99787b916e9da8c71d5d09acfe89). So if you are sure, please comment about why that adjustment is not enough, and just remove `addAutoFallbackNFSMountOptions` instead of comment it out.

pkg/nas/utils.go
```diff
@@ -112,6 +113,7 @@ func doMount(mounter mountutils.Interface, opt *Options, targetPath, volumeId, p
 		}
 		if opt.Accesspoint != "" {
 			mountFstype = "alinas"
+			combinedOptions = append(combinedOptions, "auto_fallback_nfs")
```
huww98: AccessPoint does not implies EFC, right? Can we really use this without efc?

huww98:
If we use a csi with this change and a mount-proxy without this change, we may ended up adding `auto_fallback_nfs` twice. Is this supported?

pkg/nas/utils.go
iltyty: +1, this option is exclusive to EFC, so we don't need it here.

pkg/nas/utils.go
```diff
@@ -93,6 +94,11 @@ func doMount(mounter mountutils.Interface, opt *Options, targetPath, volumeId, p
 			return errors.New("EFC Client don't support this storage type:" + opt.FSType)
 		}
 		mountFstype = "alinas"
+		combinedOptions = append(combinedOptions, "auto_fallback_nfs")
+		if _, ok := mounter.(*csimounter.ProxyMounter); ok {
+			klog.InfoS("using proxy mounter, skip no_add_cgroup option")
```
iltyty: skip -> add?

PR #1555, opened 2025-11-07
Support disk available label for lingjun node type

mowangdk:
#### What type of PR is this?
/kind feature

Support disk available labels for lingjun node

Fixes #

```release-note

```

```docs

```

### DIFF
--- a/pkg/disk/utils.go
+++ b/pkg/disk/utils.go
@@ -1428,8 +1428,5 @@ func isLingjunNode(node *v1.Node) bool {
 	if node == nil {
 		return false
 	}
-	if value := node.Labels["alibabacloud.com/lingjun-worker"]; value == "true" {
-		return true
-	}
-	return false
+	return node.Labels["alibabacloud.com/lingjun-worker"] == "true"
 }


pkg/disk/disk_test.go
```diff
@@ -15,3 +15,111 @@ limitations under the License.
...
+
+			for _, v := range tt.expected {
+				expectedSet[v] = true
+			}
+
+			for _, v := range result {
+				resultSet[v] = true
+			}
+
+			assert.Equal(t, expectedSet, resultSet, "Mismatch between expected and actual disk types")
```
huww98: Just use `assert.ElementsMatch`

pkg/disk/disk.go
```diff
@@ -261,3 +262,41 @@ func newBatcher(fromNode bool) (waitstatus.StatusWaiter[ecs.Disk], batcher.Batch
 func defaultThrottler() *throttle.Throttler {
 	return throttle.NewThrottler(clock.RealClock{}, 1*time.Second, 10*time.Second)
 }
+
+// parseLingjunNodeDiskTypes parses allowed disk types for Lingjun nodes from a comma-separated string.
+// If value is nil or empty, default to essd_auto & cloud_essd.
+// Example: "cloud_essd,cloud_auto". Unknown entries are ignored.
```
huww98, line 268: What is the propose if ignoring unknown entries? Just apply anything specified by user should be easy and simple. Also future proof if we introduce additional disk types.

pkg/disk/utils.go
```diff
@@ -1408,3 +1411,25 @@ func (d DiskSize) String() string {
...
+			return true
+		}
+	}
+	if node == nil {
+		return false
+	}
+	if value := node.Labels["alibabacloud.com/lingjun-worker"]; value == "true" {
+		return true
+	}
+	return false
```
huww98:
```suggestion
	return node.Labels["alibabacloud.com/lingjun-worker"] == "true"
```

pkg/disk/utils_test.go
```diff
@@ -989,3 +989,138 @@ func TestGetZone(t *testing.T) {
...
+				// Ensure the file does not exist
+				_ = os.Remove(LingjunConfigFile)
+			}
+
+			// Run the test
+			result := isLingjunNode(tt.node)
+			assert.Equal(t, tt.expected, result)
+
+			// Clean up for next test
+			_ = os.Remove(LingjunConfigFile)
```
huww98: Just call t.TempDir() in the t.Run. Test framework will use different subdir for sub-test.

pkg/disk/nodeserver.go
```diff
@@ -214,15 +215,17 @@ func NewNodeServer(m metadata.MetadataProvider) csi.NodeServer {
...
-		metadata:     m,
-		mounter:      utils.NewMounter(),
-		kataBMIOType: kataBMIOType,
-		k8smounter:   k8smount.NewWithoutSystemd(""),
-		podCGroup:    podCgroup,
-		clientSet:    GlobalConfigVar.ClientSet,
+		metadata:                    m,
+		mounter:                     utils.NewMounter(),
+		kataBMIOType:                kataBMIOType,
+		ebsCategoryLingjunSupported: ebsCategoryLingjunSupported,
```
huww98: You already have metadata stored here, so no need to add ebsCategoryLingjunSupported?

pkg/disk/nodeserver.go
```diff
@@ -840,17 +843,25 @@ func (ns *nodeServer) NodeGetInfo(ctx context.Context, req *csi.NodeGetInfoReque
 	}
 	var diskTypes []string
 	if !GlobalConfigVar.DiskAllowAllType {
-		diskTypes, err = GetAvailableDiskTypes(ctx, c, ns.metadata)
-		if err != nil {
-			if hasDiskTypeLabel(node) {
-				klog.ErrorS(err, "NodeGetInfo: failed to get available disk types, will use existing config.")
+		// If it's a Lingjun node, reuse Lingjun detection and report configured supported types
+		if isLingjunNode(node) {
+			// Prefer values parsed from kube-system/csi-plugin configmap
```
huww98, line 846: I suppose this is read from env EBS_CATEGORY_LINGJUN_SUPPORTED?

huww98 (changes requested):
What if we have multiple LingJun instance types that supports different disk types?

pkg/disk/nodeserver.go
huww98: I'd recommend just os.GetEnv() here, and not use metadata until we really have multiple places to get this information.

pkg/disk/utils.go
```diff
@@ -1408,3 +1411,25 @@ func (d DiskSize) String() string {
 	}
 	return fmt.Sprintf("%.3f GiB (0x%X)", float64(d.Bytes)/GBSIZE, d.Bytes)
 }
+
+// isLingjunNode checks if the node is a Lingjun node.
+// Priority:
+// 1) If the given node is the local node (matches env KUBE_NODE_NAME), detect by lingjun_config file like pkg/bmcpfs
+// 2) Fallback to node label alibabacloud.com/lingjun-worker == true
+func isLingjunNode(node *v1.Node) bool {
```
huww98, line 1419: Can we tell if a node is Lingjun instance from metadata server? I feel this can fit into cloud/metadata package.

mowangdk:
> What if we have multiple LingJun instance types that supports different disk types?

By then, the lingjun team will be required to provide an openapi.

huww98:
Can we have something more general like `EBS_CATEGORY_FALLBACK`, which is used whenever we cannot retrieve it from OpenAPI? e.g. we don't have RAM credential, or OpenAPI returned empty result.

The overall process would be:

1. Retrieve from OpenAPI
2. when unable, use the existing labels on the node object (so that user can manually config each node)
3. when not present, use EBS_CATEGORY_FALLBACK (we set EBS_CATEGORY_FALLBACK to `cloud_auto,cloud_essd` on LingJun node pool in ACK)
4. when not present, fail the NodeGetInfo

This should work for both ECS and LingJun nodes.

pkg/disk/utils_test.go
huww98: You removed this line, but you didn't call t.TempDir() in the t.Run()...

pkg/disk/utils.go
huww98: You seems reverted this change.

pkg/disk/utils_test.go
```diff
@@ -989,3 +989,131 @@ func TestGetZone(t *testing.T) {
...
+			// Setup config file if needed
+			if !tt.configError {
+				if tt.configContent != "" {
+					err := os.WriteFile(LingjunConfigFile, []byte(tt.configContent), 0644)
+					assert.NoError(t, err)
+				} else {
+					// Create an empty file
+					err := os.WriteFile(LingjunConfigFile, []byte{}, 0644)
+					assert.NoError(t, err)
+				}
```
```suggestion
			    assert.NoError(t, os.WriteFile(LingjunConfigFile, []byte(tt.configContent), 0644))
```
I think we don't need this branch.

```suggestion
			    assert.NoError(t, os.WriteFile(LingjunConfigFile, []byte(tt.configContent), 0644))
```
I think we don't need this branch.

pkg/disk/utils.go
huww98:
```suggestion
	return node.Labels["alibabacloud.com/lingjun-worker"] == "true"
```
Again

PR #1557, opened 2025-11-12
upgrade base image to debian 13

huww98:
#### What type of PR is this?

/kind cleanup

Fixes #

Now all binaries are under /usr, we can remove the logic to search both.

```release-note
Upgrade base image to debian 13
```

```docs

```

/hold
~For upstream distroless image is still preview.~ Released now


huww98:
XFS (nvme5n1): Superblock has unknown incompatible features (0x20) enabled.

Looks like mkfs.xfs is too new and enabled some features that is not supported by alinux3 kernel.

```
#define XFS_SB_FEAT_INCOMPAT_NREXT64	(1 << 5)  /* large extent counters */
```

Available at Linux v5.19 https://github.com/torvalds/linux/commit/919819f5e18097e6e888764c30625b1288d416c5

pkg/utils/util.go
```diff
@@ -673,6 +673,7 @@ func mkfsDefaultArgs(fstype, source string) (args []string) {
 	} else if fstype == "xfs" {
 		args = []string{
 			"-f",
+			"-i", "nrext64=0", // This requires kernel v5.19, so disable for now.
```
huww98, line 651: another reason to hold this. We will need to replace ASI dockerfile before merge. Old mkfs.xfs will not understand this flag.

mowangdk (commented):
## Automated Code Review

**Summary:** Upgrades the base container image from Debian 12 (Bookworm) to Debian 13 (Trixie), updating all dependent package versions and adjusting build scripts for filesystem layout changes (`/sbin` → `/usr/sbin`).

**Issues Found:**
- Path changes (`/sbin/` → `/usr/sbin/`) for `fsck`, `mkfs`, `mount`, `umount`, `resize2fs` — these must match the actual filesystem layout in Debian 13. The usrmerge in Debian 13 should have symlinks, but the explicit paths are safer.
- The `gcc` package addition in the build stage is well-documented with the Debian bug reference — good.
- The `configure` script now dumps `config.log` on failure — helpful for debugging cross-compilation issues.
- New dependencies added: `libzstd1`, `zlib1g` — verify these are actually needed by the runtime binaries.
- Package version pinning is very specific (e.g., `libc6 2.41-12+deb13u1`). This will need updates as Debian 13 receives security patches.

**Suggestions:**
- Consider running the full test suite with the new base image in CI before merging to catch any subtle behavioral differences.
- The `libext2fs2t64` naming (with `t64` suffix) is the time_t transition — this is expected for Debian 13 on 32-bit archs but verify it resolves correctly on amd64.

**Overall:** Straightforward base image upgrade with careful attention to Debian 13's filesystem changes. The explicit path adjustments and build toolchain fixes demonstrate good testing.

pkg/utils/util.go
```diff
@@ -648,6 +648,7 @@ func mkfsDefaultArgs(fstype, source string) (args []string) {
...
```
mowangdk, line 651: The comment says `nrext64=0` requires kernel v5.19 so it's disabled — but the flag is being explicitly set to 0, which *disables* the feature. The comment is correct but slightly confusing. Consider rewording to: `nrext64=0: disable large extent counters for compatibility with kernels < v5.19`.

mowangdk (commented):
## Automated Code Review

**Summary:** Upgrades the base image from Debian 12 (bookworm) to Debian 13 (trixie), updates all dependency versions, and adjusts paths for binaries that moved from `/sbin` to `/usr/sbin`.

**Overall:** Thorough upgrade with appropriate adjustments for the new filesystem layout. One concern about the xfs flag.

build/gather-node-deps.sh
```diff
@@ -6,10 +6,10 @@ mkdir -p /staging-node/var/lib/dpkg/status.d
 
 DEPS=(
     /etc/netconfig
-    /etc/mke2fs.conf /sbin/{fsck,mkfs,mount,umount}.{ext{2,3,4},xfs,nfs}
+    /etc/mke2fs.conf /usr/sbin/{fsck,mkfs,mount,umount}.{ext{2,3,4},xfs,nfs}
```
mowangdk, line 9: Good catch updating paths from `/sbin/` to `/usr/sbin/` for Debian 13 (which no longer uses the `/sbin` symlink split). Verify that `/usr/sbin/mount.nfs` and `/usr/sbin/umount.nfs` actually exist in trixie — historically NFS mount helpers may be in different locations depending on the nfs-common package version.

PR #1560, opened 2025-11-17
Add max ebs volume count for lingjun node 

mowangdk:
Add max ebs volume count for lingjun node 

Fixes #

```release-note

```

```docs

```


pkg/disk/utils.go
```diff
@@ -1165,7 +1215,46 @@ func getAvailableDiskCountFromOpenAPI(ecsClient cloud.ECSInterface, m metadata.M
...
+	}
+	resp, err := efloClient.DescribeNodeType(req)
+	if err != nil {
+		return 0, fmt.Errorf("DescribeNodeType failed: %w", err)
+	}
+	klog.InfoS("getAvailableCountFromEFLOpenAPI: ", "resp", resp)
+	return int(*resp.Body.DiskQuantity), nil
+}
+
+func getNodeTypeFromEFLOpenAPI(efloClient cloud.EFLOInterface, lingjunID string) (string, error) {
```
huww98: Can we get this from metadata server? If so (either now or in the future), I hope this can go into cloud/metadata package.

I'd expect we still use `InstanceID`/`InstanceType` key for LingJun instance. And maybe add a new `LingJunWorker` key to cloud/metadata. Also add two new providers: `LingJunConfigFile`, `EFLOOpenAPI`

pkg/disk/utils.go
```diff
@@ -1416,17 +1505,17 @@ func (d DiskSize) String() string {
...
-			return true
+			return cfg.NodeId, true
 		}
 	}
 	if node == nil {
-		return false
+		return "", false
 	}
-	return node.Labels["alibabacloud.com/lingjun-worker"] == "true"
+	return "", node.Labels["alibabacloud.com/lingjun-worker"] == "true"
```
huww98: So we will have empty ID in this case. I think this is not OK.

pkg/disk/utils_test.go
```diff
@@ -300,26 +321,35 @@ func TestGetAvailableDiskCount(t *testing.T) {
...
 		t.Run(tt.name, func(t *testing.T) {
-			actual, _ := getAvailableDiskCount(tt.node, ecsClient, testMetadata)
+			actual, _ := getAvailableDiskCount(tt.node, ecsClient, efloClient, testMetadata, tt.lingjunNodeId)
 			assert.Equal(t, tt.expected, actual)
 		})
 	}
 }
 
 func TestGetAvailableDiskCountFromAnnotation(t *testing.T) {
+	efloClient := cloud.NewMockEFLOInterface(gomock.NewController(t))
```
huww98: This variable is never used?

pkg/disk/utils_test.go
```diff
@@ -1090,24 +1141,197 @@ func TestIsLingjunNode(t *testing.T) {
...
+			expectedID:    "node-456",
+			expectedBool:  true,
 		},
 	}
 
 	for _, tt := range tests {
 		t.Run(tt.name, func(t *testing.T) {
+			// Set KUBE_NODE_NAME environment variable
+			if tt.nodeNameEnv != "" {
+				os.Setenv("KUBE_NODE_NAME", tt.nodeNameEnv)
```
```suggestion
				t.Setenv("KUBE_NODE_NAME", tt.nodeNameEnv)
```
And you don't need to restore it manually.

pkg/disk/utils.go
```diff
@@ -1165,7 +1215,46 @@ func getAvailableDiskCountFromOpenAPI(ecsClient cloud.ECSInterface, m metadata.M
...
+		return 0, err
+	}
+	klog.InfoS("getLingjunAvailableDiskCount: ", "nodeType", nodeType)
+	if nodeType == "" {
+		return 0, fmt.Errorf("cannot get node type for lingjun %s", lingjunID)
+	}
+	return getAvailableCountFromEFLOpenAPI(efloClient, nodeType)
+}
+
+func getAvailableCountFromEFLOpenAPI(efloClient cloud.EFLOInterface, nodeType string) (int, error) {
```
huww98:
```suggestion
func getAvailableCountFromEFLOOpenAPI(efloClient cloud.EFLOInterface, nodeType string) (int, error) {
```

huww98:
https://github.com/kubernetes-sigs/alibaba-cloud-csi-driver/blob/837977880d403ea7b967c6ee71e2005fbbfc8f47/pkg/disk/disk.go#L116

We are using InstanceID from metadata as NodeID, is this still accurate for LingJun instances?

pkg/disk/utils.go
mowangdk: Currently unsupported, but considering moving node type to metadata in the future if suppot. No alternative method exists for the time being.

pkg/disk/utils.go
mowangdk: I'm concerned about it too, I'll remove it.

pkg/disk/utils_test.go
```diff
@@ -345,12 +375,19 @@ func TestGetAvailableDiskCountFromAnnotation(t *testing.T) {
...
 	}
 
 	for _, tt := range tests {
 		t.Run(tt.name, func(t *testing.T) {
+			// Mock Eflo response
+			efloResp := &eflo.DescribeNodeTypeResponse{
+				Body: &eflo.DescribeNodeTypeResponseBody{DiskQuantity: tt.quantity},
+			}
+			efloClient := cloud.NewMockEFLOInterface(gomock.NewController(t))
+			efloClient.EXPECT().DescribeNodeType(gomock.Any()).Return(efloResp, nil).AnyTimes()
```
huww98: Still unused. I think we can just revert all the changes to this test-case. It reads from Node annotation, which is irrelevant to LingJun instances

pkg/disk/utils_test.go
```diff
@@ -994,23 +1032,28 @@ func TestIsLingjunNode(t *testing.T) {
 
 	// Save the original LingjunConfigFile value
 	originalLingjunConfigFile := LingjunConfigFile
+	originalNodeName := os.Getenv("KUBE_NODE_NAME")
```
huww98: Revert this one too, as we now have `t.Setenv`

pkg/disk/utils_test.go
```diff
@@ -314,12 +341,14 @@ func TestGetAvailableDiskCountFromAnnotation(t *testing.T) {
 	tests := []struct {
 		name     string
 		node     *corev1.Node
+		quantity *int32
```
huww98: This can also be reverted

pkg/disk/utils.go
```diff
@@ -1413,20 +1502,14 @@ func (d DiskSize) String() string {
 }
 
 // isLingjunNode checks if the node is a Lingjun node.
-// Priority:
-// 1) If the given node is the local node (matches env KUBE_NODE_NAME), detect by lingjun_config file like pkg/bmcpfs
-// 2) Fallback to node label alibabacloud.com/lingjun-worker == true
-func isLingjunNode(node *v1.Node) bool {
+func isLingjunNode() string {
```
huww98:
```suggestion
func getLingjunNodeID() string {
```

huww98:
Please also update docs/ram-policies/disk/node.json

pkg/disk/nodeserver.go
```diff
@@ -830,18 +830,21 @@ func (ns *nodeServer) NodeGetInfo(ctx context.Context, req *csi.NodeGetInfoReque
...
 	}
 	if err != nil {
 		return nil, err
 	}
+
 	var diskTypes []string
 	if !GlobalConfigVar.DiskAllowAllType {
 		// If it's a Lingjun node, reuse Lingjun detection and report configured supported types
-		if isLingjunNode(node) {
+		if ns.NodeID != "" {
```
huww98: `ns.NodeID` is always not empty. You need another way to distinguish Lingjun node.

pkg/disk/utils.go
```diff
@@ -1165,7 +1215,46 @@ func getAvailableDiskCountFromOpenAPI(ecsClient cloud.ECSInterface, m metadata.M
...
+	}
+	resp, err := efloClient.DescribeNodeType(req)
+	if err != nil {
+		return 0, fmt.Errorf("DescribeNodeType failed: %w", err)
+	}
+	klog.InfoS("getAvailableCountFromEFLOOpenAPI: ", "resp", resp)
+	return int(*resp.Body.DiskQuantity), nil
+}
+
+func getNodeTypeFromEFLOpenAPI(efloClient cloud.EFLOInterface, lingjunID string) (string, error) {
```
huww98:
```suggestion
func getNodeTypeFromEFLOOpenAPI(efloClient cloud.EFLOInterface, lingjunID string) (string, error) {
```

PR #1562, opened 2025-11-20
Support NAS access point RAM authentication with AK/SK

iltyty:
#### What type of PR is this?
/kind feature

* Add support for NAS access point RAM authentication with AK/SK.
* Refactor pkg/mounter: extract shared mounting logic into reusable mount interceptors.

Fixes #

Primary Changes:
1. Extract non-mounting logic from mounters into dedicated mount interceptors. For example, operations such as credential persistence and monitoring setup during OSS mounting are now handled by interceptors rather than embedded in the mounter implementations.
2. Wrap all mounters with pkg/mounter.MounterWorkflow. This workflow orchestrates the mounting process as follows:
   - Invokes the BeforeMount method of all registered interceptors.
   - Performs the actual mount using the underlying mounter.
   - Invokes the AfterMount method of all interceptors.
3. Introduce AliyunSecretInterceptor to support RAM-based authentication for NAS access points.

```release-note

```

```docs

```

### DIFF
--- a/pkg/mounter/cmd_mounter.go
+++ b/pkg/mounter/cmd_mounter.go
@@ -28,12 +28,10 @@ func NewOssCmdMounter(execPath, volumeID string, inner mount.Interface) Mounter 
 	}
 }
 
-func (m *OssCmdMounter) ExtendedMount(_ context.Context, op *MountOperation) error {
+func (m *OssCmdMounter) ExtendedMount(ctx context.Context, op *MountOperation) error {
 	if op == nil {
 		return nil
 	}
-	ctx, cancel := context.WithDeadline(context.Background(), time.Now().Add(timeout))
-	defer cancel()
 
 	cmd := exec.CommandContext(ctx, m.execPath, getArgs(op)...)
 	cmd.Stdout = os.Stdout
--- a/pkg/mounter/interceptors/ossfs_secret.go
+++ b/pkg/mounter/interceptors/ossfs_secret.go
@@ -32,7 +32,7 @@ func ossfsSecretInterceptor(ctx context.Context, op *mounter.MountOperation, han
 	}
 	if passwdFile != "" {
 		if fuseType == "ossfs" {
-			op.Args = append(op.Args, "passwd_file="+passwdFile)
+			op.Options = append(op.Options, "passwd_file="+passwdFile)
 		} else {
 			op.Args = append(op.Args, []string{"-c", passwdFile}...)
 		}
--- a/pkg/mounter/interceptors/ossfs_secret_test.go
+++ b/pkg/mounter/interceptors/ossfs_secret_test.go
@@ -45,7 +45,7 @@ func TestOssfsSecretInterceptor(t *testing.T) {
 			expectErr: true,
 		},
 		{
-			name:    "ni mount result",
+			name:    "nil mount result",
 			handler: successMountHandler,
 			op: &mounter.MountOperation{
 				FsType: "ossfs",
@@ -82,10 +82,10 @@ func TestOssfsSecretInterceptor(t *testing.T) {
 				return
 			}
 
-			assert.Len(t, tt.op.Args, 1)
-			assert.Contains(t, tt.op.Args[0], "passwd_file=")
+			assert.Len(t, tt.op.Options, 1)
+			assert.Contains(t, tt.op.Options[0], "passwd_file=")
 
-			filePath := tt.op.Args[0][len("passwd_file="):]
+			filePath := tt.op.Options[0][len("passwd_file="):]
 			if tt.expectFile {
 				assert.FileExists(t, filePath)
 			} else {
@@ -106,10 +106,10 @@ func TestOssfsSecretInterceptor(t *testing.T) {
 	result := op.MountResult.(server.OssfsMountResult)
 	<-result.ExitChan
 
-	assert.Len(t, op.Args, 1)
-	assert.Contains(t, op.Args[0], "passwd_file=")
+	assert.Len(t, op.Options, 1)
+	assert.Contains(t, op.Options[0], "passwd_file=")
 	time.Sleep(500 * time.Millisecond) // Wait for ossfs monitor interceptor to cleanup the credential file
-	assert.NoFileExists(t, op.Args[0][len("passwd_file="):])
+	assert.NoFileExists(t, op.Options[0][len("passwd_file="):])
 }
 
 func TestOssfs2SecretInterceptor(t *testing.T) {
--- a/pkg/nas/mounter.go
+++ b/pkg/nas/mounter.go
@@ -4,6 +4,7 @@ import (
 	"context"
 
 	"github.com/kubernetes-sigs/alibaba-cloud-csi-driver/pkg/mounter"
+	"github.com/kubernetes-sigs/alibaba-cloud-csi-driver/pkg/mounter/interceptors"
 	"k8s.io/klog/v2"
 	mountutils "k8s.io/mount-utils"
 )
@@ -47,7 +48,7 @@ func newNasMounter(agentMode bool, socketPath string) mounter.Mounter {
 	case !agentMode: // normal case, use connector mounter to ensure backward compatibility
 		m.alinasMounter = mounter.NewConnectorMounter(inner, "")
 	default:
-		m.alinasMounter = mounter.NewAdaptorMounter(inner)
+		m.alinasMounter = mounter.NewForMounter(mounter.NewAdaptorMounter(inner), interceptors.AlinasSecretInterceptor)
 	}
 	return m
 }


pkg/oss/nodeserver.go
```diff
@@ -161,20 +167,25 @@ func (ns *nodeServer) NodePublishVolume(ctx context.Context, req *csi.NodePublis
 		if err != nil {
 			return nil, status.Error(codes.Internal, err.Error())
 		}
-		ossfsMounter = mounter.NewOssCmdMounter(ns.ossfsPaths[opts.FuseType], req.VolumeId, ns.rawMounter)
+		ossfsMounter = mounter.NewForMounter(
+			mounter.NewOssCmdMounter(ns.ossfsPaths[opts.FuseType], req.VolumeId, ns.rawMounter),
+			ossInterceptors[opts.FuseType]...,
+		)
 	} else {
 		ossfsMounter = mounter.NewProxyMounter(socketPath, ns.rawMounter)
```
AlbeeSo: Here it looks a bit odd. `Mounter` is a relatively high-level abstraction that includes both `CmdMounter` and `ProxyMounter`. If we define a function like `NewForMounter`, it should be used to create any type of mounter, instead `NewProxyMounter`.  Or we can still use the `NewOssCmdMounter`?

pkg/oss/nodeserver.go
```diff
@@ -66,6 +67,11 @@ const (
 // use unifiedFsType instead
 var unifiedFsType = OssFsType
 
+var ossInterceptors = map[string][]mounter.MountInterceptor{
```
AlbeeSo: It'd better to explicitly indicate (by name or via a comment) that this is currently intended for `CmdMounter` only.

pkg/mounter/interceptors/ossfs_secret.go
```diff
@@ -0,0 +1,35 @@
...
+}
+
+func NewOssfs2SecretInterceptor() mounter.MountInterceptor {
+	return OssfsSecretInterceptor{
+		credFileKey: "-c",
+	}
+}
+
+func (i OssfsSecretInterceptor) BeforeMount(req *mounter.MountOperation) (*mounter.MountOperation, error) {
+	filePath, err := utils.SaveOssSecretsToFile(req.Secrets, req.FsType)
```
AlbeeSo: Seems that both `ProxyMounter` and `CmdMounter` rely on `SaveOssSecretsToFile`. Can we reuse the Interceptor for the former?

pkg/mounter/interceptors/alinas_secret.go
```diff
@@ -0,0 +1,52 @@
...
+func NewAlinasSecretInterceptor() mounter.MountInterceptor {
+	return AlinasSecretInterceptor{}
+}
+
+func (AlinasSecretInterceptor) BeforeMount(op *mounter.MountOperation, _ error) (*mounter.MountOperation, error) {
+	if op == nil || op.Secrets == nil {
+		return op, nil
+	}
+
+	tmpDir, err := os.MkdirTemp("", "alinas-")
```
iltyty: os.TempDir

pkg/mounter/interceptors/alinas_secret.go
```diff
@@ -0,0 +1,52 @@
...
+	}
+
+	credFileContent := makeCredFileContent(op.Secrets)
+	credFilePath := path.Join(tmpDir, op.VolumeID+".credentials")
+	if err = os.WriteFile(credFilePath, credFileContent, 0o600); err != nil {
+		return op, err
+	}
+
+	klog.V(4).InfoS("Created alinas credential file", "path", credFilePath)
+	op.Options = append(op.Options, "ram_config_file="+credFilePath)
```
iltyty, line 45: Test mounting two APs on the same node with different AK/SKs.

pkg/mounter/proxy_mounter.go
```diff
@@ -14,27 +15,25 @@ type ProxyMounter struct {
...
 	}
 }
 
-func (m *ProxyMounter) ExtendedMount(source, target, fstype string, options []string, params *ExtendedMountParams) error {
-	// Parameters in ExtendedMountParams are optional
-	if params == nil {
-		params = &ExtendedMountParams{}
-	}
-
+func (m *ProxyMounter) ExtendedMount(ctx context.Context, op *MountOperation) error {
```
AlbeeSo, line 27: Is it necessary to check if the op is nil?

pkg/oss/nodeserver.go
AlbeeSo:
```
		ossfsMounter = mounter.NewForMounter(
			mounter.NewProxyMounter(socketPath, ns.rawMounter),
		)
```

pkg/mounter/proxy/server/ossfs/driver.go
```diff
@@ -29,18 +29,27 @@ func init() {
...
 }
 
 func NewDriver() *Driver {
-	return &Driver{
+	driver := &Driver{
 		pids:           new(sync.Map),
 		monitorManager: server.NewMountMonitorManager(),
-		raw:            mount.NewWithoutSystemd(""),
 	}
+	m := &extendedMounter{
```
AlbeeSo, line 43: it seems that `extendedMounter` is unnecessary.
```
	driver := &Driver{
		Mounter: mounter.NewForMounter(
			mount.NewWithoutSystemd(""),
			interceptors.OssfsSecretInterceptor,
			interceptors.OssfsMonitorInterceptor,
		),
		pids:           new(sync.Map),
		monitorManager: server.NewMountMonitorManager(),
	}
	return driver
```
And use `return h.Mounter.ExtendedMount(ctx, &mounter.MountOperation` in `Mount` instead?

pkg/mounter/interceptors/alinas_secret.go
iltyty: Test passed.

pkg/mounter/proxy/server/ossfs/driver.go
iltyty: This won't work, since driver.ExtendedMount won't be called, the mount process will be:
1. OssfsSecretInterceptor
2. OssfsMonitorInterceptor
3. mount.NewWithoutSystemd("").Mount()

pkg/mounter/interceptors/ossfs_monitor.go
```diff
@@ -0,0 +1,70 @@
...
+		klog.ErrorS(errors.New("failed to assert ossfs mount result type"), "skipping monitoring of mountpoint", "mountpoint", op.Target)
+		return err
+	}
+
+	go func() {
+		exitErr := <-res.ExitChan
+		// Assume the process exits with no error upon receiving SIGTERM,
+		// and exits with an error in case of unexpected failures.
+		if err == nil && exitErr != nil {
+			monitor.HandleMountFailureOrExit(err)
```
AlbeeSo: I think it's `monitor.HandleMountFailureOrExit(exitError)` here.

AlbeeSo:
lgtm, ptal @mowangdk

pkg/mounter/fuse_pod_manager/oss/registry.go
```diff
@@ -2,13 +2,15 @@ package oss
...
 	corev1 "k8s.io/api/core/v1"
 	"k8s.io/client-go/kubernetes"
 )
 
 var (
-	fstypeToFactory = map[string]func(*corev1.ConfigMap, metadata.MetadataProvider) OSSFuseMounterType{}
-	fstypeToPath    = map[string]string{}
+	fstypeToFactory      = map[string]func(*corev1.ConfigMap, metadata.MetadataProvider) OSSFuseMounterType{}
+	fstypeToPath         = map[string]string{}
+	fstypeToInterceptors = map[string][]mounter.MountInterceptor{}
```
mowangdk, line 13: Are you certain there won't be any concurrent reading situations?

pkg/mounter/cmd_mounter.go
```diff
@@ -7,41 +7,32 @@ import (
...
 		execPath:  execPath,
-		volumeId:  volumeId,
+		volumeID:  volumeID,
 		Interface: inner,
 	}
 }
 
-func (m *OssCmdMounter) ExtendedMount(source, target, fstype string, options []string, params *ExtendedMountParams) error {
+func (m *OssCmdMounter) ExtendedMount(_ context.Context, op *MountOperation) error {
 	ctx, cancel := context.WithDeadline(context.Background(), time.Now().Add(timeout))
```
mowangdk: Why not utilizing the context passed to this function

pkg/mounter/interceptors/alinas_secret.go
```diff
@@ -9,33 +10,29 @@ import (
...
 	if err != nil {
-		return op, err
+		return err
 	}
 
 	credFileContent := makeCredFileContent(op.Secrets)
 	credFilePath := path.Join(tmpDir, op.VolumeID+".credentials")
 	if err = os.WriteFile(credFilePath, credFileContent, 0o600); err != nil {
-		return op, err
+		os.RemoveAll(tmpDir) // cleanup in case of error
```
mowangdk: We need to avoid RemoveAll function call in our repo

pkg/mounter/interceptors/ossfs_secret.go
```diff
@@ -0,0 +1,64 @@
...
+		return handler(ctx, op)
+	}
+
+	passwdFile, err := utils.SaveOssSecretsToFile(op.Secrets, op.FsType)
+	if err != nil {
+		return err
+	}
+	if passwdFile != "" {
+		if fuseType == "ossfs" {
+			op.Args = append(op.Args, "passwd_file="+passwdFile)
```
AlbeeSo: I guess it should be op.Options for passed_file?

pkg/mounter/fuse_pod_manager/oss/registry.go
iltyty: Write operations only occur in the init function. At runtime, only concurrent reads happen.

mowangdk: Yeah, I mean we must avoid concurrent reads too

PR #1582, opened 2025-12-10
disk: allow unpublish pci-pf-stub driver

huww98:
#### What type of PR is this?

/kind bug

Fixes #

```release-note
Fix failing NodeUnpublishVolume for runD pods if the disk is already detached from controller
```

```docs

```


mowangdk:
I believe this may not be sufficient. Could it affect the disk newly attached to this node?

huww98:
This code is in `umountRunDVolumes`, which is only called in unpublish.

PR #1584, opened 2025-12-12
refactor(oss): extract runtime type determination logic

AlbeeSo:
Refactor NodePublishVolume to use DetermineRuntimeType function for better code clarity. The function implements a decision tree based on directAssigned, socketPath, and skipAttach to determine runtime type (COCO, RunD, RunC, ECI).

Changes:
- Add DetermineRuntimeType function with decision tree implementation
- Add RuntimeType enum and comprehensive unit tests
- Refactor NodePublishVolume to use runtime type instead of boolean checks
- Improve error handling by catching invalid combinations early

Maintains 100% logic equivalence with improved readability.

Fixes #

```release-note

```

```docs

```

### DIFF
--- a/pkg/oss/utils.go
+++ b/pkg/oss/utils.go
@@ -478,17 +478,62 @@ const (
 	RuntimeTypeUnknown RuntimeType = "unknown"
 )
 
+// runtimeTypeResult represents the result of runtime type determination
+type runtimeTypeResult struct {
+	runtimeType RuntimeType
+	err         error
+}
+
+// runtimeTypeLookupTable is a lookup table for all 8 possible combinations of
+// (directAssigned, hasSocketPath, skipAttach).
+// Index calculation: index = (directAssigned ? 4 : 0) + (hasSocketPath ? 2 : 0) + (skipAttach ? 1 : 0)å
+var runtimeTypeLookupTable = [8]runtimeTypeResult{
+	// Index 0: directAssigned=false, hasSocketPath=false, skipAttach=false -> error (should not occur)
+	{
+		runtimeType: RuntimeTypeUnknown,
+		err:         fmt.Errorf("socket path is empty in non csi-agent mode (should not occur)"),
+	},
+	// Index 1: directAssigned=false, hasSocketPath=false, skipAttach=true -> MicroVM
+	{
+		runtimeType: RuntimeTypeMicroVM,
+		err:         nil,
+	},
+	// Index 2: directAssigned=false, hasSocketPath=true, skipAttach=false -> RunC
+	{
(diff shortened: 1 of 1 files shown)


pkg/oss/utils.go
```diff
@@ -467,6 +467,85 @@ func parseDirectAssigned(runtimeClassValue, directAssignedValue string) bool {
...
+//   - directAssigned=true, socketPath=false, skipAttach=false: COCO
+//   - directAssigned=false, socketPath=true, skipAttach=true: RunD (pure rund cluster, no need to specify directAssigned)
+//   - directAssigned=false, socketPath=true, skipAttach=false: RunC
+//   - directAssigned=false, socketPath=false, skipAttach=true: ECI
+//   - directAssigned=false, socketPath=false, skipAttach=false: error (should not occur)
+//
+// Returns:
+//   - RuntimeType: the determined runtime type
+//   - error: if the combination is invalid (should not occur scenarios)
+func DetermineRuntimeType(directAssigned bool, socketPath string, skipAttach bool) (RuntimeType, error) {
```
huww98, line 541: I don't like this change. We don't actually care about the name of each runtime. We should have an abstract level to extract what is really different between these runtimes, and write code based on those differences. We can add comments about typical runtime, but don't depends on the runtime name in code.

AlbeeSo: The actual difference is 'skipAttach, directAssigned amd socketPath'. I write down the meaning of them but It's still difficult to manage the code if we don't name the runtime type clearly.

huww98: I think `skipAttach` should be renamed to something like `agentMode`. Most branches should already be self-explanatory.

The only difficult case is
```
	// Note: In non-csi-agent environment (!ns.skipAttach),
	//   if DirectAssigned is True, it's a confidential container scenario (coco)
	if opts.DirectAssigned && !ns.skipAttach {
		return ns.publishDirectVolume(ctx, req, opts)
	}
```
and we already have comments about it.

Particularly, `socketPath == ""` is a lot better than `runtimeType == RuntimeTypeMicroVM`, because the former clearly tells us we don't have mount proxy for this case.

Similarly, `ns.agentMode` (renamed from `skipAttach`) should be better than `runtimeType == RuntimeTypeRunD || runtimeType == RuntimeTypeMicroVM`. When run as agent, we don't need to care about globalmount.

AlbeeSo: If the logic is more difficult? It's hard to manage the code. I dont want to spend minute to consider every if-else.

mowangdk:
We can simplify the logic by potentially removing coco and using a sidecar instead 

/lgtm
/approve

PR #1585, opened 2025-12-15
Support nic lingjun node type

mowangdk:
Support nic type lingjun node

Fixes #

```release-note

```

```docs

```

### DIFF
--- a/pkg/cloud/metadata/lingjun.go
+++ b/pkg/cloud/metadata/lingjun.go
@@ -1,0 +1,50 @@
+package metadata
+
+import (
+	"encoding/json"
+	"errors"
+	"os"
+
+	"k8s.io/klog/v2"
+)
+
+type LingjunMetaData struct {
+	RegionId     string `json:"RegionId"`
+	ZoneId       string `json:"ZoneId"`
+	NodeId       string `json:"NodeId"`
+	InstanceType string `json:"InstanceType"`
+}
+
+func NewLingJunMetadata(lingjunConfigFile string) (*LingjunMetaData, error) {
+	lm := LingjunMetaData{}
+	data, err := os.ReadFile(lingjunConfigFile)
+	if err != nil {
+		if errors.Is(err, os.ErrNotExist) {
+			klog.V(3).InfoS("lingjun metadata file does not exist, use env metadata only", "file", lingjunConfigFile)
+		} else {
+			klog.ErrorS(err, "Failed to read lingjun metadatafile", "file", lingjunConfigFile)
+		}
+		return nil, err
+	}
+	err = json.Unmarshal(data, &lm)
+	if err != nil {
+		return nil, err
+	}
+	return &lm, nil
+}
+
+func (m *LingjunMetaData) Get(key MetadataKey) (string, error) {
+
+	switch key.String() {
+	case "RegionID":
+		return m.RegionId, nil
+	case "ZoneID":
+		return m.ZoneId, nil
+	case "InstanceID":
+		return m.NodeId, nil
+	case "InstanceType":
+		return m.InstanceType, nil
+	default:
+		return "", ErrUnknownMetadataKey
+	}
+}
--- a/pkg/cloud/metadata/lingjun_test.go
+++ b/pkg/cloud/metadata/lingjun_test.go
@@ -1,0 +1,200 @@
+package metadata
+
+import (
+	"os"
+	"path/filepath"
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+// LingjunTestCases contains multiple test cases for Lingjun metadata
+var LingjunTestCases = map[string]string{
+	"standard": `{
+	"RegionId": "cn-hangzhou",
+	"ZoneId": "cn-hangzhou-a",
+	"NodeId": "i-1234567890abcdef0",
+	"InstanceType": "ecs.g6.large",
+	"EniMac": "00:11:22:33:44:55",
+	"AckNicName": "eth1",
+	"NimitzNetworkMode": "VPC",
+	"NimitzInterface": ["eth0", "eth1"]
+}`,
+	"minimal": `{
+	"RegionId": "cn-beijing",
+	"ZoneId": "cn-beijing-c",
+	"NodeId": "i-0987654321fedcba0",
+	"InstanceType": "ecs.c5.large"
+}`,
+	"extended": `{
+	"RegionId": "cn-shanghai",
+	"ZoneId": "cn-shanghai-b",
+	"NodeId": "i-1122334455aabbccdd",
+	"InstanceType": "ecs.r5.xlarge",
+	"EniMac": "aa:bb:cc:dd:ee:ff",
+	"AckNicName": "eth0",
+	"NimitzNetworkMode": "ENI",
+	"NimitzInterface": ["eth0", "eth1", "eth2"]
+}`,
+}
+
+func TestNewLingJunMetadata(t *testing.T) {
+	t.Run("successful parsing - standard", func(t *testing.T) {
+		// Setup test file
+		tempDir := t.TempDir()
+		lingjunConfigFile := filepath.Join(tempDir, "lingjun_config_standard")
+
+		err := os.WriteFile(lingjunConfigFile, []byte(LingjunTestCases["standard"]), 0644)
+		require.NoError(t, err)
+
+		// Test the function
+		result, err := NewLingJunMetadata(lingjunConfigFile)
+		require.NoError(t, err)
+		require.NotNil(t, result)
+
+		assert.Equal(t, "cn-hangzhou", result.RegionId)
+		assert.Equal(t, "cn-hangzhou-a", result.ZoneId)
+		assert.Equal(t, "i-1234567890abcdef0", result.NodeId)
+		assert.Equal(t, "ecs.g6.large", result.InstanceType)
+	})
+
+	t.Run("successful parsing - minimal", func(t *testing.T) {
+		// Setup test file
+		tempDir := t.TempDir()
+		lingjunConfigFile := filepath.Join(tempDir, "lingjun_config_minimal")
+
+		err := os.WriteFile(lingjunConfigFile, []byte(LingjunTestCases["minimal"]), 0644)
+		require.NoError(t, err)
+
+		// Test the function
+		result, err := NewLingJunMetadata(lingjunConfigFile)
+		require.NoError(t, err)
+		require.NotNil(t, result)
+
+		assert.Equal(t, "cn-beijing", result.RegionId)
+		assert.Equal(t, "cn-beijing-c", result.ZoneId)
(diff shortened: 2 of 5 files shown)


pkg/cloud/metadata/lingjun.go
```diff
@@ -0,0 +1,51 @@
...
+			return m.RegionId, nil
+		case "ZoneId":
+			return m.ZoneId, nil
+		case "NodeId":
+			return m.NodeId, nil
+		case "InstanceType":
+			return m.InstanceType, nil
+		}
+	}
+	return "", ErrUnknownMetadataKey
```
huww98:
```suggestion
		switch v {
		case "RegionId":
			return m.RegionId, nil
		case "ZoneId":
			return m.ZoneId, nil
		case "NodeId":
			return m.NodeId, nil
		case "InstanceType":
			return m.InstanceType, nil
		default:
			return "", ErrUnknownMetadataKey
		}
```
And remove `LingjunConfig` completely.

pkg/cloud/metadata/lingjun_test.go
```diff
@@ -0,0 +1,121 @@
...
+			ZoneId:            "cn-hangzhou-a",
+			NodeId:            "i-1234567890abcdef0",
+			InstanceType:      "ecs.g6.large",
+			EniMac:            "00:11:22:33:44:55",
+			AckNicName:        "eth1",
+			NimitzNetworkMode: "VPC",
+			NimitzInterface:   []string{"eth0", "eth1"},
+		}
+
+		data, err := json.Marshal(testData)
```
huww98: I think we should copy a string from real LingJun node. Or we are just testing Go stdlib.

pkg/cloud/metadata/lingjun_test.go
```diff
@@ -0,0 +1,121 @@
...
+		require.NotNil(t, result)
+
+		assert.Equal(t, testData.RegionId, result.RegionId)
+		assert.Equal(t, testData.ZoneId, result.ZoneId)
+		assert.Equal(t, testData.NodeId, result.NodeId)
+		assert.Equal(t, testData.InstanceType, result.InstanceType)
+		assert.Equal(t, testData.EniMac, result.EniMac)
+		assert.Equal(t, testData.AckNicName, result.AckNicName)
+		assert.Equal(t, testData.NimitzNetworkMode, result.NimitzNetworkMode)
+		assert.Equal(t, testData.NimitzInterface, result.NimitzInterface)
```
huww98:
```suggestion
		assert.Equal(t, testData, result)
```

pkg/cloud/metadata/metadata.go
```diff
@@ -136,11 +136,24 @@ func newImmutableProvider(provider MetadataProvider, name string) *immutableProv
...
 func NewMetadata() *Metadata {
-	return &Metadata{
+	defaultMetadata := &Metadata{
 		providers: []MetadataProvider{
 			newImmutableProvider(&ENVMetadata{}, "env"),
 		},
 	}
+	data, err := os.ReadFile(LingjunConfigFile)
+	if err != nil {
+		klog.V(1).InfoS("can't read lingjun metadata file, use env metadata only", "file", LingjunConfigFile, "error", err)
```
huww98: Maybe we should log "no such file" error at V(3), and other errors with klog.Error

pkg/cloud/metadata/lingjun_test.go
```diff
@@ -0,0 +1,181 @@
...
+	"NimitzInterface": ["eth0", "eth1", "eth2"]
+}`,
+}
+
+// getLingjunMetadata returns the metadata JSON string for a given test case name
+func getLingjunMetadata(caseName string) []byte {
+	if data, exists := LingjunTestCases[caseName]; exists {
+		return []byte(data)
+	}
+	// Return default case if not found
```
huww98: Why so complex.. Just don't make it not found.

pkg/cloud/metadata/lingjun.go
```diff
@@ -0,0 +1,43 @@
...
+	ZoneId            string   `json:"ZoneId"`
+	NodeId            string   `json:"NodeId"`
+	InstanceType      string   `json:"InstanceType"`
+	EniMac            string   `json:"EniMac"`
+	AckNicName        string   `json:"AckNicName"`
+	NimitzNetworkMode string   `json:"NimitzNetworkMode"`
+	NimitzInterface   []string `json:"NimitzInterface"`
+}
+
+func NewLingJunMetadata(data []byte) (*LingjunMetaData, error) {
```
huww98: Can we hide all the error handling in this func? e.g. just pass in a path, log and return nil on error. Then we just need this in `NewMetadata`:
```go
if lm := NewLingJunMetadata(path); lm != nil {
	defaultMetadata.providers = append(defaultMetadata.providers, newImmutableProvider(lm, "lingjun"))
}
```
So that lingjun.go is more self-contained.

pkg/cloud/metadata/lingjun_test.go
```diff
@@ -0,0 +1,220 @@
...
+	"EniMac": "aa:bb:cc:dd:ee:ff",
+	"AckNicName": "eth0",
+	"NimitzNetworkMode": "ENI",
+	"NimitzInterface": ["eth0", "eth1", "eth2"]
+}`,
+}
+
+func TestNewLingJunMetadata(t *testing.T) {
+	// Create a temporary directory for test files
+	tempDir := t.TempDir()
```
huww98: Move this to each test case, to avoid any possibility of interference between different cases.

pkg/cloud/metadata/lingjun_test.go
```diff
@@ -0,0 +1,220 @@
...
+	originalConfigFile := LingjunConfigFile
+
+	t.Run("successful parsing - standard", func(t *testing.T) {
+		// Setup test file
+		LingjunConfigFile = filepath.Join(tempDir, "lingjun_config_standard")
+		defer func() {
+			LingjunConfigFile = originalConfigFile
+		}()
```
huww98: How about just pass `LingjunConfigFile` into `NewLingJunMetadata`? Replacing global variable should only be used as last resort, as it prevents parallel test.

pkg/cloud/metadata/metadata.go
```diff
@@ -136,11 +136,24 @@ func newImmutableProvider(provider MetadataProvider, name string) *immutableProv
...
 	}
+	lm, err := NewLingJunMetadata()
+	if err != nil {
+		if errors.Is(err, os.ErrNotExist) {
+			klog.V(3).InfoS("lingjun metadata file does not exist, use env metadata only", "file", LingjunConfigFile)
+		} else {
+			klog.ErrorS(err, "Failed to read lingjun metadatafile", "file", LingjunConfigFile)
+		}
+		return defaultMetadata
+	}
```
huww98, line 149: OK, but I'd prefer these code to also go into `NewLingJunMetadata`, so that we can have test coverage over it.

pkg/disk/utils.go
```diff
@@ -1177,7 +1177,12 @@ func getAvailableDiskCount(node *v1.Node, ecsClient cloud.ECSInterface, efloC cl
 		return count, nil
 	}
 	if lingjunNodeId != "" {
-		return getLingjunAvailableDiskCount(efloC, lingjunNodeId)
+		count, err := getLingjunAvailableDiskCount(efloC, lingjunNodeId)
+		if err != nil {
+			klog.ErrorS(err, "getAvailableDiskCount: failed to get lingjun available disk count", "lingjunID", lingjunNodeId)
```
huww98: Do we want to ignore all errors? Can we be more specific here?

pkg/disk/utils_test.go
```diff
@@ -1222,6 +1222,39 @@ func TestGetNodeTypeFromEFLOOpenAPI(t *testing.T) {
...
+					Return(resp, nil)
+			},
+			expectedType: "",
+			expectError:  true,
+		},
+		{
+			name:      "empty_node_type",
+			lingjunID: "node-123",
+			setupMock: func() {
+				// 模拟返回的 NodeType 为空字符串
```
huww98: Not sure about this case. Do you expect it to return success?

huww98: Also, please avoid Chinese in this repo.

pkg/cloud/metadata/lingjun.go
```diff
@@ -0,0 +1,48 @@
...
+
+type LingjunMetaData struct {
+	RegionId          string   `json:"RegionId"`
+	ZoneId            string   `json:"ZoneId"`
+	NodeId            string   `json:"NodeId"`
+	InstanceType      string   `json:"InstanceType"`
+	EniMac            string   `json:"EniMac"`
+	AckNicName        string   `json:"AckNicName"`
+	NimitzNetworkMode string   `json:"NimitzNetworkMode"`
+	NimitzInterface   []string `json:"NimitzInterface"`
```
huww98: How about remove fields that we don't use? This way, we:
* reduce the possibility of being breaked by changes of this file.
* slightly improve the performance as we don't need to deserialize these extra fields.

pkg/disk/utils_test.go
```diff
@@ -1096,12 +1181,116 @@ func TestGetLingjunNodeID(t *testing.T) {
...
+	ctrl := gomock.NewController(t)
+	defer ctrl.Finish()
+
+	mockEFLOClient := cloud.NewMockEFLOInterface(ctrl)
+
+	// 为成功的场景准备mock数据
+	nodeType := "ebmgvn"
+	lingjunNodeResp := &eflo.DescribeNodeResponse{
+		Body: &eflo.DescribeNodeResponseBody{
+			NodeType: &nodeType,
```
```suggestion
			NodeType: ptr.To("ebmgvn"),
```
Also avoid Chinese.

pkg/disk/utils_test.go
```diff
@@ -1096,12 +1181,116 @@ func TestGetLingjunNodeID(t *testing.T) {
 	}
 }
 
+func TestGetLingjunAvailableDiskCount(t *testing.T) {
+	ctrl := gomock.NewController(t)
+	defer ctrl.Finish()
+
+	mockEFLOClient := cloud.NewMockEFLOInterface(ctrl)
```
huww98: We'd better create a new `ctrl`/`mockEFLOClient` for each case, to avoid any interference.

Besides, `Finish()` is not nessesary:
```
New in go1.14+, if you are passing a *testing.T into NewController function you no longer need to call ctrl.Finish() in your test methods.
```

pkg/bmcpfs/controllerserver_test.go
```diff
@@ -48,6 +48,30 @@ func TestParseVolumeHandle(t *testing.T) {
...
+			volumeHandle:   "+fs-12345",
+			expectedFsId:   "",
+			expectedFsetId: "fs-12345",
+		},
+		{
+			name:           "delimiter at end",
+			volumeHandle:   "fs-12345+",
+			expectedFsId:   "fs-12345",
+			expectedFsetId: "",
+		},
```
huww98: OK, but I'm not sure if you add this by mistake? This is unrelated to other changes.

PR #1586, opened 2025-12-16
build: fix libblkid install path

huww98:
#### What type of PR is this?

/kind bug

Currently, we install libblkid in /usr/local/lib, this path is configured in
/etc/ld.so.conf.d/libc.conf, in package libc-bin, which is not included in
distroless images. So we are actually still linking to the old version of
libblkid.

Viewing the search path by /lib/aarch64-linux-gnu/ld-linux-aarch64.so.1 --help:

```
Shared library search path:
  (libraries located via /etc/ld.so.cache)
  /lib/aarch64-linux-gnu (system search path)
  /usr/lib/aarch64-linux-gnu (system search path)
  /lib (system search path)
  /usr/lib (system search path)
```

We don't have /etc/ld.so.cache which is created by ldconfig, also in libc-bin.
And /usr is going to be merged for debian. So our only choice is overwriting
the file installed by apt in /usr/lib/aarch64-linux-gnu.

Also refined the install script to exclude more unnecessary files.

Fixes #

```release-note
Fix rare false detection of file system on new encrypted disk.
```

```docs

```

### DIFF
--- a/build/multi/Dockerfile.multi
+++ b/build/multi/Dockerfile.multi
@@ -76,11 +76,13 @@ cd /src
 SOURCE_DATE_EPOCH=$(stat -c %Y /src.tar.xz)
 export SOURCE_DATE_EPOCH
 echo "util-linux released at $(date --date "@$SOURCE_DATE_EPOCH" --iso-8601=seconds)"
-./configure --disable-all-programs --enable-blkid --enable-libblkid --prefix=/usr/local \
+./configure --disable-all-programs --enable-blkid --enable-libblkid --libdir=\${prefix}/lib/$HOST \
     --disable-nls --disable-bash-completion --disable-asciidoc --disable-dependency-tracking --disable-static --host=$HOST
 make -j
-make install-strip DESTDIR=/out
-cd /out/usr/local && rm -r include share lib/pkgconfig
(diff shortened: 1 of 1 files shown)


huww98:
/hold
for verify

huww98:
/unhold
```
# xxd -a ./hfsplus.img 
00000000: 0000 0000 0000 0000 0000 0000 0000 0000  ................
*
00000400: 482b 0000 0000 0000 0000 0000 0000 0000  H+..............
00000410: 0000 0000 0000 0000 0000 0000 0000 0000  ................
00000420: 0000 0000 0000 0000 0000 0201 0000 0000  ................
00000430: 0000 0000 0000 0000 0000 0000 0000 0000  ................
*
00000520: 0005 0000 0000 0000 0000 0000 0000 0000  ................
00000530: 0000 0000 0000 0000 0000 0000 0000 0000  ................
*
00001ff0: 0000 0000 0000 0000 0000 0000 0000 0000  ................
# cp ./hfsplus.img /proc/545966/root/
# nsenter -t 545966 -m -- /sbin/blkid /hfsplus.img
# echo $?
2
```

old version will output `/hfsplus.img: BLOCK_SIZE="513" TYPE="hfsplus"`, which is false positive.

PR #1589, opened 2025-12-18
helm: allow to skip installing volumesnapshotclass

AlbeeSo:

Fixes #

```release-note

```

```docs

```


huww98:
CRDs cannot use template. And I think this is not logical, `volumeSnapshot.controller.enabled` is intended for someone who have deployed it somewhere else, so that we don't need to deploy it with our controller. It does not means turning off volumeSnapshot function.

AlbeeSo:
> CRDs cannot use template. And I think this is not logical, `volumeSnapshot.controller.enabled` is intended for someone who have deployed it somewhere else, so that we don't need to deploy it with our controller. It does not means turning off volumeSnapshot function.

We need to have way to skip create snapshotclass as we fix it as v1 and will fail in older cluster.

huww98:
lgtm, please also update the commit message.

huww98:
> We need to have way to skip create snapshotclass as we fix it as v1 and will fail in older cluster.

Then admin should update CRDs manually. It is not breaking for older API version.

PR #1591, opened 2025-12-24
oss(credentials): support STS.Token in nodePublishSecretRef and rotation by republish

AlbeeSo:
#### What type of PR is this?
/kind feature

1. recreate OSS csidriver and set `requiredRepublish` to true.
2. create secret like
`kubectl create -n default secret generic oss-secret --from-literal='accessKeyId=<yourAccessKeyFromAssumeRole>' --from-literal='accessKeySecret=<yourAccessKeySecretFromAssumeRole>' --from-literal='securityToken=<yourSecurityTokenFromAssumeRole>' --from-literal='Expiration=2025-12-22T04:11:50Z'`
3. create a OSS PV with nodePublishSecretRef with the secret created.
4. rotate the credentials in secret (remain more than 20 minutes expiration)
5. check the OSS PV can still access server, and the audit logs of OSS shows the new credentials are used. 

Fixes #

```release-note

```

```docs

```

### DIFF
--- a/pkg/mounter/interceptors/ossfs_secret.go
+++ b/pkg/mounter/interceptors/ossfs_secret.go
@@ -148,12 +148,26 @@ func rotateTokenFiles(dir string, secrets map[string]string) (rotated bool, err 
 	// Check if any file needs update before writing
 	// Read current data directory if symlink exists
 	currentDataDir := ""
-	if linkTarget, readErr := os.Readlink(dir); readErr == nil {
+	linkTarget, err := os.Readlink(dir)
+	if err == nil {
 		// dir is a symlink, resolve to actual directory
 		currentDataDir = resolveSymlinkTarget(dir, linkTarget)
+	} else if os.IsNotExist(err) {
+		// dir doesn't exist, this is the first call
+		// currentDataDir remains empty
+	} else {
+		// Error reading symlink - check if dir is a regular directory
+		// If it's a regular directory, we cannot safely convert it to symlink because
+		// rename would change the inode, breaking file handles that clients may have opened.
+		// We must return an error to prevent data inconsistency.
+		fileInfo, statErr := os.Stat(dir)
+		if statErr == nil && fileInfo.IsDir() {
+			// dir is a regular directory, cannot safely rotate
+			return false, fmt.Errorf("cannot rotate token files: %s is a regular directory, not a symlink. ", dir)
+		}
+		// Some other error (permission denied, etc.) - cannot proceed with rotation
+		return false, fmt.Errorf("failed to read symlink %s: %w. Cannot proceed with token rotation", dir, err)
 	}
-	// Note: dir should always be a symlink path that doesn't exist on first call,
-	// so we don't need to handle the case where dir is a regular directory
 
 	anyNeedsUpdate := false
 	if currentDataDir == "" {
--- a/pkg/mounter/interceptors/ossfs_secret_test.go
+++ b/pkg/mounter/interceptors/ossfs_secret_test.go
@@ -14,6 +14,7 @@ import (
 	mounterutils "github.com/kubernetes-sigs/alibaba-cloud-csi-driver/pkg/mounter/utils"
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
+	"golang.org/x/sys/unix"
 	"k8s.io/klog/v2"
 	mountutils "k8s.io/mount-utils"
 )
@@ -800,6 +801,62 @@ func TestRotateTokenFiles(t *testing.T) {
 			}
 		})
 	}
+
+	// Test regular directory should return error
+	t.Run("regular directory should return error", func(t *testing.T) {
+		tmpDir := t.TempDir()
+		parentDir := filepath.Join(tmpDir, "parent")
+		err := os.MkdirAll(parentDir, 0o755)
+		require.NoError(t, err)
+
+		// Create a regular directory (not a symlink) with token files
+		regularDir := filepath.Join(parentDir, "sts")
+		err = os.MkdirAll(regularDir, 0o755)
+		require.NoError(t, err)
+
+		// Create token files in the regular directory
+		secrets := map[string]string{
+			mounterutils.KeyAccessKeyId:     "oldAKID",
+			mounterutils.KeyAccessKeySecret: "oldAKSecret",
+			mounterutils.KeySecurityToken:   "oldToken",
+		}
+		for key, value := range secrets {
+			filePath := filepath.Join(regularDir, key)
+			err := os.WriteFile(filePath, []byte(value), 0o600)
+			require.NoError(t, err)
+		}
+
+		// Verify it's a regular directory, not a symlink
+		fileInfo, err := os.Stat(regularDir)
+		require.NoError(t, err)
+		assert.True(t, fileInfo.IsDir())
+		_, err = os.Readlink(regularDir)
+		assert.Error(t, err) // Should fail because it's not a symlink
+
+		// Call rotateTokenFiles with new values - should return error
+		newSecrets := map[string]string{
+			mounterutils.KeyAccessKeyId:     "newAKID",
+			mounterutils.KeyAccessKeySecret: "newAKSecret",
+			mounterutils.KeySecurityToken:   "newToken",
+		}
+		rotated, err := rotateTokenFiles(regularDir, newSecrets)
+		assert.Error(t, err, "should return error for regular directory")
+		assert.False(t, rotated, "should not have rotated")
+		assert.Contains(t, err.Error(), "regular directory")
+		assert.Contains(t, err.Error(), "not a symlink")
+
+		// Verify regularDir is still a regular directory (not converted)
+		fileInfo, err = os.Stat(regularDir)
+		require.NoError(t, err)
+		assert.True(t, fileInfo.IsDir())
+		_, err = os.Readlink(regularDir)
+		assert.Error(t, err) // Still not a symlink
+
+		// Verify old files are still there (not rotated)
+		oldAK, err := os.ReadFile(filepath.Join(regularDir, mounterutils.KeyAccessKeyId))
+		require.NoError(t, err)
+		assert.Equal(t, "oldAKID", string(oldAK))
+	})
 }
 
 // TestOssfsSecretInterceptor_TokenRotation tests token rotation scenarios.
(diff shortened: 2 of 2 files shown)


pkg/mounter/interceptors/ossfs_secret.go
```diff
@@ -62,3 +118,107 @@ func ossfsSecretInterceptor(ctx context.Context, op *mounter.MountOperation, han
...
+		rotated = fileUpdate || rotated
+	}
+	return
+}
+
+// prepareCredentialFiles returns:
+//  1. file:    path of ossfs credential file for fixed AKSK
+//  2. dir:     dorectory of ossfs credential files for token
+//  4. error
+func prepareCredentialFiles(fuseType, target string, secrets map[string]string) (file, dir string, err error) {
```
mowangdk, line 302: A separate dir value isn't needed if it can be included in the file string. Use another meaningful value if you want to distinguish aksk and token.

pkg/mounter/utils/filelock.go
```diff
@@ -0,0 +1,53 @@
...
+	fileLocks = sync.Map{}
+)
+
+func getFileLock(path string) *sync.RWMutex {
+	lock, _ := fileLocks.LoadOrStore(path, &sync.RWMutex{})
+	return lock.(*sync.RWMutex)
+}
+
+// WriteFileWithLock safely writes data to file with locking
+func WriteFileWithLock(path string, data []byte, perm os.FileMode) (done bool, err error) {
```
mowangdk: Does this lock secure the secret file update? If we write a new token to a separate file and then rename it, we wouldn't need locks?

pkg/mounter/interceptors/ossfs_secret.go
```diff
@@ -56,9 +124,205 @@ func ossfsSecretInterceptor(ctx context.Context, op *mounter.MountOperation, han
...
+		}()
+	}
+
+	rotated = true
+	return
+}
+
+// prepareCredentialFiles returns:
+//  1. file:    path of ossfs credential file for fixed AKSK
+//  2. dir:     dorectory of ossfs credential files for token
```
iltyty: "dorectory" -> "directory"
4. error -> 3. error

pkg/mounter/utils/filelock.go
AlbeeSo: ptal again.

pkg/mounter/interceptors/ossfs_secret.go
```diff
@@ -35,6 +35,55 @@ func removeIgnoreNotExist(path string) {
...
+		return false, nil
+	}
+
+	// Content is different or file doesn't exist, need to write
+	// Create temporary file in the same directory
+	dir := filepath.Dir(path)
+	tmpFile := filepath.Join(dir, fmt.Sprintf(".%s.tmp.%d", filepath.Base(path), time.Now().UnixNano()))
+
+	// Write to temporary file
+	if err = commonutils.WriteAndSyncFile(tmpFile, data, perm); err != nil {
```
huww98: I think we don't need to sync here. sync is slow. And if we really crash, we will always rewrite this file after reboot, right? So it is OK for this file to corrupt on crash.

pkg/mounter/interceptors/ossfs_secret.go
```diff
@@ -3,40 +3,157 @@ package interceptors
...
+		return false, err
+	}
+	if exists && sameContent {
+		return false, nil
+	}
+
+	// Content is different or file doesn't exist, need to write
+	// Create temporary file in the same directory
+	dir := filepath.Dir(path)
+	tmpFile := filepath.Join(dir, fmt.Sprintf(".%s.tmp.%d", filepath.Base(path), time.Now().UnixNano()))
```
huww98:
```suggestion
	tmpFile := path + ".next"
```
I think we can just use a fixed name? So that we will not worry about leaking the tmp files. And no need to clean it up when rename failed.

pkg/mounter/interceptors/ossfs_secret.go
```diff
@@ -3,40 +3,157 @@ package interceptors
...
+		klog.V(4).InfoS("created ossfs token directory", "dir", tokenDir)
+		if fuseType == mounterutils.OssFsType {
+			op.Options = append(op.Options, "passwd_file="+tokenDir)
+		} else {
+			// ossfs2
+			// For ossfs2, file-path is a common option configuration after -o, so append to op.Options
+			op.Options = append(op.Options,
+				fmt.Sprintf("oss_sts_multi_conf_ak_file=%s", filepath.Join(tokenDir, mounterutils.KeyAccessKeyId)),
+				fmt.Sprintf("oss_sts_multi_conf_sk_file=%s", filepath.Join(tokenDir, mounterutils.KeyAccessKeySecret)),
+				fmt.Sprintf("oss_sts_multi_conf_token_file=%s", filepath.Join(tokenDir, mounterutils.KeySecurityToken)),
```
huww98: If ossfs2 takes these 3 path seperately, it is garanteed to face race condition: ossfs opens one file, CSI replaces them, and ossfs opens other files, then ossfs will get inconsistent credential. Hopes ossfs2 will immediately retry in this case, and not cause IOError for user.

For ossfs1 to work correctly, it should first open a fd to the dir, them use `openat` API to open individual files relative to the fd. Can you check how ossfs1 implements this?

BTW, what's the point of this comment?
```
// For ossfs2, file-path is a common option configuration after -o, so append to op.Options
```
It is also an option after -o for OSSFS1, right?

pkg/mounter/interceptors/ossfs_secret.go
```diff
@@ -56,9 +173,215 @@ func ossfsSecretInterceptor(ctx context.Context, op *mounter.MountOperation, han
...
+func rotatePasswdFile(path string, data []byte, perm os.FileMode) (rotated bool, err error) {
+	return writeFileAtomically(path, data, perm)
+}
+
+// rotateTokenFiles rotates (or initializes) token files using symlink for atomic updates.
+// This function uses a symlink-based approach similar to Kubernetes configmap volume plugin:
+// 1. Create a temporary data directory (e.g., ..data_tmp_<timestamp>)
+// 2. Write all token files to the temporary directory
+// 3. Atomically switch the ..data symlink to point to the new directory
+// 4. Create symlinks for each token file pointing to ..data/<filename>
```
huww98: I think this can be simplified: If path is /run/ossfs/volumeID/sts, we can make:
```
/run/ossfs/
└── volumeID
    ├── sts -> sts.20260101000000
    └── sts.20260101000000
        ├── ak
        ├── expiration
        ├── sk
        └── token
```
i.e., we don't need a symlink for every file. kubelet did that because workload expect the files exist at the root of the volume, while we can make an extra level of dir in the volume.

pkg/mounter/interceptors/ossfs_secret.go
```diff
@@ -56,9 +173,215 @@ func ossfsSecretInterceptor(ctx context.Context, op *mounter.MountOperation, han
...
+			removeIgnoreNotExist(tmpFileLinkPath)
+			klog.Errorf("failed to atomically replace symlink for %s: %v", key, err)
+			// Continue with other files
+		}
+	}
+
+	// Clean up old data directory if it exists
+	// Only remove the token files we know about, not the entire directory
+	if currentDataDir != "" && currentDataDir != tmpDataDir {
+		// Remove old data directory asynchronously to avoid blocking
```
huww98: Not nessesary IMO. cleanup should be blocking, as we don't want to leak credentials on disk.

And I think we should keep `currentDataDir` for a while? Because ossfs may still holding a fd to the old dir and trying to read from it. Maybe holding two most recent dirs, and remove the first one before creating the third one?

pkg/mounter/interceptors/ossfs_secret.go
AlbeeSo: I prefer to use a defer deletion function to deal with this issue instead of using fixed names, it will bring many overwriting corner cases.

pkg/mounter/interceptors/ossfs_secret.go
AlbeeSo: ossfs1 uses `openat`.
And I modified the mentioned comment. In fact it means the comparison between the different credential methods of ossfs2.

pkg/mounter/interceptors/ossfs_secret.go
AlbeeSo: done, with a subdirectory passthrough to clients instead.

pkg/mounter/interceptors/ossfs_secret.go
AlbeeSo: with the last-updated accesskeyid to solve this issue, instead of holding the former dir.

pkg/mounter/interceptors/ossfs_secret.go
```diff
@@ -211,11 +180,17 @@ func rotateTokenFiles(dir string, secrets map[string]string) (rotated bool, err
 
 	// Check if any file needs update before writing
 	// Read current data directory if symlink exists
-	dataLinkPath := filepath.Join(dir, "..data")
 	currentDataDir := ""
-	if linkTarget, readErr := os.Readlink(dataLinkPath); readErr == nil {
-		currentDataDir = filepath.Join(dir, linkTarget)
+	if linkTarget, readErr := os.Readlink(dir); readErr == nil {
```
mowangdk: If there is readErr, we must know what error it is, even if it doesn't affect process. please add log for record

pkg/mounter/interceptors/ossfs_secret.go
```diff
@@ -56,9 +111,216 @@ func ossfsSecretInterceptor(ctx context.Context, op *mounter.MountOperation, han
...
+	defer removeIgnoreNotExist(tmpLinkPath)
+
+	// Atomically rename the temporary symlink to dir
+	// Note: dir should not exist on first call, so this will create the symlink
+	// On subsequent calls, this will atomically replace the existing symlink
+	// This is atomic on most filesystems and ensures all readers see either all old files or all new files
+	// IMPORTANT: Once this rename completes, all new opens of dir/ will resolve to the new directory.
+	// However, clients that have already opened file handles will continue to read from the old directory
+	// (pointed to by their open file descriptors), ensuring no interruption during rotation.
+	if err = os.Rename(tmpLinkPath, dir); err != nil {
```
mowangdk, line 245: I think ```os.Rename``` will throw an error if the directory already exists and won't overwrite it.

```
// Rename renames (moves) oldpath to newpath.
// If newpath already exists and is not a directory, Rename replaces it.
// If newpath already exists and is a directory, Rename returns an error.
// OS-specific restrictions may apply when oldpath and newpath are in different directories.
// Even within the same directory, on non-Unix platforms Rename is not an atomic operation.
// If there is an error, it will be of type *LinkError.
```

huww98: but target is a symlink in this case.

pkg/mounter/interceptors/ossfs_secret.go
AlbeeSo: In this whole function, it assumes that the `dir` is not exists (the first call for mounting), or a symlink (the sequential calls for rotating).  So here I mute the error logs, and do not consider the regular directory case for another comment u left below. 
The latest commit I return error for these corner cases, and solve the rename issue also. (But this will not happen I guess.)

PR #1592, opened 2025-12-25
refactor(nas): get node information only in losetup mode

iltyty:
#### What type of PR is this?
/kind cleanup

Get node information only in losetup mode, otherwise the nas driver will failed to start when no kubeconfig is provided.

Fixes #

```release-note

```

```docs

```

### DIFF
--- a/pkg/nas/internal/config.go
+++ b/pkg/nas/internal/config.go
@@ -126,12 +126,7 @@ func GetNodeConfig() (*NodeConfig, error) {
 		klog.Info("enabled nas volume stats")
 	}
 
-	// get node info
 	nodeName := os.Getenv("KUBE_NODE_NAME")
-	node, err := kubeClient.CoreV1().Nodes().Get(context.Background(), nodeName, metav1.GetOptions{})
-	if err != nil {
-		return nil, err
-	}
 	config.NodeName = nodeName
 
 	// check if losetup enabled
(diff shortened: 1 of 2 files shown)


pkg/nas/internal/config.go
```diff
@@ -126,20 +126,19 @@ func GetNodeConfig() (*NodeConfig, error) {
...
 		config.EnableLosetup, _ = parseBool(value)
 	}
 	if config.EnableLosetup {
 		klog.Info("enabled nas losetup mode")
+		nodeName := os.Getenv("KUBE_NODE_NAME")
+		node, err := kubeClient.CoreV1().Nodes().Get(context.Background(), nodeName, metav1.GetOptions{})
+		if err != nil {
+			return nil, err
+		}
+		config.NodeName = nodeName
```
huww98: `config.NodeName` is also used as CSI NodeID, So I think we still need this when `!config.EnableLosetup`

mowangdk: How about renaming sidecar mode's nodeName to podName, as I haven't find use case for nodeName in the nas driver.

huww98: CI failure also caused by this issue:

```
  I1226 18:57:46.163485   18655 main.go:109] "Received NotifyRegistrationStatus call" status="&RegistrationStatus{PluginRegistered:false,Error:RegisterPlugin error -- plugin registration failed with err: error adding CSI driver node info: driverNodeID must not be empty,}"
  E1226 18:57:46.163520   18655 main.go:111] "Registration process failed with error, restarting registration container" err="RegisterPlugin error -- plugin registration failed with err: error adding CSI driver node info: driverNodeID must not be empty"
```

huww98: > How about renaming sidecar mode's nodeName to podName

I think we can still use KUBE_NODE_NAME env name, and let caller to determine what value is appropriate.

PR #1607, opened 2026-01-04
refactor: make kube config optional in nas/oss node drivers

iltyty:
#### What type of PR is this?
/kind cleanup

Make kube config optional on oss and nas node servers, supporting reading relevant configurations from env variables.

Fixes #

```release-note

```

```docs

```

### DIFF
--- a/pkg/nas/internal/config.go
+++ b/pkg/nas/internal/config.go
@@ -12,8 +12,8 @@ import (
 	"github.com/kubernetes-sigs/alibaba-cloud-csi-driver/pkg/nas/cloud"
 	"github.com/kubernetes-sigs/alibaba-cloud-csi-driver/pkg/nas/interfaces"
 	"github.com/kubernetes-sigs/alibaba-cloud-csi-driver/pkg/options"
+	"github.com/kubernetes-sigs/alibaba-cloud-csi-driver/pkg/utils"
 	corev1 "k8s.io/api/core/v1"
-	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/client-go/dynamic"
 	"k8s.io/client-go/kubernetes"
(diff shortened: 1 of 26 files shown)


pkg/nas/internal/config.go
```diff
@@ -158,6 +167,18 @@ func GetNodeConfig() (*NodeConfig, error) {
 	return config, nil
 }
 
+func getValueFromEnvOrConfigMap(envKey, cmKey string, cm *corev1.ConfigMap) string {
```
huww98: Can you reuse `pkg/utils/config.go`? It is designed specifically for reading from configmap or env.

pkg/oss/oss.go
```diff
@@ -64,17 +65,29 @@ func NewDriver(endpoint string, m metadata.MetadataProvider, serviceType utils.S
 		nodeName = "controller" // any non-empty value to avoid csi-common panic
 	}
 
-	cfg := options.MustGetRestConfig()
-	clientset := kubernetes.NewForConfigOrDie(cfg)
-	crdCfg := options.GetRestConfigForCRD(*cfg)
-	cnfsGetter := cnfsv1beta1.NewCNFSGetter(dynamic.NewForConfigOrDie(crdCfg))
+	cfg, err := options.GetRestConfig()
```
AlbeeSo, line 64: would it be better if we add an ENV to declare the current environment?  Im afraid if the nil clients will cause any nil point panic or resources leak in the normal environment.

pkg/nas/nodeserver.go
```diff
@@ -55,7 +55,7 @@ type nodeServer struct {
 
 func newNodeServer(config *internal.NodeConfig) *nodeServer {
 	// configure efc dadi service discovery
-	if config.EnableEFCCache {
+	if config.EnableEFCCache && config.KubeClient != nil {
```
mowangdk: If kubeClient is nil with EFCCache enabled, an error must be thrown(but not crash), as checking kubeClient isn't necessary.

pkg/nas/internal/config.go
```diff
@@ -107,21 +119,20 @@ func GetNodeConfig() (*NodeConfig, error) {
 		config.EnablePortCheck, _ = parseBool(value)
 	}
 
-	// get csi-plugin configmap
-	cm, err := kubeClient.CoreV1().ConfigMaps(configMapNamespace).Get(context.Background(), configMapName, metav1.GetOptions{})
-	if err != nil {
-		if !apierrors.IsNotFound(err) {
+	var cm *corev1.ConfigMap
+	if kubeClient != nil {
```
mowangdk: Is there any way we can do to  minimize the nil judgments?

pkg/oss/oss.go
mowangdk: Adding additional ENV isn't the ideal solution we seek.

PR #1613, opened 2026-01-09
feat(nas): support access point rrsa authentication

iltyty:
#### What type of PR is this?
/kind feature

Support NAS access point RRSA authentication.

Fixes #

```release-note

```

```docs

```


pkg/mounter/interceptors/alinas_secret.go
```diff
@@ -2,56 +2,293 @@ package interceptors
...
+	return sts20150401.NewClient(stsConfig)
+}
+
+func (i *AlinasSecretInterceptor) Intercept(ctx context.Context, op *mounterutils.MountRequest, handler mounter.MountHandler) (err error) {
+	if op == nil || op.AuthConfig == nil {
 		return handler(ctx, op)
 	}
+	if op.AuthConfig.AuthType != rrsaAuthType && (op.AuthConfig.AccessKey == "" || op.AuthConfig.AccessSecret == "") {
+		return handler(ctx, op)
+	}
```
mowangdk, line 86: Can we combine these two if statements? There doesn't seem to be any difference between them

pkg/mounter/interceptors/alinas_secret.go
```diff
@@ -2,56 +2,293 @@ package interceptors
...
+		}
+	}
+
+	klog.V(4).InfoS("Created alinas credential file", "path", credFilePath)
+	op.Options = append(op.Options, "ram_config_file="+credFilePath)
+	if err = handler(ctx, op); err != nil {
+		return
+	}
+
+	if op.AuthConfig != nil && op.AuthConfig.AuthType == rrsaAuthType {
```
mowangdk, line 108: Can we move these statements into a switch block? Just emit an error if the credential file doesn't exist.

pkg/mounter/interceptors/alinas_secret.go
```diff
@@ -2,56 +2,293 @@ package interceptors
...
+	providerArn, roleArn := mounterutils.GetArn(provider, op.AuthConfig.AccountID, op.AuthConfig.RoleName)
+
+	req := &sts20150401.AssumeRoleWithOIDCRequest{
+		DurationSeconds: tea.Int64(stsTokenDurationSeconds),
+		OIDCToken:       tea.String(string(oidcToken)),
+		OIDCProviderArn: tea.String(providerArn),
+		RoleArn:         tea.String(roleArn),
+		RoleSessionName: tea.String(session),
+	}
+	resp, err := i.stsClient.AssumeRoleWithOIDC(req)
```
mowangdk, line 214: Will there be a scenario where a node doesn't have ```AssumeRoleWithOIDCRequests``` policys?

pkg/mounter/interceptors/alinas_secret.go
```diff
@@ -2,56 +2,293 @@ package interceptors
...
+	if err != nil || resp.Body == nil || resp.Body.Credentials == nil {
```
mowangdk, line 215: We should at least log the error, so that we can identify why request failed

mowangdk (commented):
@AlbeeSo Please take a look as well

mowangdk (commented):
## Automated Code Review

**Summary:** Adds RRSA (RAM Role for Service Account) authentication support for NAS access points, including OIDC token exchange via STS and mount-proxy integration for credential injection.

**Issues Found:**
- The `AssumeRoleWithOIDC` method was added to the `STSInterface` — ensure the mock implementation matches the real STS client signature exactly.
- The mounter refactoring (`MountOperation` → `MountRequest` from `mounter/utils`) is a good consolidation, but it touches multiple mounter implementations (`AdaptorMounter`, `OssCmdMounter`, `ConnectorMounter`). Ensure all mount paths still work correctly.
- The `alinas_secret.go` interceptor file shows a significant rewrite (from 56 lines of context visible). This is the security-sensitive path — ensure credential handling doesn't leak secrets in logs.
- `csi-agent/main.go` changes from `WrapNodeServerWithValidator` to `WrapNodeServer` — verify this doesn't remove important request validation.

**Suggestions:**
- The STS token exchange is a security-critical path. Add unit tests covering: successful OIDC exchange, expired token handling, invalid role ARN, and network failures.
- Document the end-to-end flow: ServiceAccount → projected token → STS AssumeRoleWithOIDC → NAS access credentials.

**Overall:** Significant feature enabling fine-grained NAS access control via RRSA. The mounter refactoring is clean. Security-critical paths need thorough test coverage.

mowangdk (commented):
## Automated Code Review

**Summary:** Adds NAS access point RRSA authentication support via OIDC token exchange, with a background refresh loop that keeps tokens fresh while the mount point is active.

**Overall:** Significant feature with good test coverage. Some concurrency and error handling concerns.

pkg/mounter/interceptors/alinas_secret.go
```diff
@@ -2,56 +2,293 @@ package interceptors
...
+	}
+
+	if op.AuthConfig != nil && op.AuthConfig.AuthType == rrsaAuthType {
+		go i.startTokenRefreshLoop(op)
+	}
+	return
+}
+
+func (i *AlinasSecretInterceptor) startTokenRefreshLoop(op *mounterutils.MountRequest) {
+	if op == nil || op.Target == "" || op.AuthConfig == nil {
```
mowangdk, line 115: The `startTokenRefreshLoop` goroutine runs forever with a 5-second ticker polling `isMountPoint`. This is a high polling frequency for a stat syscall. Consider using a longer interval (e.g., 30s-60s) since token refresh only needs to happen every ~5 minutes anyway, and the mount point check just gates the loop lifetime.

pkg/mounter/interceptors/alinas_secret.go
```diff
@@ -2,56 +2,293 @@ package interceptors
...
+			}
+		}
+	}
+
+	klog.V(4).InfoS("Created alinas credential file", "path", credFilePath)
+	op.Options = append(op.Options, "ram_config_file="+credFilePath)
+	if err = handler(ctx, op); err != nil {
+		return
+	}
+
```
mowangdk, line 107: `go i.startTokenRefreshLoop(op)` captures `op` by pointer. If the caller modifies `op` after `Intercept` returns, the goroutine will see the mutations. This is likely fine given the current calling pattern, but it's fragile. Consider copying the relevant fields (VolumeID, Target, AuthConfig) into the goroutine's closure.

pkg/mounter/interceptors/alinas_secret.go
```diff
@@ -2,56 +2,293 @@ package interceptors
...
+	if regionID == "" {
+		return
+	}
+
+	provider, err := credentials.NewProvider()
+	if err != nil {
+		return
+	}
+
+	cred := alicred.FromCredentialsProvider(provider.GetProviderName(), provider)
```
mowangdk, line 75: `newSTSClient` returns `(nil, nil)` when `regionID` is empty. Callers of `NewAlinasSecretInterceptor` will succeed, but later calls to `refreshRRSAToken` check `i.stsClient == nil` and just log an error without returning it clearly. The interceptor will silently fail token refresh. Consider returning an error from `NewAlinasSecretInterceptor` when regionID is empty but RRSA is expected, or make the nil-client path more explicit.

PR #1621, opened 2026-01-20
disk: introduce data cache

huww98:
#### What type of PR is this?

/kind feature

Use a faster disk as a cache layer over cloud disk.

Fixes #

```release-note
disk: support using a faster disk as a cache layer over cloud disk.
```

```docs

```

### DIFF
--- a/pkg/disk/csi_agent.go
+++ b/pkg/disk/csi_agent.go
@@ -111,5 +111,10 @@ func (a *CSIAgent) NodeUnpublishVolume(ctx context.Context, req *csi.NodeUnpubli
 	if err != nil {
 		return nil, status.Errorf(codes.Internal, "failed to cleanup %s: %v", req.TargetPath, err)
 	}
+	// NodePublishVolume set up the data cache (via setupDisk); tear it down
+	// here, its symmetric counterpart.
+	if err := datacache.Teardown(ctx, a.ns.dmControl, req.VolumeId); err != nil {
+		return nil, status.Errorf(codes.Internal, "teardown DataCache for %s: %v", req.VolumeId, err)
+	}
 	return &csi.NodeUnpublishVolumeResponse{}, nil
 }
--- a/pkg/disk/csi_agent.go
+++ b/pkg/disk/csi_agent.go
@@ -8,6 +8,7 @@ import (
 	"os"
 
 	"github.com/container-storage-interface/spec/lib/go/csi"
+	"github.com/kubernetes-sigs/alibaba-cloud-csi-driver/pkg/disk/datacache"
 	"github.com/kubernetes-sigs/alibaba-cloud-csi-driver/pkg/disk/mounter"
 	"github.com/kubernetes-sigs/alibaba-cloud-csi-driver/pkg/utils"
 	"google.golang.org/grpc/codes"
@@ -29,6 +30,14 @@ func NewCSIAgent() *CSIAgent {
 	}
 	GlobalConfigVar.MetricEnable = true
 
+	dmControl, err := datacache.OpenDmControl()
+	if err != nil {
+		klog.Fatalf("failed to open device-mapper control: %v", err)
+	}
+	if dmControl == nil {
+		klog.Info("device-mapper unavailable, data cache disabled")
+	}
+
 	return &CSIAgent{
 		ns: &nodeServer{
 			mounter:     utils.NewMounter(),
@@ -36,6 +45,7 @@ func NewCSIAgent() *CSIAgent {
 			unixMounter: mounter.UnixMounter{},
 			podCGroup:   podCgroup,
 			locks:       utils.NewVolumeLocks(),
+			dmControl:   dmControl,
 			ad: DiskAttachDetach{
 				dev:    DefaultDeviceManager,
 				devMap: &devMap{}, // Nobody will add to this map.
--- a/pkg/disk/data_cache.go
+++ b/pkg/disk/data_cache.go
@@ -25,13 +25,11 @@ const (
 type dataCache struct {
 	Size resource.Quantity
 	Mode DataCacheMode
-	Path string
 }
 
 const (
 	DataCacheModeKey = "dataCacheMode"
 	DataCacheSizeKey = "dataCacheSize"
-	DataCachePathKey = "dataCachePath"
 )
 
 func getDataCacheOpts(opts map[string]string, d *dataCache) error {
@@ -50,15 +48,7 @@ func getDataCacheOpts(opts map[string]string, d *dataCache) error {
 		return fmt.Errorf("unrecognized %s: %s", DataCacheModeKey, m)
 	}
 
-	d.Path = opts[DataCachePathKey]
-	if d.Path != "" && d.Path[0] != '/' {
-		return fmt.Errorf("%s must be an absolute path", DataCachePathKey)
-	}
-
-	if d.Mode != "" || d.Path != "" || !d.Size.IsZero() {
-		if d.Path == "" {
-			return fmt.Errorf("must specify %s for dataCache", DataCachePathKey)
-		}
+	if d.Mode != "" || !d.Size.IsZero() {
 		if d.Size.IsZero() {
 			return fmt.Errorf("must specify non-zero %s for dataCache", DataCacheSizeKey)
 		}
@@ -251,9 +241,11 @@ func resizeDmCache(logger klog.Logger, size uint64, volumeID string) error {
 	return updateTable(dmCtrl, volumeID, size, args)
 }
 
-func cacheFilePath(base, volumeID string) (meta, data string) {
-	meta = filepath.Join(base, volumeID+".meta")
-	data = filepath.Join(base, volumeID+".data")
+const DataCachePath = "/var/alibaba-cloud-csi/data-cache"
+
+func cacheFilePath(volumeID string) (meta, data string) {
+	meta = filepath.Join(DataCachePath, volumeID+".meta")
+	data = filepath.Join(DataCachePath, volumeID+".data")
 	return meta, data
 }
 
@@ -279,7 +271,7 @@ func setupDataCache(logger klog.Logger, d *dataCache, device, volumeID string) (
 	}
 
 	size := d.Size.Value()
-	meta, data := cacheFilePath(d.Path, volumeID)
+	meta, data := cacheFilePath(volumeID)
 
 	data, dataFd, err := allocCacheFile(logger, data, size)
 	if err != nil {
@@ -323,9 +315,6 @@ func teardownDmCache(logger klog.Logger, volumeID string) error {
 	return nil
 }
 
-// FIXME: nowhere to get this path on Unstage
-const DataCachePath = "/var/alibaba-cloud-csi/data-cache"
-
 func clean(path string) error {
 	err := os.RemoveAll(path)
 	if errors.Is(err, os.ErrNotExist) {
@@ -341,6 +330,6 @@ func teardownDataCache(logger klog.Logger, volumeID string) error {
 	}
 	// Note: loop device has LO_FLAGS_AUTOCLEAR set, so it is auto removed after teardownDmCache.
 
-	meta, data := cacheFilePath(DataCachePath, volumeID)
+	meta, data := cacheFilePath(volumeID)
 	return errors.Join(clean(meta), clean(data))
 }
--- a/pkg/disk/data_cache.go
+++ b/pkg/disk/data_cache.go
@@ -275,6 +275,10 @@ func setupDataCache(logger klog.Logger, d *dataCache, device, volumeID string) (
 
 	data, dataFd, err := allocCacheFile(logger, data, size)
 	if err != nil {
+		if errors.Is(err, os.ErrNotExist) {
+			logger.V(1).Info("data cache not exist on node, proceed without cache")
+			return device, nil
+		}
 		return "", fmt.Errorf("failed to allocate cache file: %v", err)
 	}
 	defer loggedClose(logger, dataFd) // be sure to close these FDs after setupDmCache, or loop device will be removed
--- a/pkg/disk/data_cache.go
+++ b/pkg/disk/data_cache.go
@@ -1,0 +1,346 @@
+package disk
+
+import (
+	"bytes"
+	"errors"
+	"fmt"
+	"os"
+	"path/filepath"
+	"structs"
+	"syscall"
+	"unsafe"
+
+	"golang.org/x/sys/unix"
+	"k8s.io/apimachinery/pkg/api/resource"
+	"k8s.io/klog/v2"
+)
+
+type DataCacheMode string
+
+const (
+	DataCacheWritethrough DataCacheMode = "writethrough"
+	DataCacheWriteback    DataCacheMode = "writeback"
+)
+
+type dataCache struct {
+	Size resource.Quantity
+	Mode DataCacheMode
+	Path string
+}
+
+const (
+	DataCacheModeKey = "dataCacheMode"
+	DataCacheSizeKey = "dataCacheSize"
+	DataCachePathKey = "dataCachePath"
+)
+
+func getDataCacheOpts(opts map[string]string, d *dataCache) error {
+	if s := opts[DataCacheSizeKey]; s != "" {
+		size, err := resource.ParseQuantity(s)
+		if err != nil {
+			return fmt.Errorf("invalid %s: %w", DataCacheSizeKey, err)
+		}
+		d.Size = size
+	}
+
+	switch m := DataCacheMode(opts[DataCacheModeKey]); m {
+	case "", DataCacheWriteback, DataCacheWritethrough:
+		d.Mode = m
+	default:
+		return fmt.Errorf("unrecognized %s: %s", DataCacheModeKey, m)
+	}
+
+	d.Path = opts[DataCachePathKey]
+	if d.Path != "" && d.Path[0] != '/' {
+		return fmt.Errorf("%s must be an absolute path", DataCachePathKey)
+	}
+
+	if d.Mode != "" || d.Path != "" || !d.Size.IsZero() {
+		if d.Path == "" {
+			return fmt.Errorf("must specify %s for dataCache", DataCachePathKey)
+		}
+		if d.Size.IsZero() {
+			return fmt.Errorf("must specify non-zero %s for dataCache", DataCacheSizeKey)
+		}
+		if d.Mode == "" {
+			d.Mode = DataCacheWritethrough
+		}
+	}
+	return nil
+}
+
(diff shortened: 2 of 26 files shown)


pkg/disk/data_cache.go
```diff
@@ -0,0 +1,339 @@
...
+	default:
+		return fmt.Errorf("unrecognized %s: %s", DataCacheModeKey, m)
+	}
+
+	if d.Mode != "" || !d.Size.IsZero() {
+		if d.Size.IsZero() {
+			return fmt.Errorf("must specify non-zero %s for dataCache", DataCacheSizeKey)
+		}
+		if d.Mode == "" {
+			d.Mode = DataCacheWritethrough
```
mowangdk: I prefer to enable Write-back mode by default to push the performance to its maximum potential. For data-critical applications like databases, the risk is mitigated at the application layer, where internal binlogs and WAL are already in place to ensure high availability and data consistency

pkg/disk/data_cache.go
```diff
@@ -0,0 +1,339 @@
...
+	}
+	conf := unix.LoopConfig{
+		Fd:   uint32(fd),
+		Size: 4 << 10,
+		Info: unix.LoopInfo64{
+			Flags: unix.LO_FLAGS_DIRECT_IO | unix.LO_FLAGS_AUTOCLEAR,
+		},
+	}
+	copy(conf.Info.File_name[:], path)
+	err = unix.IoctlLoopConfigure(loop, &conf) // Since Linux kernel 5.8
```
mowangdk: TODO: add this limitation in docs

pkg/disk/nodeserver.go
```diff
@@ -620,6 +620,17 @@ func (ns *nodeServer) setupDisk(ctx context.Context, device, targetPath string,
 		omitfsck = true
 	}
 
+	// Setup DataCache
```
mowangdk, line 659: We need to eliminate cache devices from the VFIO stack to prevent incompatibility issues.

pkg/disk/data_cache.go
huww98: > enable Write-back mode by default

Then we need to wait for write back on unstage. This could take quite a long time. So this should only be enabled at new Kubernetes version, which don't do force detach when the node is healthy.

> For data-critical applications like databases

I'm not sure how much benifest there are. Database issues fsync() frequently, which will force write-back from Google search. And RAM can also hold some dirty cache. Do you think DBs need GBs of dirty cache that RAM cannot handle?

mowangdk (commented):
## Automated Code Review

**Summary:** Introduces dm-cache (device-mapper cache) support for cloud disks, using local NVMe/SSD as a cache tier via loop devices. Includes setup, resize, and teardown of dm-cache backed by fallocated files.

**Issues Found:**
- **Unsafe pointer usage**: The `dmi_t` struct uses `unsafe.Pointer` with `unix.Syscall` for ioctl calls. While necessary for dm-ioctl, ensure the struct layout exactly matches the kernel's expectations — any padding mismatch could corrupt memory.
- **Error handling in `allocCacheFile`**: If `IoctlLoopConfigure` fails, the loop device fd is leaked (opened but never closed). Add a `defer loggedClose(logger, loop)` before the configure call, or close on error path.
- **Race condition potential**: `loop_get_free()` gets a free loop device number, but between getting the number and opening it, another process could claim it. Consider retrying on EBUSY.
- **`structs.HostLayout`**: This is a Go 1.23+ feature. Confirm the project's minimum Go version supports this.
- The `DataCachePath` constant uses `/var/alibaba-cloud-csi/data-cache` — this is a host path, so the Helm chart correctly adds the volume mount (seen in the template diff).

**Suggestions:**
- Add cleanup logic (or documentation) for what happens when a node reboots — loop devices configured with `LO_FLAGS_AUTOCLEAR` will disappear, but the dm-cache device entries may not persist across reboots either.
- Consider adding metrics for cache hit/miss ratios to help users understand cache effectiveness.
- Unit tests for `getDataCacheOpts` parsing would be valuable.

**Overall:** Complex but well-implemented feature leveraging dm-cache for performance. The low-level kernel interface code is inherently delicate — careful testing on target kernels is essential. The fd leak in the error path should be fixed.

pkg/disk/data_cache.go
```diff
@@ -0,0 +1,339 @@
...
+
+func dmIoctl(fd int, action uintptr, volumeID string, flags uint32) syscall.Errno {
+	dm := unix.DmIoctl{
+		Version:    [3]uint32{4, 0, 0},
+		Data_size:  unix.SizeofDmIoctl,
+		Data_start: unix.SizeofDmIoctl,
+		Flags:      flags,
+	}
+	copy(dm.Name[:], volumeID)
+	_, _, err := unix.Syscall(unix.SYS_IOCTL, uintptr(fd), action, uintptr(unsafe.Pointer(&dm)))
```
mowangdk: `allocCacheFile` opens the loop device fd (`loop`) but never closes it on error paths after `Open` succeeds. If `IoctlLoopConfigure` fails, the loop fd leaks. Add a deferred close that is cancelled on success, or restructure so the fd is always returned.

mowangdk (commented):
## Automated Code Review

**Summary:** Introduces dm-cache support as a local cache layer over cloud disks using device-mapper, loop devices, and fallocate.

**Overall:** Powerful feature with careful dm-cache management. Some concerns about resource leaks on error paths and unsafe pointer usage.

pkg/disk/data_cache.go
```diff
@@ -0,0 +1,339 @@
...
+		},
+		DmTargetSpec: unix.DmTargetSpec{
+			Sector_start: 0,
+			Length:       size,
+			Target_type:  [16]byte{'c', 'a', 'c', 'h', 'e'},
+		},
+	}
+	copy(dmi.Name[:], volumeID)
+	copy(dmi.Args[:], args)
+	_, _, errno := unix.Syscall(unix.SYS_IOCTL, uintptr(dmCtrl), unix.DM_TABLE_LOAD, uintptr(unsafe.Pointer(&dmi)))
```
mowangdk: The correctness of passing `unsafe.Pointer(&dmi)` to `SYS_IOCTL` depends on the struct layout matching what the kernel expects exactly. Consider adding a compile-time size assertion (e.g., `var _ [4096 - unsafe.Sizeof(dmi_t{})]byte`) to catch layout drift if the Go struct definitions change.

pkg/disk/data_cache.go
```diff
@@ -0,0 +1,339 @@
...
+	meta = filepath.Join(DataCachePath, volumeID+".meta")
+	data = filepath.Join(DataCachePath, volumeID+".data")
+	return meta, data
+}
+
+func dataCacheDevicePath(volumeID string) string {
+	return "/dev/mapper/" + volumeID
+}
+
+func setupDataCache(logger klog.Logger, d *dataCache, device, volumeID string) (string, error) {
```
mowangdk: `getBlockDeviceCapacity(device)` is called but there's no error handling visible. If it returns 0 (e.g., device not ready), dm-cache is set up with size 0, which would be invalid. Consider validating the return value before calling `setupDmCache`.

pkg/disk/nodeserver.go
```diff
@@ -762,6 +772,11 @@ func (ns *nodeServer) NodeUnstageVolume(ctx context.Context, req *csi.NodeUnstag
 		}
 	}
 
+	// Teardown DataCache
+	if cacheErr := teardownDataCache(logger, req.VolumeId); cacheErr != nil {
```
mowangdk: Data cache teardown happens after unmount. If `teardownDataCache` fails, the function returns an error but the volume is already unmounted. This leaves a lingering dm-cache device that blocks future unstage attempts. Consider whether teardown failures should be non-fatal (log and continue) to avoid permanently blocking the unstage flow.

huww98:
## Manual end-to-end verification on a real ACK cluster

The in-tree dm-cache integration tests auto-run only where the host supports it (Linux + root + loop + dm-cache) and skip cleanly elsewhere. On prow (`verify-unit`) they skip as designed — `loop subsystem unavailable: stat /dev/loop-control: no such file or directory` — while the kernel-free unit tests (`TestFlushToClean_*`, parse/reconcile/`waitForCacheDevice`) run and pass. So I validated the real-kernel paths manually on a cluster.

Cluster: ACK, single ecs.u2a node, Alibaba Cloud Linux 4.0.3, kernel `6.6.102-5.3.alnx4`, cloud_essd origin disks. Built the `csi-plugin` image from this branch and drove real `NodeStageVolume`/`NodeUnstageVolume` through a StorageClass with `dataCacheMode: writeback`, `dataCacheSize`.

### Writeback flush on unstage (data integrity)

Wrote a 200 MiB file through a writeback-cached volume, deleted the pod, then re-mounted the same disk in a fresh pod. Because teardown destroys the entire cache (dm device + loop devices + `.meta`/`.data` files), a matching checksum can only come from the origin cloud disk — i.e. it proves the dirty blocks were flushed.

- `sha256` before delete == `sha256` after re-mount ✅
- Unstage log: `switched dm-cache to cleaner policy, waiting for flush dirty=2` → `dm-cache flushed policy="cleaner" writeback=false` → `teardown dm-cache`
- After teardown: no leftover dm device, loop devices auto-cleared (`LO_FLAGS_AUTOCLEAR`), cache files deleted.

### Flush scales with dirty volume (fio under sustained load)

Ran fio (256k random writes, iodepth 32, 3 GiB working set over a 4 GiB writeback cache). The kernel only drains the cache after ~1s idle (`dm_iot_idle_for(HZ)` → `clean_target_met`), so under continuous IO the dirty set stays large:

- Steady state: `used ≈ 12317/16384`, `dirty ≈ 11500` (holding).
- On teardown with `dirty=10765`: `switched to cleaner` → `dm-cache flushed` after **~26 s** (~2.7 GiB written back to the origin at ~105 MiB/s) → clean teardown.

This confirms the drain loop converges and completes within the CSI timeout even with a large dirty set. (There is no table knob to lengthen the background flush period; keeping the device busy is what holds blocks dirty.)

### /dev/mapper udev race → potential reformat of a populated disk (fixed)

The raw device-mapper ioctls resume the device without the udev synchronization `dmsetup` performs, so `/dev/mapper/<vol>` appears asynchronously. On the pre-fix image, every stage logged:

```
setup dm-cache
Mount failed: special device /dev/mapper/<vol> does not exist
```

`FormatAndMount` reads that ENOENT as "unformatted", so if the node races into existence just before its `mkfs`, an already-populated disk would be reformatted. Verified `blkid` on an absent node exits 2 (→ treated as unformatted) on this kernel, confirming the vector.

After adding the `waitForCacheDevice` poll in `setupDataCache`:

- Fresh stage: goes straight to format, **no** "does not exist".
- **Re-stage of a populated disk** (write data → delete pod → recreate): goes straight to `mount successful` in ~28 ms, **no format attempt**, data intact. This is the case that was at risk.

### Online expansion — filesystem and block (writeback)

Expanded both a filesystem and a raw block PVC from 20Gi → 40Gi while running, data written beforehand:

| | filesystem | block |
|---|---|---|
| dm-cache target | `resize dm-cache ... size=83886080 oldSize=41943040` | same |
| after resize | ext4 grown on the cache device, `Expand successful capacity="40 GiB"` | `skipping fs expand for block volume` (target grown, no fs/partition step) |
| device size in pod | 20 GiB → ~40 GiB | 21474836480 → 42949672960 B (exactly 40 GiB) |
| pre-existing data | checksum unchanged across resize | checksum unchanged across resize |
| newly-grown region | filled 1 GiB more, OK | wrote+read 50 MiB at 30 GiB offset, OK |

Then deleted both pods (writeback flush on unstage) and re-mounted the block PVC in a fresh pod. With the cache fully destroyed, both the head data **and** the region written *after* the resize matched byte-for-byte — so the flush persisted everything (including post-resize writes) to the origin. Node left clean (no leftover dm/loop devices). Covered in tests by `TestDataCacheResize`.

### Reboot recovery / restage reconcile

- An aborted teardown leaves the cache in cleaner policy; a subsequent StageVolume reconciles it back to writeback/mq (integration test `TestDataCacheStageReconcilesPolicy`, also observed live).

pkg/disk/data_cache.go
huww98: Fixed

pkg/disk/nodeserver.go
huww98: I think no. we need to ensure teardown success, to avoid leak device, files, or miss writeback.

pkg/disk/data_cache.go
huww98: fixed

mowangdk:
> Wrote a 200 MiB file through a writeback-cached volume, deleted the pod, then re-mounted the same disk in a fresh pod. 

Could you confirm whether this has been tested on a fresh node? The desired behavior is to remove the metadata during NodeUnstageVolume, while retaining the cache for reuse across multiple pods, particularly in image disk use cases.

pkg/disk/csi_agent.go
```diff
@@ -29,13 +30,22 @@ func NewCSIAgent() *CSIAgent {
 	}
 	GlobalConfigVar.MetricEnable = true
 
+	dmControl, err := datacache.OpenDmControl()
```
mowangdk, line 33: We probably don’t need to support caching for serverless use cases.

pkg/disk/datacache/data_cache_linux.go
```diff
@@ -0,0 +1,215 @@
...
+	dmStatusTableFlag = unix.DM_STATUS_TABLE_FLAG
+)
+
+func fallocate(path string, size int64) (int, error) {
+	fd, err := unix.Open(path, unix.O_CREAT|unix.O_RDWR|unix.O_CLOEXEC, 0600)
+	if err != nil {
+		return 0, fmt.Errorf("failed to open %q: %w", path, err)
+	}
+
+	err = unix.Fallocate(fd, 0, 0, size)
```
mowangdk, line 30: How to free the allocated size? i don't think it would be recycled by close fd

pkg/disk/utils.go
```diff
@@ -586,6 +587,11 @@ func getDiskVolumeOptions(
 	}
 	diskVolArgs.RequestGB = requestGB
 
+	err = datacache.GetOpts(volOptions, &diskVolArgs.DataCache)
```
mowangdk, line 590: Please add unittest for the changing code

pkg/utils/io/device.go
```diff
@@ -51,3 +53,16 @@ func (RealDevTmpFS) ListDisks() ([]string, error) {
 	}
 	return disks, nil
 }
+
```
mowangdk, line 56: Please add ut

pkg/disk/csi_agent.go
huww98: No harm to also support it?

pkg/disk/datacache/data_cache_linux.go
huww98: By delete the file in NodeUnstageVolume.

pkg/utils/io/device.go
huww98: added

pkg/disk/utils.go
huww98: added

huww98:
> remove the metadata during NodeUnstageVolume, while retaining the cache

That would be very complex. How do we know if there will be next pod? Do we always preserve the cache on Unstage and scan for filesystem usage to GC least recently used cache under pressure?

And if that disk is modified on another node, then attached back to our node, the cache would be stall and we will never know.

> particularly in image disk use cases

Even more complex. I think image cache may use different diskID for the same image?

pkg/disk/csi_agent.go
mowangdk: I don’t think there will be a local disk in the serverless case, so this may be dead code.

pkg/disk/datacache/data_cache_linux.go
mowangdk: If NodeStageVolume does not succeed, will kubelet avoid calling NodeUnstageVolume?

huww98: If we return Aborted (see latest changes), kubelet will always call NodeUnstageVolume.

mowangdk:
> And if that disk is modified on another node, then attached back to our node, the cache would be stall and we will never know.

I think it won't be modified on another node unless force deleting

> Even more complex. I think image cache may use different diskID for the same image?

I have checked with a colleague, and there will be one disk in ACK mode. Could you please double-check?

huww98:
> there will be one disk in ACK mode

What about the pod is deleted, then 3 minutes later another pod comes with the same image?

pkg/disk/datacache/data_cache_linux.go
mowangdk: Ignore it. I thought it was called by NodeStageVolume.

pkg/disk/nodeserver.go
```diff
@@ -714,6 +740,16 @@ func (ns *nodeServer) unmountTargetPath(logger klog.Logger, targetPath, volumeID
 	return nil
 }
 
+func (ns *nodeServer) teardownStageTarget(ctx context.Context, targetPath, volumeID string) error {
+	if err := ns.unmountTargetPath(klog.FromContext(ctx), targetPath, volumeID); err != nil {
+		return err
+	}
+	if cacheErr := datacache.Teardown(ctx, ns.dmControl, volumeID); cacheErr != nil {
```
mowangdk, line 764: We should add a log or event to indicate that the disk cache setting is ignored on runD.

huww98:
### e2e: "DataCache not effective" warning event

Verified the new fallback warning surfaces in `kubectl describe pod` on a real cluster (ACK, kernel 6.6.102). Reproduced the silent-fallback case by renaming the node's cache dir away so `Setup` mounts the origin instead of a dm-cache device, then created a pod with a `dataCacheMode: writeback, dataCacheSize: 4Gi` StorageClass:

```
$ kubectl describe pod dc-warn-pod
...
Events:
  Type     Reason             Age   From        Message
  ----     ------             ----  ----        -------
  Warning  DataCacheFallback  5s    csi-plugin  DataCache enabled but not effective, check cache path exists on node
```

Node log confirmed the fallback that triggered it:
```
data_cache.go:193 "data cache not exist on node, proceed without cache" volumeID="d-..."
```

The event lands in the pod's namespace and references the pod (via the `csi.storage.k8s.io/pod.{name,namespace,uid}` volume-context keys, populated because the diskplugin CSIDriver sets `podInfoOnMount: true`), so `kubectl describe pod` finds it. `From` is now `csi-plugin` (the ACK node component) rather than `csi-controller-server` — `NewEventRecorder` takes a component argument; node-side callers use `csi-plugin`, controller-side keep `csi-controller-server`.

PR #1623, opened 2026-01-29
refactor(metrics): restructure pkg/metrics and expose GetMetrics for volume Prometheus metrics

iltyty:
#### What type of PR is this?
/kind refactor

Restructure pkg/metrics and expose GetMetrics for volume Prometheus metrics.

Fixes #

```release-note

```

```docs

```


mowangdk (commented):
## Automated Code Review

**Summary:** Refactors the metrics package to expose a `GetMetrics()` function that returns structured `[]*Metric` objects, decoupling metric collection from Prometheus-specific channel-based collection. Also adds a `Get()` method to the `Collector` interface.

**Issues Found:**
- **Breaking interface change**: Adding `Get() []*Metric` to the `Collector` interface is a breaking change — all existing implementations must be updated. Verify that all collectors (disk, NFS, grpc) implement this new method.
- The `convertLabelsToString` function dereferences `*pair.Value` without nil checking. If the prometheus client model returns a nil value pointer, this will panic.
- `GetMetrics()` calls `initCSICollector()` if the instance is nil — this introduces a potential race if called concurrently from multiple goroutines. The original `newCSICollector` had the same pattern, but exposing it publicly increases risk.
- Renamed `newCSICollector` → `initCSICollector` but the nil-check guard (`if csiCollectorInstance != nil { return }`) isn't thread-safe without synchronization.
- The `typedFactorDesc` struct now embeds `*MetaDesc` instead of `*prometheus.Desc`, and the `mustNewConstMetric` method was removed — ensure no remaining callers.

**Suggestions:**
- Consider using `sync.Once` for `initCSICollector` to prevent race conditions.
- Add a brief doc comment on `GetMetrics` explaining its purpose (programmatic access vs Prometheus scraping).
- The `prometheus/client_model` dependency promotion from indirect to direct is correct since it's now explicitly imported.

**Overall:** Good architectural improvement separating metric collection from Prometheus transport. The interface addition needs careful verification that all implementations are updated. Thread-safety of initialization should be addressed.

mowangdk (commented):
## Automated Code Review

**Summary:** Restructures the metrics package to expose a `GetMetrics` API and adds a `Get()` method to the `Collector` interface, enabling programmatic access to volume metrics outside of Prometheus scrape.

**Overall:** Good refactoring direction. One interface design concern.

pkg/metric/collector.go
```diff
@@ -38,15 +38,15 @@ func registerCollector(collector string, factory collectorFactoryFunc, relatedDr
 type Collector interface {
 	// Get new metrics and expose them via prometheus registry.
 	Update(ch chan<- prometheus.Metric) error
```
mowangdk, line 40: Adding `Get() []*Metric` to the `Collector` interface is a breaking change for any external implementations. Since this is an internal interface, it's likely fine, but note that every collector must now implement `Get()` — if a collector has no meaningful `Get()` semantics (like the gRPC exec time collector), it should return nil explicitly. Verify all collectors implement this.

PR #1679, opened 2026-04-24
Support sysconfig for disk & nas

mowangdk:
#### What type of PR is this?

/kind feature

Support read ahead for disk & nas

Fixes #

```release-note

```

```docs

```

### DIFF
--- a/pkg/disk/utils.go
+++ b/pkg/disk/utils.go
@@ -729,6 +729,10 @@ func getMountedVolumeDevice(mnts []k8smount.MountInfo, targetPath string) string
 	return ""
 }
 
+func allowSysConfigKey(key string) bool {
+	return strings.HasPrefix(key, "queue/")
+}
+
 func isDeviceMountedAt(mnts []k8smount.MountInfo, device, targetPath string) bool {
 	for _, mnt := range mnts {
 		if mnt.MountPoint == targetPath && mnt.Source == device {
--- a/pkg/disk/utils.go
+++ b/pkg/disk/utils.go
@@ -729,10 +729,6 @@ func getMountedVolumeDevice(mnts []k8smount.MountInfo, targetPath string) string
 	return ""
 }
 
-func allowSysConfigKey(key string) bool {
-	return strings.HasPrefix(key, "queue/")
-}
-
 func isDeviceMountedAt(mnts []k8smount.MountInfo, device, targetPath string) bool {
 	for _, mnt := range mnts {
 		if mnt.MountPoint == targetPath && mnt.Source == device {
--- a/pkg/nas/nodeserver.go
+++ b/pkg/nas/nodeserver.go
@@ -37,6 +37,7 @@ import (
 	"github.com/kubernetes-sigs/alibaba-cloud-csi-driver/pkg/mounter"
 	"github.com/kubernetes-sigs/alibaba-cloud-csi-driver/pkg/nas/internal"
 	"github.com/kubernetes-sigs/alibaba-cloud-csi-driver/pkg/utils"
+	utilsio "github.com/kubernetes-sigs/alibaba-cloud-csi-driver/pkg/utils/io"
 	"github.com/kubernetes-sigs/alibaba-cloud-csi-driver/pkg/utils/rund/directvolume"
 	"google.golang.org/grpc/codes"
 	"google.golang.org/grpc/status"
@@ -95,6 +96,7 @@ type Options struct {
 	MountProtocol string `json:"mountProtocol"`
 	ClientType    string `json:"clientType"`
 	FSType        string `json:"fsType"`
+	SysConfigs    map[string]string
 	AkID          string
 	AkSecret      string
 }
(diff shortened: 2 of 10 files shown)


pkg/disk/utils.go
```diff
@@ -729,6 +729,10 @@ func getMountedVolumeDevice(mnts []k8smount.MountInfo, targetPath string) string
 	return ""
 }
 
+func allowSysConfigKey(key string) bool {
+	return strings.HasPrefix(key, "queue/")
+}
```
huww98: Is this a breaking change? I think we don't have this restriction before.

pkg/nas/nodeserver.go
```diff
@@ -253,6 +255,12 @@ func (ns *nodeServer) NodePublishVolume(ctx context.Context, req *csi.NodePublis
 	opt.AkID = req.Secrets[akIDKey]
 	opt.AkSecret = req.Secrets[akSecretKey]
 
+	var err error
+	opt.SysConfigs, err = utilsio.ParseSysConfigs(req.VolumeContext["sysConfig"], allowSysConfigKey)
```
huww98, line 259: 似乎局部变量就够了？我们需要这个opt的字段吗？

pkg/utils/io/sysconfig.go
```diff
@@ -0,0 +1,69 @@
...
+// Set the sysconfig for the device using a key and value.
+//
+// Generally, the value is set to "/sys/dev/block/<major:minor>/<key>".
+// Specially, BDI settings like "bdi/read_ahead_kb" are set to /sys/class/bdi/<major:minor>/<key-suffix>,
+// since non-block filesystems like NFS provide their own BDI but don't get a directory under /sys/dev/block.
+func (m *SysConfigManager) Set(key, value string) error {
+	devNum := fmt.Sprintf("%d:%d", m.major, m.minor)
+	var path string
+	key = filepath.Clean(key)
+	if strings.HasPrefix(key, "bdi/") {
```
huww98: prefer strings.CutPrefix

pkg/utils/io/sysconfig.go
```diff
@@ -0,0 +1,69 @@
...
+	for _, kv := range strings.Split(s, ",") {
+		key, value, found := strings.Cut(kv, "=")
+		if !found || key == "" {
+			return nil, errors.New("invalid sysconfig format")
+		}
+		// Prevent users from accessing unexpected files through:
+		// 1. "subsystem" symlinks like /sys/block/vda/subsystem/vdb
+		// 2. relative paths containing ".." like ../../../root/.ssh/id_rsa
+		key = filepath.Clean(key)
+		if strings.HasPrefix(key, "../") || strings.Contains(key, "subsystem") || !allow(key) {
```
huww98: this subsystem check can be fragile, e.g. `/sys/block/nvme0n1/device/device/driver/` can access all PCI device nodes.
It can also be a breaking change if someone is really using subsystem link.

Also note we cannot just use `strings.HasPrefix(key, "subsystem")`, because we also have `/sys/block/nvme0n1/device/subsystem`

pkg/utils/io/sysconfig.go
```diff
@@ -0,0 +1,69 @@
...
+	"strings"
+
+	"golang.org/x/sys/unix"
+	"k8s.io/klog/v2"
+)
+
+const defaultSysfsPath = "/sys"
+
+// ParseSysConfigs accepts s in the format 'key1=value1,key2=value2'.
+func ParseSysConfigs(s string, allow func(key string) bool) (map[string]string, error) {
```
huww98: Prefer slice over map for the return value.
Slice is faster, takes less memory, and can be iterated at deterministic order.

mowangdk:
test:

[root@xxxx ~]# mount | grep nas
xxxxx-sfnx.cn-hangzhou.extreme.nas.aliyuncs.com:/ on /var/lib/kubelet/pods/ca270b20-9873-411c-9d6e-d350d6208939/volumes/kubernetes.io~csi/nas-access-point-pv/mount type nfs (rw,relatime,vers=3,rsize=1048576,wsize=1048576,namlen=255,hard,nolock,noresvport,proto=tcp,timeo=600,retrans=2,sec=sys,mountaddr=192.168.1.6,mountvers=3,mountport=2049,mountproto=tcp,local_lock=all,addr=192.168.1.6)
alinas-ap-xxxxx.xxxxx-hhb36.cn-hangzhou.tls.127.0.1.1-FZN7FRun:/ on /var/lib/kubelet/pods/f146cb66-1c0d-4cc1-8df4-727b59e52fb2/volumes/kubernetes.io~csi/nas-access-point-pv-1/mount type nfs (rw,relatime,vers=3,rsize=1048576,wsize=1048576,namlen=255,hard,nolock,noresvport,proto=tcp,port=12049,timeo=600,retrans=2,sec=sys,mountaddr=127.0.1.1,mountvers=3,mountport=12049,mountproto=tcp,local_lock=all,addr=127.0.1.1)

[root@xxxxx ~]# cat /sys/class/bdi/0:223/read_ahead_kb
1024

[root@xxxx ~]# cat /sys/class/bdi/0:265/read_ahead_kb
1024

pkg/disk/utils.go
mowangdk: Why the e2e test passed if this is a breaking change .....

mowangdk: I remember we got this e2e test for disk?

pkg/nas/nodeserver.go
mowangdk: I don’t want to change this in this PR. Since there are no other references besides this method, we can remove it later.

pkg/utils/io/sysconfig.go
mowangdk: This block was copied from a deprecated PR. We should remove this subsystem check?

pkg/disk/utils.go
huww98: I think we did not test this sysConfig function in CI. AFAIK, this is only used on ASI.

PR #1682, opened 2026-04-30
feat(oss): implement version-aware ResourceVersion strategy for fuse pod delete

AlbeeSo:
Implement intelligent ResourceVersion selection based on Kubernetes version to balance data consistency and etcd pressure in large-scale clusters.

#### What type of PR is this?

/kind feature

Key Changes:
- Add K8s version detection using Discovery API at startup (main.go)
- Implement ShouldConstrainResourceVersion() with version-based logic:
  * K8s >= 1.31: Use RV="" (Consistent Reads from Cache supported)
  * K8s < 1.31: Use RV="0" (avoid etcd quorum read pressure)
  * K8s unknown: Use RV="0" (conservative approach)
- Add Featuregate override support: ConstrainFusePodDeleteRV=true/false
- Replace NewFilteredPodInformer with custom ListWatch to control RV

Design Decisions:
- K8s >= 1.31: Consistent Reads from Cache provides both consistency
  and performance (reads from cache instead of etcd)
- K8s < 1.31: Force RV="0" to avoid etcd quorum reads that could
  cause mount/unmount hangs in large clusters
- Version detection happens once at startup, converted to bool flag
  to avoid repeated version checks in hot path

Logging Levels:
- Info: K8s >= 1.31 (normal operation with new feature)
- Warning: K8s < 1.31 (using legacy approach with etcd pressure risk)
- Error: K8s version unknown (conservative fallback)

Testing:
- Add TestShouldConstrainResourceVersion with 8 test cases:
  * K8s version detection (1.28, 1.30, 1.31, 1.32, nil)
  * Featuregate override (true/false/invalid)
- Update TestDelete to cover both constrainRV modes
- All existing tests updated for new function signatures

Files Modified:
- main.go: Add detectKubernetesVersion(), pass to OSS driver
- pkg/oss/oss.go: Accept and propagate k8sVersion
- pkg/oss/controllerserver.go: Store k8sVersion field
- pkg/mounter/fuse_pod_manager/oss/registry.go: 
  ShouldConstrainResourceVersion() with Featuregate support
- pkg/mounter/fuse_pod_manager/fuse_pod_manager.go: 
  Use constrainResourceVersion flag in Delete()
- pkg/mounter/fuse_pod_manager/oss/oss_fuse_pod_manager.go: 
  Update constructor signature
- Test files: Update all test cases for new signatures

Impact:
- Prevents fuse pod delete from missing newly created pods
- Reduces etcd pressure in K8s 1.31+ clusters (3x performance gain)
- Avoids mount/unmount hangs in K8s < 1.31 large clusters
- Backward compatible with Featuregate override for manual control

Fixes #

```release-note

```

```docs

```


pkg/mounter/fuse_pod_manager/oss/registry.go
```diff
@@ -65,13 +69,50 @@ func GetAllRegisteredFuseTypes() []string {
 	return types
 }
 
+// ShouldConstrainResourceVersion determines whether to constrain ResourceVersion to "" for consistent reads
+// Priority: ENV override > K8s version detection
+// - ENV OSS_FUSE_CONSTRAIN_RV=true/false: If set, use this value
+// - K8s >= 1.31: false (Consistent Reads from Cache supported, no need to constrain)
+// - K8s < 1.31: true (Constrain RV="0" to use watch cache and avoid etcd pressure)
+// - K8s unknown: true (Conservative approach, constrain RV="0")
+func ShouldConstrainResourceVersion(k8sVersion *k8sver.Version) bool {
```
mowangdk, line 77: Change OSS_FUSE_CONSTRAIN_RV to  featuregate? User can't set env in managed-csiprovisioner

mowangdk:
From AI:
  ---
  1. Dead k8sVersion fields on controllerServer and OSS structs

  The PR adds k8sVersion *k8sver.Version to both the OSS struct (pkg/oss/oss.go) and controllerServer struct (pkg/oss/controllerserver.go), stores the value, but never reads it from either struct. The version is
  consumed once in GetAllOSSFusePodManagers() → ShouldConstrainResourceVersion() and the bool result is baked into FusePodManager. These fields are dead code and should be removed.

  2. klog.Errorf for unknown K8s version should be klog.Warningf

  In ShouldConstrainResourceVersion():
  klog.Errorf("K8s version unknown, constraining ResourceVersion to '0' ...")

  This is a degraded-mode fallback, not an error. In main.go, the same condition (failed version detection) is logged as klog.Warningf. Should be consistent — use klog.Warningf here too.

  3. Unnecessary variable scope widening in main.go

  kubeClient is hoisted to the outer scope (var kubeClient kubernetes.Interface) but is never used outside the else block. Only k8sVersion needs the outer scope. kubeClient could remain a block-scoped := inside the
   else, with k8sVersion assigned separately:

  var k8sVersion *k8sversion.Version
  // ...
  } else {
      kubeClient, err := kubernetes.NewForConfig(cfg)
      // ...
      k8sVersion, err = detectKubernetesVersion(kubeClient)
  }

PR #1605, opened 2026-01-02
disk: get disk quantity and identify LingJun node from metadata

huww98:
#### What type of PR is this?

/kind cleanup

#### What this PR does / why we need it:

Currently, the InstanceType query logic is bound to the volume count logic, primarily because we need both Node.status.volumesAttached (for currently managed disks) and node.annotation (for DiskQuantity) from the same get node response, and the `status` must be fetched in the middle of the volume count logic. This makes it hard to extract more info (e.g. nvmeSupport) from the fetched metadata apart from DiskQuantity.

So I moved `DiskQuantity` logic to `cloud/metadata`, it fits well because we have 3 different places to get this info.
To implement this, two major refactor is done in `cloud/metadata`:
- Support non-string value. In-package component now all returns `any` for the value, but we still keep it type-safe for all public API. This allows us to add 2 new non-string metadata: DiskQuantity (int32) and MachineKind (enum: ECS/LingJun).
- Session support. Introduce a new `m.WithSession(ctx)` API to inject a context and allow retry of previously failed fetchers. So that we can keep retry as the `NodeGetInfo` CSI GRPC call retries.

Then, two new fetchers for ECS `DescribeInstanceTypes` and EFLO `DescribeNodeType` is added. K8s fetcher is extended to parse the annotation.

The disk driver is changed to use the added metadata fields.

#### Which issue(s) this PR fixes:

Fixes #

#### Special notes for your reviewer:

/hold
based on #1599, merge it first

#### Does this PR introduce a user-facing change?

```release-note
NONE
```

#### Additional documentation e.g., KEPs (Kubernetes Enhancement Proposals), usage docs, etc.:

```docs

```

PR #1685, opened 2026-05-06
labeler: add centralized metadata labeler

huww98:
#### What type of PR is this?

/kind feature

Introduce a centralized controller-side labeler that watches Nodes, resolves
ECS instance properties, and writes node.csi.alibabacloud.com/max-disk
annotation plus disktype.* labels. This lets Node plugins skip OpenAPI
entirely in NodeGetInfo and removes the need to ship addon credentials
to every Node.

Runs as a sidecar inside the csi-provisioner Deployment with leader
election. Uses a shared Node informer (with transform to drop unneeded
managed fields) so memory stays bounded even on 10k-node clusters.

Fixes #

```release-note
Add a labeler sidecar for nodes without credential
```

```docs

```

/hold

merge #1605 first


huww98:
Kubernetes deps bump already merged at #1713, we need that to make informer to work with synctest.
/unhold

mowangdk (commented):
## PR Review: `labeler: add centralized metadata labeler`

### Overview

This PR introduces a controller-side labeler sidecar that watches Nodes, resolves ECS instance properties, and writes `node.csi.alibabacloud.com/max-disk` annotation + `disktype.*` labels. This removes the need for each Node plugin to call OpenAPI directly, eliminating per-node credential distribution. The change also significantly refactors the driver lifecycle (`main.go`, driver structs, graceful shutdown).

**Scope**: ~1,200 lines of new code (excluding vendor), ~500 lines of tests, ~400 lines of refactoring across existing packages, and ~34k lines of vendored `k8s.io/client-go/informers`.

---

### Critical: Bug in `useLabeler` NodeGetInfo path

In `pkg/disk/nodeserver.go` (new code in the `useLabeler` branch):

```go
if ns.useLabeler {
    if _, err = getNode(); err != nil {  // ← result discarded
        return nil, status.Errorf(codes.Internal, "failed to get current node for labeler: %v", err)
    }
    maxVolumesNum, err = getVolumeCountFromLabelerAnnotation(node)  // ← node is nil here
```

The return value of `getNode()` is discarded with `_`, but `node` (the outer `var *v1.Node`) is used on the next line. Unless `getNode` is a closure that sets `node` as a side effect, `node` will be `nil` here, causing a nil-pointer dereference. The original code pattern was `node, err = getNode()`. This should be:

```go
node, err = getNode()
```

---

### High Severity

**1. No re-reconcile path when labeler annotation is missing at NodeGetInfo time**

When `--use-labeler=true`, `NodeGetInfo` reads the `max-disk` annotation and returns an error if it's absent (`"max-disk annotation not found"`). If the labeler hasn't processed the node yet (e.g., on a fresh node joining the cluster), this will cause `NodeGetInfo` to fail repeatedly until the labeler catches up. There's no retry/backoff logic - the kubelet will call again, but the error message could be clearer and this race condition should be documented or handled more gracefully (e.g., wait with a short timeout).

**2. `reconcile` creates a new `Metadata` chain per node**

In `reconcile.go`:
```go
m := metadata.NewMetadata()
m.AddKubernetesNode(node)
m.EnableEFLO(r.EFLO)
m.EnableOpenAPI(r.ECS)
```

This builds a fresh metadata chain for every node reconcile. While functionally correct, the `EnableOpenAPI` / `EnableEFLO` calls set up `lazyInit` wrappers that include the full provider chain. For high node counts, this creates significant GC pressure. Consider whether the metadata providers can be shared or pre-built per reconciler.

---

### Medium Severity

**3. `filterMap` over-allocates**

In `labeler.go`:
```go
result := make(map[string]string, len(m))
```

Most node labels/annotations are filtered out, so allocating `len(m)` capacity wastes memory. On 10k-node clusters with nodes carrying 30+ labels, this adds up. Use `0` or a small constant.

**4. `os.Getenv("EBS_CATEGORY_LINGJUN_SUPPORTED")` called per-reconcile**

In `reconcile.go`:
```go
diskTypes := disk.ParseLingjunNodeDiskTypes(os.Getenv("EBS_CATEGORY_LINGJUN_SUPPORTED"))
```

This reads the environment variable on every Lingjun node reconcile. Should be read once at startup and stored in the `Reconciler` struct.

**5. `DescribeDisks` not cached/batched across nodes**

Every node reconcile makes a `DescribeDisks` API call (via `GetAttachedCloudDisks`). With 10 workers and a leader change triggering reconcile of all nodes, this creates a burst of API calls. Unlike `instanceTypeCache` and `diskTypesCache`, there's no deduplication for this call. Consider rate limiting or noting this as an expected load on leader election.

**6. `MachineKindVirtualKubelet` added but not used in labeler**

`pkg/cloud/metadata/k8s.go` adds detection for virtual-kubelet nodes (`MachineKindVirtualKubelet`), but `reconcile.go` treats unknown kinds (including virtual-kubelet) as skip:
```go
default:
    logger.V(3).Info("skip unknown machine kind", "kind", machineKind)
```

This is likely intentional (virtual-kubelets don't have real instances), but the log message says "unknown" which is misleading for a recognized kind. Consider adding an explicit case.

---

### Low Severity / Suggestions

**7. Leader election timing**

`leaseDuration: 30s`, `renewDeadline: 20s` is relatively tight. A brief API server hiccup could cause leadership flips, each triggering a full reconcile sweep. Consider 45s/30s or documenting the rationale.

**8. No `DeleteFunc` on node informer event handler**

The informer only handles Add/Update. Deleted nodes leave their work items in the queue (handled gracefully by the not-found check in reconcile), but there's no cleanup logging. This is fine functionally but could be confusing during debugging.

**9. Removing `InstanceID` from OpenAPI fetcher**

`pkg/cloud/metadata/openapi.go` removes `InstanceID` from the `OpenAPIFetcher`. This is correct since IMDS already provides it, but it's a subtle behavioral change for any code path that relied on falling through to OpenAPI for InstanceID.

---

### What's Done Well

- **Node transform** (`nodeTransform`): Stripping unused fields from the informer cache is the right approach for 10k-node clusters. The field selection is precise and well-documented.
- **TTL cache with single-flight**: Clean implementation with proper error semantics (share errors with concurrent callers, don't cache them). The `CompareAndSwap` for expiry replacement avoids thundering herd on cache misses.
- **Test coverage**: Good test suite covering happy paths, error paths, skip conditions, and cache behavior. Using `synctest` for deterministic time control is a nice touch.
- **Separation of annotations**: Using `max-disk` (labeler-written) vs `allocatable-disk` (NodeGetInfo-written) avoids conflicts between the two modes.
- **Graceful shutdown refactor** in `main.go`: Moving from `signal.Notify` + `os.Exit(0)` to `signal.NotifyContext` + `context.Cancel` + `wg.Wait()` is a solid improvement.
- **EFLO split**: Splitting the monolithic EFLO fetcher into `EfloNodeFetcher` (per-node DescribeNode) and `EfloNodeTypeFetcher` (per-nodeType DescribeNodeType) is cleaner and enables better caching.

---

### Summary

The architecture is sound - centralizing metadata resolution into a controller sidecar is the right approach for removing per-node credentials. The **`node` nil-pointer bug in the `useLabeler` path** is the must-fix before merge. The other issues are optimization opportunities and hardening suggestions. The refactoring of driver lifecycle and ECS v1→v2 migration are clean.

pkg/cloud/utils.go
```diff
@@ -31,6 +33,32 @@ func UnmarshalAcsResponse(jsonBytes []byte, res responses.AcsResponse) {
...
 
+// UnmarshalV2Response populates an ECS v2 SDK response struct from the
+// raw API JSON body. It simulates the SDK's internal deserialization
+// pipeline: parse JSON → wrap in {"body":...} → dara.Convert to struct.
+//
+// Example:
+//
+//	var resp ecs20140526.DescribeDisksResponse
+//	cloud.UnmarshalV2Response([]byte(`{...}`), &resp)
+func UnmarshalV2Response(jsonBytes []byte, res any) {
```
mowangdk, line 44: Why do we need to implement this function separately? Can we just use the one in the SDK?

pkg/bmcpfs/bmcpfs.go
```diff
@@ -98,35 +97,24 @@ func parseNodeID(nodeID string) (instanceID string, kind nodeKind) {
...
-type Driver struct {
-	endpoint string
-	servers  common.Servers
-}
-
-func NewDriver(meta *metadata.Metadata, endpoint string, serviceType utils.ServiceType) *Driver {
-	var driver Driver
-	driver.endpoint = endpoint
-	driver.servers.IdentityServer = newIdentityServer()
+func NewDriver(meta *metadata.Metadata, endpoint string, serviceType utils.ServiceType) *common.Servers {
```
mowangdk: The naming is confusing; the return value is not the driver anymore.

pkg/labeler/labeler.go
```diff
@@ -0,0 +1,251 @@
...
+		NodeLister:        nodeInformer.Lister(),
+		Queue:             queue,
+		instanceTypeCache: NewTTLCache[string, int32](defaultInstanceTypeTTL),
+		nodeTypeCache:     NewTTLCache[string, int32](defaultInstanceTypeTTL),
+		diskTypesCache:    NewTTLCache[diskTypeCacheKey, []string](defaultInstanceTypeTTL),
+	}
+
+	_, err := nodeInformer.Informer().AddEventHandler(cache.ResourceEventHandlerFuncs{
+		AddFunc:    func(obj any) { r.enqueue(obj) },
+		UpdateFunc: func(_, newObj any) { r.enqueue(newObj) },
```
mowangdk, line 181: Why should we need UpdateFunc? Do we wanna to support instance upgrade?

pkg/labeler/labeler.go
```diff
@@ -0,0 +1,251 @@
...
+			ResourceVersion: node.ResourceVersion,
+			UID:             node.UID,
+			Labels:          filteredLabels,
+			Annotations:     filteredAnnotations,
+		},
+		Spec: v1.NodeSpec{
+			ProviderID: node.Spec.ProviderID,
+		},
+		Status: v1.NodeStatus{
+			VolumesInUse:    node.Status.VolumesInUse,
```
mowangdk, line 235: VolumesInUse and VolumeAttached should have no relation to the labeler.

mowangdk (commented):
Please also fix the comments ai summited

pkg/cloud/utils.go
huww98: Didn't find one approprate. Current implementation is already minimal and try to reuse what we can find from SDK.

pkg/labeler/labeler.go
huww98: If someone removed the annotation, we can re-annotate it. Besides, if we delete then re-create a Node object with the same name, we may not get new Add event.

pkg/labeler/labeler.go
huww98: They are used when calculating the max volume count, to exclude the volumes managed by CSI.

pkg/bmcpfs/bmcpfs.go
huww98: Renamed to NewServers

huww98:
> Please also fix the comments ai summited

Took number 3. Others are not correct or not worth fixing.

pkg/disk/nodeserver.go
```diff
@@ -809,13 +862,23 @@ func (ns *nodeServer) NodeGetInfo(ctx context.Context, req *csi.NodeGetInfoReque
...
-	if maxVolumesNum == 0 {
+	regionID := metadata.MustGet(m, metadata.RegionID)
+	zoneID := metadata.MustGet(m, metadata.ZoneID)
+	instanceID := metadata.MustGet(m, metadata.InstanceID)
+
+	if ns.useLabeler {
+		// The labeler has already provisioned max-disk annotation on the node.
+		// Skip all OpenAPI calls below (DescribeInstance/Disks/AvailableResource).
+		// This is the credential-less path for nodes.
+		if _, err = getNode(); err != nil {
```
mowangdk, line 873: However, I don’t like setting values in a closure.

huww98: It is necessary because `getUnmanagedDiskCount` calls `getNode` internally.

PR #1688, opened 2026-05-09
mount-proxy: contextual log and cancel

huww98:
#### What type of PR is this?

/kind cleanup

Currently, if there are concurrent request in mount-proxy, their logs are difficult to read. Added a sequence number to each log.
When client request cancelled, we also cancel it at server side, by detecting the EOF.
Greatly simplify the socket code. We don't need those complex unix package logic.
Correctly handle multiple package, and raise the max message size to 256MB, basically unlimited. The limit is placed just to avoid OOM.

Fixes #

```release-note
NONE
```

```docs

```

### DIFF
--- a/pkg/mounter/proxy/client/client.go
+++ b/pkg/mounter/proxy/client/client.go
@@ -83,7 +83,12 @@ func (c *client) doRequest(req *proxy.Request) (*proxy.Response, error) {
 	}
 
 	var response proxy.Response
-	return &response, json.Unmarshal(p[:end], &response)
+	err = json.Unmarshal(p[:end], &response)
+	if err != nil {
+		return nil, fmt.Errorf("unmarshal response: %w", err)
+	}
+	klog.V(2).InfoS("response from mount-proxy", "seq", response.Seq)
+	return &response, nil
 }
 
 func (c *client) Mount(req *proxy.MountRequest) (*proxy.Response, error) {
--- a/pkg/mounter/proxy/client/client.go
+++ b/pkg/mounter/proxy/client/client.go
@@ -1,7 +1,7 @@
 package client
 
 import (
-	"bytes"
+	"context"
 	"encoding/json"
 	"errors"
 	"fmt"
@@ -9,8 +9,6 @@ import (
 	"time"
 
 	"github.com/kubernetes-sigs/alibaba-cloud-csi-driver/pkg/mounter/proxy"
-	"github.com/kubernetes-sigs/alibaba-cloud-csi-driver/pkg/mounter/utils"
-	"golang.org/x/sys/unix"
 	"k8s.io/klog/v2"
 )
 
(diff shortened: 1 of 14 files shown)


huww98:
## E2E Test Results: mount-proxy compatibility

Tested on ACK cluster (cn-beijing, single node) with OSS static PV mounting via ossfs fuse pod.

### Test Matrix

| Test | csi-plugin | mount-proxy (ossfs) | Result |
|------|-----------|---------------------|--------|
| Baseline | New (proxy-log HEAD) | New (proxy-log HEAD) | :white_check_mark: Pass |
| Backward compat | Old (v1.35.3) | New (proxy-log HEAD) | :white_check_mark: Pass |
| Forward compat | New (proxy-log HEAD) | Old (v1.91.9.ack.1-b81059f) | :white_check_mark: Pass |

### Verification

- OSS volume mounted successfully via ossfs fuse pod in `ack-csi-fuse` namespace
- Test pod wrote and read back data on the mounted path
- New contextualized logs (`seq=N` on mount operations) confirmed in new mount-proxy:
  ```
  handler.go:26] "Start to recvmsg" seq=1
  handler.go:29] "finished recvmsg" seq=1
  driver.go:128] "Started ossfs" seq=1 pid=69 args=[...]
  driver.go:176] "Successfully mounted" seq=1 mountpoint="..."
  handler.go:72] "Start to sendmsg" seq=1
  handler.go:77] "Succeeded to sendmsg" seq=1 n=10
  ```
- Old mount-proxy works with new csi-plugin (old log format, no `seq=`), confirming protocol backward compatibility

pkg/mounter/proxy/client/client.go
```diff
@@ -19,78 +17,94 @@ const (
...
-	data, err := json.Marshal(req)
-	if err != nil {
-		return nil, err
-	}
-
-	unixConn, ok := conn.(*net.UnixConn)
-	if !ok {
-		return nil, errors.New("failed to cast conn to *net.UnixConn")
+	closeConn := func() {
+		if err := conn.Close(); err != nil && !errors.Is(err, net.ErrClosed) {
```
mowangdk, line 40: Should we log errClosed instead of filtering it?

pkg/mounter/proxy/client/client.go
```diff
@@ -19,78 +17,94 @@ const (
...
-	err = unix.Sendmsg(socket, append(data, proxy.MessageEnd), nil, nil, 0)
-	if err != nil {
-		return nil, fmt.Errorf("sendmsg: %w", err)
-	}
-	klog.V(4).InfoS("sendmcg successfully for request", "socket", c.socketPath)
+	// Close connection on context cancellation so server can detect it.
+	requestDone := make(chan struct{})
+	go func() {
+		select {
+		case <-ctx.Done():
```
mowangdk, line 59: ```conn.SetDeadline``` is not enough?

pkg/mounter/proxy/server/server.go
```diff
@@ -0,0 +1,70 @@
...
+			continue
+		}
+		s.wg.Go(func() {
+			s.handle(conn)
+		})
+	}
+}
+
+func (s *Server) handle(conn *net.UnixConn) {
+	seq := s.connSeq.Add(1)
```
mowangdk, line 52: Could the sequence number overflow?

pkg/mounter/proxy/client/client.go
huww98: net.ErrClosed happens if we already closed it before. We call closeConn from two places, so we need to filter it to avoid noise.

pkg/mounter/proxy/client/client.go
huww98: No, SetDeadline only allow read/write to return early. Besides, context can be cancelled before deadline.

pkg/mounter/proxy/server/server.go
huww98: It is Int64. Almost impossible. If overflow, it will wrap aroud, returning negative values.

PR #1689, opened 2026-05-11
Add expired ttl to bmcpfs cache

mowangdk:
/kind feature

Fixes #

Set expired time for bmcpfs vscid cache

```release-note

```

```docs

```

### DIFF
--- a/pkg/bmcpfs/internal/vsc.go
+++ b/pkg/bmcpfs/internal/vsc.go
@@ -120,12 +120,18 @@ func (m *LingjunVscManager) GetVsc(vscId string) (*Vsc, error) {
 
 type vscWithErr struct {
 	*Vsc
-	err error
+	err      error
+	cachedAt time.Time
+}
+
+func (e vscWithErr) isExpired(ttl time.Duration) bool {
+	return time.Since(e.cachedAt) > ttl
 }
 
 type PrimaryVscManagerWithCache struct {
 	VscManager
 	retryTimes int
+	cacheTTL   time.Duration
 
 	cond *sync.Cond
 	// Instance ID to VSC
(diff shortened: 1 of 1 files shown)


huww98:
https://github.com/kubernetes-sigs/alibaba-cloud-csi-driver/pull/1685/changes#diff-96a4c875a7837a47b2cc30484e97163d4a43a4e0b709d109e582844ebc9fb773

Is this useful to you?

huww98:
Note for background: when the node is repaired, the VSC may be destroyed.

huww98 (approved):
/hold
suggesting add unit-test for this. Given that there is no test for PrimaryVscManagerWithCache currently, this is better to have. unhold if you are ok with this.

huww98:
Also, maybe we can recognize some error code when the VSC is not vaild to proactively invalidate the cache, rather than waiting for timeout, which is slow and incur extra API call every 5 minutes.

pkg/bmcpfs/internal/vsc.go
```diff
@@ -255,7 +263,7 @@ func (m *PrimaryVscManagerWithCache) EnsurePrimaryVsc(ctx context.Context, insta
 func (m *PrimaryVscManagerWithCache) GetPrimaryVscOf(instanceId string) (*Vsc, error) {
 	m.cond.L.Lock()
 	cachedVsc, exists := m.cache[instanceId]
-	if exists && cachedVsc.Vsc != nil {
+	if exists && cachedVsc.Vsc != nil && !cachedVsc.isExpired(m.cacheTTL) {
```
huww98, line 266: When expired, and VscManager.GetPrimaryVscOf returns `nil, nil`, this will not be cached, each detaching volume need to query API once. Maybe fine though.

PR #1692, opened 2026-05-12
disk: enable controller on Windows

huww98:
#### What type of PR is this?

/kind feature

Fixes #

```release-note
NONE
```

```docs

```

/hold
wait for #1691

### DIFF
--- a/pkg/disk/cloud.go
+++ b/pkg/disk/cloud.go
@@ -1,5 +1,3 @@
-//go:build !windows
-
 /*
 Copyright 2019 The Kubernetes Authors.
 
@@ -41,6 +39,7 @@ import (
 	"github.com/kubernetes-sigs/alibaba-cloud-csi-driver/pkg/disk/batcher"
 	"github.com/kubernetes-sigs/alibaba-cloud-csi-driver/pkg/disk/waitstatus"
 	"github.com/kubernetes-sigs/alibaba-cloud-csi-driver/pkg/utils"
+	perrors "github.com/pkg/errors"
 	"google.golang.org/grpc/codes"
 	"google.golang.org/grpc/status"
 	v1 "k8s.io/api/core/v1"
@@ -61,10 +60,28 @@ const (
 	DISK_RESIZE_PROCESSING_TIMEOUT = 30 * time.Second
 )
 
+// Disk status returned in ecs.DescribeDisks
+const (
+	DiskStatusInuse     = "In_use"
+	DiskStatusAttaching = "Attaching"
+	DiskStatusDetaching = "Detaching"
+	DiskStatusAvailable = "Available"
+)
+
 const (
 	SnapshotStatusAccomplished = "accomplished"
 	DiskMultiAttachDisabled    = "Disabled"
 	DiskMultiAttachEnabled     = "Enabled"
+
+	// InstanceStatusStopped ecs stopped status
+	InstanceStatusStopped = "Stopped"
+)
+
+const (
+	// DiskAttachedKey attached key
+	DiskAttachedKey = "k8s.aliyun.com"
+	// DiskAttachedValue attached value
+	DiskAttachedValue = "true"
 )
 
 type DiskAttachDetach struct {
@@ -77,8 +94,7 @@ type DiskAttachDetach struct {
 	detachThrottler *throttle.Throttler
 	detaching       sync.Map
 
-	dev    *DeviceManager
-	devMap *devMap
+	repo *diskRepo
 }
 
 type DiskCreateDelete struct {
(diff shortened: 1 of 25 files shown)


mowangdk (commented):
## Automated Code Review

**Summary:** Removes Windows build constraint from `cloud.go`, moves disk status constants and attach/detach logic into shared code, and replaces `DeviceManager`/`devMap` with a unified `diskRepo`. This enables the disk controller to operate on Windows nodes.

**Issues Found:**
- The `forceDetachAllowed` function was removed entirely. If this was only unreachable on Windows, that's fine, but please confirm no callers remain that relied on this logic for Linux BDF scenarios.
- `perrors` import alias added (`github.com/pkg/errors`) — ensure this doesn't conflict with existing `"errors"` stdlib usage in the same file.
- Constants like `DiskAttachedKey`/`DiskAttachedValue` were moved — verify all references across the codebase compile correctly.

**Suggestions:**
- Consider adding a brief comment on `diskRepo` explaining what it unifies (device management previously split between `DeviceManager` and `devMap`).

**Overall:** Well-structured refactor that cleanly separates platform-independent controller logic. The removal of dead code (`forceDetachAllowed`, BDF constants) is good housekeeping.

mowangdk (commented):
## Automated Code Review

**Summary:** Refactors disk attach/detach logic to enable the controller server on Windows by introducing a `diskRepo` abstraction that encapsulates device management, with platform-specific implementations.

**Overall:** Clean platform abstraction. The `diskRepo` nil check pattern (`ad.repo == nil`) to distinguish controller vs. node context is effective. One concern about the `forceDetachAllowed` relocation.

pkg/disk/cloud.go
```diff
@@ -236,9 +149,56 @@ func chooseAttachAction(disk *ecs.Disk, instanceID string) (attachAction, error)
...
+	klog.Infof("forceDetachAllowed: diskResponse: %+v", diskResponse)
+	if len(disks) == 0 {
+		klog.Warningf("forceDetachAllowed: no disk found: %s", disk.DiskId)
+		return false, perrors.Wrapf(err, "forceDetachAllowed: Get disk empty, ID=%s", disk.DiskId)
+	}
+	bdfTagExist := false
+	for _, tag := range disks[0].Tags.Tag {
+		if tag.TagKey == DiskBdfTagKey {
+			bdfTagExist = true
+		}
```
mowangdk, line 196: In `forceDetachAllowed`, when `DescribeDisks` returns an empty list, the code wraps `err` with `perrors.Wrapf` — but at that point `err` is nil (since the call succeeded with 0 results). This produces a misleading error message like `<nil>: forceDetachAllowed: Get disk empty, ID=...`. Use `perrors.Errorf` instead (consistent with the similar pattern for instances below).

pkg/disk/cloud.go
```diff
@@ -236,9 +149,56 @@ func chooseAttachAction(disk *ecs.Disk, instanceID string) (attachAction, error)
...
+	describeDisksRequest.DiskIds = "[\"" + disk.DiskId + "\"]"
+	diskResponse, err := ecsClient.DescribeDisks(describeDisksRequest)
+	if err != nil {
+		klog.Warningf("forceDetachAllowed: error with DescribeDisks: %s, %s", disk.DiskId, err.Error())
+		return false, perrors.Wrapf(err, "DescribeInstances, instanceId=%s", disk.InstanceId)
+	}
+	disks := diskResponse.Disks.Disk
+	klog.Infof("forceDetachAllowed: diskResponse: %+v", diskResponse)
+	if len(disks) == 0 {
+		klog.Warningf("forceDetachAllowed: no disk found: %s", disk.DiskId)
```
mowangdk, line 189: The error message says `DescribeInstances` but the actual failing call is `DescribeDisks`. This was a pre-existing bug carried over from the move, but worth fixing now: change to `DescribeDisks, diskId=%s`.

pkg/disk/cloud.go
huww98: Fixed. Replaced with `fmt.Errorf` and also removed the deprecated `github.com/pkg/errors` import from this file entirely.

pkg/disk/cloud.go
huww98: Fixed. Changed to `fmt.Errorf("DescribeDisks %s: %w", disk.DiskId, err)`.

pkg/disk/disk_repo_unix.go
```diff
@@ -0,0 +1,137 @@
...
+	"context"
+	"errors"
+	"fmt"
+	"strings"
+
+	"k8s.io/apimachinery/pkg/util/sets"
+	"k8s.io/klog/v2"
+)
+
+type diskRepo struct {
```
mowangdk, line 15: What does diskRepo mean?

pkg/disk/utils_unix.go
```diff
@@ -0,0 +1,185 @@
+//go:build !windows
```
mowangdk: Why is the header is differents from the others?

pkg/disk/disk_repo_unix.go
huww98: a collection of fields only meaningful from node. Infra needed to find and interact with real block device

pkg/disk/utils_unix.go
huww98: `!windows` is basically the same as `unix` for us? since we only support linux, darwin, windows.

mowangdk: So, I’d like to keep the head the same as the others.

huww98: changed to unix

PR #1698, opened 2026-05-21
Support vsc attachments for ecs node type

mowangdk:
/kind feature

None

```release-note

```

```docs

```

### DIFF
--- a/pkg/bmcpfs/controllerserver.go
+++ b/pkg/bmcpfs/controllerserver.go
@@ -29,15 +29,17 @@ import (
 	"time"
 
 	openapi "github.com/alibabacloud-go/darabonba-openapi/v2/client"
+	ecsclient "github.com/alibabacloud-go/ecs-20140526/v7/client"
 	efloclient "github.com/alibabacloud-go/eflo-controller-20221215/v3/client"
 	nasclient "github.com/alibabacloud-go/nas-20170626/v4/client"
 	"github.com/alibabacloud-go/tea/tea"
 	alicred_old "github.com/aliyun/credentials-go/credentials"
 	"github.com/container-storage-interface/spec/lib/go/csi"
 	"github.com/kubernetes-sigs/alibaba-cloud-csi-driver/pkg/bmcpfs/internal"
+	"github.com/kubernetes-sigs/alibaba-cloud-csi-driver/pkg/cloud/metadata"
 	"github.com/kubernetes-sigs/alibaba-cloud-csi-driver/pkg/common"
 	"github.com/kubernetes-sigs/alibaba-cloud-csi-driver/pkg/credentials"
-	"github.com/kubernetes-sigs/alibaba-cloud-csi-driver/pkg/nas/cloud"
+	nasCloud "github.com/kubernetes-sigs/alibaba-cloud-csi-driver/pkg/nas/cloud"
 	"github.com/kubernetes-sigs/alibaba-cloud-csi-driver/pkg/utils"
 	"github.com/kubernetes-sigs/alibaba-cloud-csi-driver/pkg/version"
 
@@ -57,10 +59,17 @@ type controllerServer struct {
 	skipDetach     bool
 }
 
-func newControllerServer(region string) (*controllerServer, error) {
+func newControllerServer(meta *metadata.Metadata) (*controllerServer, error) {
+	region := metadata.MustGet(meta, metadata.RegionID)
 	efloClient, err := newEfloClient(region)
 	if err != nil {
 		return nil, err
+	}
+
+	// Init ECS Client (v2 SDK)
+	ecsClient, err := newECSClient(region)
+	if err != nil {
+		return nil, fmt.Errorf("error building ecs client: %w", err)
 	}
 
 	skipDetach := false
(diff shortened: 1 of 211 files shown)


pkg/om/utils_test.go
```diff
@@ -4,25 +4,29 @@ import (
 	"os"
 	"path/filepath"
 	"testing"
-
-	"github.com/stretchr/testify/assert"
-	"github.com/stretchr/testify/require"
```
huww98, line 9: Why?

huww98:
Can we enable VSC for existing ECS nodes, Currently and in the future? If so, we should not encode it into NodeID, since that is immutable. Can we determine if a node support VSC from controller side?

pkg/bmcpfs/controllerserver.go
```diff
@@ -273,15 +283,14 @@ func (cs *controllerServer) ControllerPublishVolume(ctx context.Context, req *cs
...
-	if !strings.HasPrefix(req.NodeId, LingjunNodeIDPrefix) || cs.skipDetach {
+	instanceID, kind := parseNodeID(req.NodeId)
+	if !kind.supportsVSC() || cs.skipDetach {
 		return &csi.ControllerUnpublishVolumeResponse{}, nil
 	}
-	// Create Primary vsc for Lingjun node
-	lingjunInstanceID := strings.TrimPrefix(req.NodeId, LingjunNodeIDPrefix)
-	if LingjunNodeIDPrefix == "" {
-		return nil, status.Error(codes.InvalidArgument, "invalid node id")
+	if instanceID == "" {
```
huww98: dead code, supportsVSC ensures `instanceID` != ""

mowangdk:
> Can we enable VSC for existing ECS nodes, Currently and in the future? If so, we should not encode it into NodeID, since that is immutable. Can we determine if a node support VSC from controller side?

No, it’s tied to the instance type, and AFAIK it doesn’t have feature that upgrade inplace

pkg/bmcpfs/internal/vsc.go
```diff
@@ -51,45 +77,87 @@ type Vsc struct {
...
+	GetVsc(vscId, instanceId string) (*Vsc, error)
 }
 
-func NewVscManager(client *efloclient.Client) VscManager {
-	return &LingjunVscManager{client: client}
+func NewVscManager(client *efloClient.Client, ecsClient *ecsClient.Client) VscManager {
+	return &LingjunVscManager{efloclient: client, ecsclient: ecsClient}
 }
 
 type LingjunVscManager struct {
```
huww98:
```suggestion
type VscManager struct {
```
Since now ECS is also included

huww98:
### Design: Consider splitting `LingjunVscManager` into separate ECS/Eflo backend implementations

Currently every method in `LingjunVscManager` (`CreatePrimaryVscFor`, `GetPrimaryVscOf`, `GetVsc`) branches on `isECSInstance()` to dispatch between ECS and Eflo APIs. This has a few downsides:

1. **Implicit coupling** — `isECSInstance()` re-derives the node kind from the instance ID prefix (`"i-"`), but the caller already knows this from `parseNodeID()`. These two classification systems must stay in sync with no compile-time guarantee.
2. **`GetVsc` signature pollution** — the `instanceId` parameter was added only for dispatch, not because the API needs it.
3. **Extensibility** — adding a third backend means more if/else in every method.

Suggestion: extract a `VscBackend` interface that each backend implements independently, and encapsulate the instance + backend into a single object:

```go
type VscBackend interface {
    CreatePrimaryVsc(instanceId string) (string, error)
    GetPrimaryVscOf(instanceId string) (*Vsc, error)
    GetVsc(vscId string) (*Vsc, error)
}

type efloVscBackend struct { client *efloClient.Client }
type ecsVscBackend struct { client *ecsClient.Client }

// NodeInstance binds an instance ID with its resolved backend,
// so downstream code doesn't need to dispatch.
type NodeInstance struct {
    id      string
    backend VscBackend
}
```

The caller resolves the node once (using the kind it already knows from `parseNodeID`), then passes the `NodeInstance` to the manager:

```go
func (cs *controllerServer) resolveNode(nodeID string) (*NodeInstance, error) {
    instanceID, kind := parseNodeID(nodeID)
    // select backend by kind...
    return &NodeInstance{id: instanceID, backend: cs.backends[kind]}, nil
}

func (m *PrimaryVscManagerWithCache) EnsurePrimaryVsc(ctx context.Context, node *NodeInstance, refresh bool) (string, error) {
    // cache keyed by node.id
    // on miss: node.backend.CreatePrimaryVsc(node.id)
}
```

This eliminates `isECSInstance()` entirely, removes the extra `instanceId` param from `GetVsc`, and makes `PrimaryVscManagerWithCache` purely responsible for caching — no dispatch logic.

### Minor items

- `LingjunConfigFile` in `nodeserver.go:46` is already exported as `metadata.LingjunConfigFile` — can reuse that.
- `newEfloClient()` is duplicated from `pkg/disk/utils.go` — consider extracting to a shared location.
- `newECSClient()` could reuse `utils.GetEcsConfig()` from `pkg/utils/openapi.go` as a base.
- Dead check at `controllerserver.go:290`: `instanceID == ""` is unreachable because `parseNodeID` guarantees empty instanceID implies `nodeKindUnknown`, which returns earlier via `!kind.supportsVSC()`.

PR #1709, opened 2026-05-29
nas: support AccessMode and readonly field for readonly mounts

huww98:
#### What type of PR is this?

/kind feature

The NAS driver currently ignores both `req.Readonly` and `VolumeCapability.AccessMode` for readonly semantics. This PR adds proper support:

- Appends "ro" to mount options when `req.Readonly` is true or the access mode is `READER_ONLY`
- Updates `ValidateVolumeCapabilities` to accept all standard mount-type access modes (previously only accepted `MULTI_NODE_MULTI_WRITER`) and reject block volumes

Also refactors `Options.Options` from a comma-separated string to `[]string` for consistency with the mount infrastructure.

NONE

Split into two logical commits:
1. Refactor `Options.Options` from `string` to `[]string`
2. Add AccessMode and readonly support

**Potentially confusing points:**

1. **Proxy server still splits options internally (intentional)**
   `addAutoFallbackNFSMountOptions` in the mount-proxy server keeps its own `SplitMountOptions` call, even though the NAS nodeserver now passes pre-split options. The proxy server receives requests from multiple clients (including bmcpfs) that may still send comma-separated strings. Server and client may also be upgraded independently, so the server must remain backward-compatible.

2. **`readOnly` flag does NOT account for user-specified "ro" in mountFlags**
   If a user puts `ro` in SC `mountOptions` but uses a RW access mode, `readOnly` stays `false`. The mount is effectively read-only (kernel honors the option), but `chmod` will still be attempted (fails with EROFS, logged but not fatal). This is acceptable — detecting "ro" from options is not straightforward since users could specify `"ro,rw"` where `ro` doesn't actually mean the mount is read-only (last option wins). Reliably deriving readonly state from options is complex and probably not worthwhile.

3. **Duplicate "ro" option is possible**
   If `ro` is in mountFlags AND the access mode is `ReadOnlyMany`, `opt.Options` will contain `["...", "ro", "ro"]`. The kernel tolerates duplicates. For the same reason as above, deduplicating is not worth the complexity.

4. **Subpath auto-creation strips "ro" for temporary root mount**
   When auto-creating a subpath directory, the code mounts the NAS root without "ro" to a private temp path (kubelet plugin dir, 0700), creates the directory, then unmounts. The final pod-visible mount retains "ro". Safe because the temp mount is never exposed to pods and NAS server-side ACLs still apply.

5. **`ValidateVolumeCapabilities` accepts `SINGLE_NODE_WRITER`**
   Intentional — NAS can serve `ReadWriteOnce` PVCs (the CO enforces the single-node constraint). The newer modes (`SINGLE_NODE_SINGLE_WRITER`, `SINGLE_NODE_MULTI_WRITER`) are not listed because Kubernetes only sends those to drivers that advertise the corresponding capability.

```release-note
NAS driver now respects readonly volume access modes and the readonly field in NodePublishVolume requests.
```

```docs
NONE
```

### DIFF
--- a/pkg/nas/nodeserver.go
+++ b/pkg/nas/nodeserver.go
@@ -306,6 +306,15 @@ func (ns *nodeServer) NodePublishVolume(ctx context.Context, req *csi.NodePublis
 		}
 	}
 
+	readOnly := req.GetReadonly()
+	if !readOnly {
+		switch req.GetVolumeCapability().GetAccessMode().GetMode() {
+		case csi.VolumeCapability_AccessMode_MULTI_NODE_READER_ONLY,
+			csi.VolumeCapability_AccessMode_SINGLE_NODE_READER_ONLY:
+			readOnly = true
+		}
+	}
+
 	var runtimeVal string
 	if ns.config.KubeClient != nil {
 		runtimeVal = utils.GetPodRunTime(ctx, req, ns.config.KubeClient)
@@ -314,8 +323,12 @@ func (ns *nodeServer) NodePublishVolume(ctx context.Context, req *csi.NodePublis
 	// running in runc/runv mode
 	if runtimeVal == utils.RunvRunTimeTag {
 		fileName := filepath.Join(mountPath, utils.CsiPluginRunTimeFlagFile)
+		options := opt.Options
+		if readOnly {
+			options = append(options, "ro")
+		}
 		runvOptions := RunvNasOptions{
-			Options:    strings.Join(opt.Options, ","),
+			Options:    strings.Join(options, ","),
 			Server:     opt.Server,
 			ModeType:   opt.ModeType,
 			Mode:       opt.Mode,
@@ -370,6 +383,10 @@ func (ns *nodeServer) NodePublishVolume(ctx context.Context, req *csi.NodePublis
 		}
 	} else if len(opt.Options) == 1 && strings.EqualFold(opt.Options[0], "none") {
 		opt.Options = nil
+	}
+
+	if readOnly {
+		opt.Options = append(opt.Options, "ro")
 	}
 
 	notMounted, err := ns.mounter.IsLikelyNotMountPoint(mountPath)
@@ -433,6 +450,9 @@ func (ns *nodeServer) NodePublishVolume(ctx context.Context, req *csi.NodePublis
 
 	// do losetup nas logical
 	if ns.config.EnableLosetup && opt.MountType == LosetupType {
+		if readOnly {
+			return nil, status.Error(codes.InvalidArgument, "losetup volumes do not support readonly mode")
+		}
 		if err := ns.mountLosetupPv(mountPath, opt, req.VolumeId); err != nil {
 			klog.Errorf("NodePublishVolume: mount losetup volume(%s) error %s", req.VolumeId, err.Error())
 			return nil, errors.New("NodePublishVolume, mount Losetup volume error with: " + err.Error())
@@ -477,7 +497,7 @@ func (ns *nodeServer) NodePublishVolume(ctx context.Context, req *csi.NodePublis
 
 	}
 	// change the mode
-	if opt.Mode != "" && opt.Path != "/" {
+	if opt.Mode != "" && opt.Path != "/" && !readOnly {
 		var args []string
 		if opt.ModeType == "recursive" {
 			args = append(args, "-R")
--- a/pkg/nas/nodeserver.go
+++ b/pkg/nas/nodeserver.go
@@ -37,6 +37,7 @@ import (
 	"github.com/kubernetes-sigs/alibaba-cloud-csi-driver/pkg/features"
 	"github.com/kubernetes-sigs/alibaba-cloud-csi-driver/pkg/losetup"
 	"github.com/kubernetes-sigs/alibaba-cloud-csi-driver/pkg/mounter"
+	mounterutils "github.com/kubernetes-sigs/alibaba-cloud-csi-driver/pkg/mounter/utils"
 	"github.com/kubernetes-sigs/alibaba-cloud-csi-driver/pkg/nas/internal"
 	"github.com/kubernetes-sigs/alibaba-cloud-csi-driver/pkg/utils"
 	utilsio "github.com/kubernetes-sigs/alibaba-cloud-csi-driver/pkg/utils/io"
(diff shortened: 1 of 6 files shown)


huww98:
## E2E test results — readonly mount enforcement

Tested NAS readonly mounts on two kernel versions with two different mechanisms.

### Test 1: `ReadOnlyMany` access mode (no `volumeMount.readOnly` or `pvc.readOnly`)

PVC with `accessModes: [ReadOnlyMany]`, pod does NOT set `volumeMount.readOnly` or `persistentVolumeClaim.readOnly`. Relies solely on the CSI driver's access mode detection to append `ro`. Note: Kubernetes does not automatically enforce readonly on the container bind mount based on PVC access mode alone — that requires explicit `readOnly` fields.

| Kernel | Superblock | VFS mount (container) | Write blocked? |
|--------|-----------|----------------------|---------------|
| 5.10.134-19.3 | `ro` | `rw` (runc clears MNT_READONLY) | **Yes** — `sb_rdonly()` enforces |
| 6.6.102-5.3 | `rw` | `rw` (runc clears MNT_READONLY) | **No** — writes succeed |

### Test 2: `persistentVolumeClaim.readOnly: true` (RWX access mode)

PVC with `accessModes: [ReadWriteMany]`, pod sets `volumes[].persistentVolumeClaim.readOnly: true`. This causes kubelet to pass `readonly=true` in NodePublishVolumeRequest AND enforce readonly on the container bind mount.

| Kernel | Superblock | VFS mount (container) | Write blocked? |
|--------|-----------|----------------------|---------------|
| 5.10.134-19.3 | `ro` | `ro` | **Yes** — both sb and VFS enforce |
| 6.6.102-5.3 | `rw` | `ro` | **Yes** — VFS `MNT_READONLY` enforces (runc preserves it) |

### Test 3: Subpath auto-creation with readonly

PV with a non-existent `path` (`/subpath-ro-test-new`), `persistentVolumeClaim.readOnly: true`. Tests that the driver can create the subpath directory (requires a temporary rw root mount) and still mount the final volume as readonly.

| Kernel | Subpath created? | Final mount | Write blocked? |
|--------|-----------------|-------------|---------------|
| 5.10.134-19.3 | **Yes** | `ro` superblock + `ro` VFS | **Yes** |
| 6.6.102-5.3 | **Yes** | `rw` superblock + `ro` VFS | **Yes** |

CSI driver flow (from logs):
1. First mount attempt with `ro` to `/subpath-ro-test-new` → "access denied by server" (path doesn't exist)
2. Temporary root mount WITHOUT `ro` to private kubelet dir → `MkdirAll` creates the subpath
3. Unmount temp root
4. Final mount with `ro` to the now-existing subpath → success

### Root cause of Test 1 failure on 6.6

Kernel commit `52cb7f8f` ("nfs: ignore SB_RDONLY when mounting nfs"), backported to Anolis 6.6.102-5, explicitly clears `SB_RDONLY` from `fc->sb_flags` before creating the NFS superblock:

```c
// fs/nfs/super.c nfs_get_tree_common()
if (server->flags & NFS_MOUNT_UNSHARED)
    compare_super = NULL;
else
    fc->sb_flags &= ~SB_RDONLY;  // <-- NFS superblock is always rw
```

This means on 6.6, NFS superblocks are always `rw` regardless of mount options. The `ro` option only sets VFS-level `MNT_READONLY` on the host mount point. When the container runtime sets up the pod's bind mount without a readonly signal (Test 1 scenario), it does `mount -o remount,bind,rw` which clears `MNT_READONLY`. Since the superblock is not `SB_RDONLY`, writes go through.

In Test 2, kubelet tells the container runtime that the volume is readonly, so runc does NOT remount the bind as rw — `MNT_READONLY` is preserved and writes are blocked.

### Fix status

The reverts ([9bb9de0d2403](https://gitee.com/anolis/cloud-kernel/commit/9bb9de0d24030d932cc849b51eb90b689b460a69), [fa1e99021d41](https://gitee.com/anolis/cloud-kernel/commit/fa1e99021d4190d3f77fac2f188fdcc9ac384886), [293ec7a37c89](https://gitee.com/anolis/cloud-kernel/commit/293ec7a37c89e173ef25d7130e74f10220c10f96)) are present on [`anolis/release/release-6.6.102-6.y`](https://gitee.com/anolis/cloud-kernel/tree/release/release-6.6.102-6.y) (`6.6.102-6_rc1` tag) but NOT on the `5.3.y` or `5.4.y` branches.

### Conclusion

- The CSI driver correctly appends `ro` to mount options for both `req.Readonly` and reader-only access modes
- `persistentVolumeClaim.readOnly: true` provides reliable enforcement on ALL kernel versions (kubelet enforces readonly on the container bind mount)
- `ReadOnlyMany` access mode alone is insufficient on 6.6.102-5.x kernels due to the NFS superblock change — will be resolved in 6.6.102-6
- For users on 6.6 kernels who need strict readonly, recommend using `persistentVolumeClaim.readOnly: true` or `volumeMount.readOnly: true`

pkg/nas/nodeserver.go
```diff
@@ -310,8 +323,12 @@ func (ns *nodeServer) NodePublishVolume(ctx context.Context, req *csi.NodePublis
 	// running in runc/runv mode
 	if runtimeVal == utils.RunvRunTimeTag {
 		fileName := filepath.Join(mountPath, utils.CsiPluginRunTimeFlagFile)
```
mowangdk, line 325: `options := opt.Options` copies the slice header but shares the backing array. If `opt.Options` has spare capacity, `append` on the next line will mutate the backing array. The function returns early in this code path so it's not a current bug, but it's a latent mutation hazard.

Since `slices` is already imported elsewhere:
```go
options := slices.Clone(opt.Options)
```

pkg/nas/nodeserver.go
```diff
@@ -429,6 +450,9 @@ func (ns *nodeServer) NodePublishVolume(ctx context.Context, req *csi.NodePublis
 
 	// do losetup nas logical
 	if ns.config.EnableLosetup && opt.MountType == LosetupType {
```
mowangdk, line 452: Nit: this readonly rejection happens after directory creation, mount point checks, and default option setup. Moving it earlier (right after `readOnly` is determined, around line 315) would avoid unnecessary work when the request will be rejected anyway.

mowangdk (commented):
## Code Review

**Summary:** Well-structured PR with a clean two-commit split. Refactors `Options.Options` from comma-separated string to `[]string`, then adds proper readonly mount support via `req.Readonly` and `READER_ONLY` access modes.

**Overall:** Solid change with excellent PR description. Two items I'd prioritize: a latent slice aliasing hazard in the runv path, and missing `NodePublishVolume` test coverage for the new readonly logic — the core feature (checking `req.Readonly`, inferring readonly from access mode, appending "ro", skipping chmod) isn't exercised by any test.

pkg/nas/utils.go
```diff
@@ -173,11 +171,15 @@ func doMount(m mounter.Mounter, opt *Options, targetPath, volumeId, podUid strin
 		return err
 	}
 	defer os.Remove(tmpPath)
+	// mount without "ro" since we need to create the subpath directory
```
mowangdk, line 174: Nice catch

pkg/nas/nodeserver.go
huww98: This is common for Go code. Since this is correct now. I don't want to clone it.

pkg/nas/nodeserver.go
huww98: very few user use EnableLosetup, let's just keep it, this makes code more coherent.

PR #1715, opened 2026-06-05
feat(fuse_pod_manager): add configurable metrics mode enabling logic

AlbeeSo:
#### What type of PR is this?

/kind feature
Allow disable metrics-mode by env.

Fixes #

```release-note

```

```docs

```

### DIFF
--- a/pkg/mounter/fuse_pod_manager/fuse_pod_manager.go
+++ b/pkg/mounter/fuse_pod_manager/fuse_pod_manager.go
@@ -6,6 +6,7 @@ import (
 	"errors"
 	"fmt"
 	"maps"
+	"os"
 	"strings"
 	"time"
 
@@ -75,6 +76,13 @@ const (
 	MetricsModeDisabled = "disabled"
 	MetricsModeEnabled  = "enabled"
 )
+
+func ShouldEnableMetrics(configMetricsMode string) bool {
+	if configMetricsMode != "" {
+		return configMetricsMode != MetricsModeDisabled
+	}
+	return os.Getenv("METRICS_MODE") != MetricsModeDisabled
+}
 
 type FusePodContext struct {
 	context.Context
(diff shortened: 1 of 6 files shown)


mowangdk (commented):
## Review

### Summary

This PR adds support for disabling metrics mode via the `METRICS_MODE` environment variable, complementing the existing configmap-based approach. The env var acts as a fallback when no configmap value is set.

### Logic

```
ShouldEnableMetrics(configMetricsMode):
  if configMetricsMode != "" → return configMetricsMode != "disabled"  (config wins)
  else → return os.Getenv("METRICS_MODE") != "disabled"  (env fallback, default: enabled)
```

### Positives

- Clean abstraction — `ShouldEnableMetrics` consolidates the decision logic in one place
- Config takes priority over env, which is the expected precedence
- Tests cover the priority/override behavior well (config overrides env)
- Applied consistently in both ossfs and ossfs2

### Issues

**1. Any non-"disabled" value enables metrics (by design, but worth noting)**

`ShouldEnableMetrics` returns `true` for `METRICS_MODE=foo` or any typo. This is consistent with the previous inline check (`config.MetricsMode != MetricsModeDisabled`), so it's not a regression. But consider whether a stricter check (only enable for empty or "enabled") would be more defensive.

**2. No release-note**

This introduces a new user-facing env var (`METRICS_MODE`) but the release-note block is empty. Operators need to know this configuration option exists.

**3. `fakeOssfs.config.MetricsMode` not reset between subtests in ossfs test**

In `TestAddDefaultMountOptions_ossfs`, existing test cases that don't set `metricsMode` will get empty string (Go zero value), which works. But the test relies on implicit zero-value reset. Adding an explicit reset or using a fresh config per subtest would make the tests more robust against future refactoring.

**4. Minor: test env var not explicitly unset for "no env" cases**

Same pattern as other PRs in this repo — when `tt.metricsEnv` is empty, `t.Setenv` is not called. If `METRICS_MODE` is set in the CI environment, the "no config, no env, default enabled" test case could fail. Low probability but easy to guard against.

### Verdict

Small, focused change with good test coverage. The main gap is documentation — add a release note describing the new `METRICS_MODE` env var and its interaction with the configmap setting.

PR #1722, opened 2026-06-15
feat(customfuse): add generic FUSE driver for RunC   and sandbox scenarios

AlbeeSo:
#### What type of PR is this?

/feature
## Summary

This PR introduces the **customfuse** CSI driver, a generic framework for running any FUSE client as a managed pod in both RunC container and sandbox scenarios. Users supply the FUSE client image and entrypoint script; the driver handles pod lifecycle, mount propagation, and credential injection.

### 1. Generic FUSE Driver Architecture
- Supports any FUSE client (JuiceFS, s3fs, JindoFS, etc.) via configurable image + entrypoint
- Configuration flows from PV `volumeAttributes` + `Secret` into entrypoint.sh as env vars
- Optional ConfigMap-injected entrypoint for iterating mount logic without image rebuilds

### 2. Sandbox Support
- Add `--mount-proxy-sock` and `--customfuse-mount-proxy-sock` CLI flags as default mount-proxy socket paths for csi-agent mode
- `SKIP_GLOBAL_MOUNT` env var to skip globalmount and mount directly to target path
- csi-agent integration via `customfuse.NewCSIAgent()`

## Examples: JuiceFS CE + Alibaba Cloud OSS

Four demo patterns showing different configuration approaches:

| Demo                    | Description                                               |
| ----------------------- | --------------------------------------------------------- |
| `1-baked-in`            | All config hardcoded in image ENV                         |
| `2-standard`            | Config from PV volumeAttributes + Secret (recommended)    |
| `3-advanced-entrypoint` | Custom resource controls: cache quota, credential cleanup |
| `4-configmap`           | Entrypoint via ConfigMap — iterate without image rebuild  |

Each demo includes Dockerfile, entrypoint.sh, and PV/PVC/Secret YAML.

The README also includes quick-reference snippets for adapting to other FUSE clients (s3fs, JindoFS).

## New CLI Flags

| Flag                            | Scope                          | Purpose                                                              |
| ------------------------------- | ------------------------------ | -------------------------------------------------------------------- |
| `--mount-proxy-sock`            | csi-agent / OSS / NAS / BMCPFS | Default mount-proxy socket path (fallback when no ControllerPublish) |
| `--customfuse-mount-proxy-sock` | csi-agent / customfuse         | Default mount-proxy socket path (fallback when no ControllerPublish) |

## Breaking Changes

None. customfuse is a new driver and does not affect existing OSS/NAS/BMCPFS drivers.

Fixes #

```release-note

```

```docs

```


deploy/charts/alibaba-cloud-csi-driver/templates/controller.yaml
```diff
@@ -551,6 +551,75 @@ spec:
             - name: bmcpfs-provisioner-dir
               mountPath: /csi
 {{- end -}}
+{{- if and .Values.csi.customfuse.enabled .Values.csi.customfuse.controller.enabled }}
```
mowangdk: We don’t want CustomFuse to be integrated with our default driver in the same provisioner; it must remain an independent component.

examples/customfuse/1-baked-in/Dockerfile
```diff
@@ -0,0 +1,74 @@
...
+    --mount=type=cache,target=/root/.cache/go-build \
+    export GOOS=$TARGETOS && \
+    export GOARCH=$TARGETARCH && \
+    export CGO_ENABLED=0 && \
+    go build -o /out/csi-mount-proxy-server ./cmd/mount-proxy-server && \
+    go build -o /out/csi-mount-proxy-client ./cmd/mount-proxy-client
+
+# Stage 2: assemble the final image on top of the FUSE client.
+# TODO: replace with actual JuiceFS image tag
+FROM juicedata/mount:ce-v1.2.0
```
mowangdk, line 25: Can we obtain this image from our CI/CD environment?

pkg/mounter/utils/utils.go
```diff
@@ -39,8 +40,33 @@ const (
...
+// generated by GetMountProxySocketPath — this is the authoritative source in the
+// normal csi-plugin flow.
+//
+// overrideSock is the nodeServer.mountProxySock field, set by:
+//   - csi-plugin: --mount-proxy-sock / --customfuse-mount-proxy-sock flag
+//   - csi-agent: NewCSIAgent constructor
+//
+// In sandbox agent scenarios there is no ControllerPublish, so publishContext is
+// empty and overrideSock provides the socket path directly as fallback.
+func ResolveMountProxySocket(publishContext map[string]string, overrideSock string) string {
```
mowangdk, line 58: Should the mountProxy arguments have higher priority?

AlbeeSo: Global setting should be default and the individual one should have a higher priority.

examples/customfuse/1-baked-in/Dockerfile
AlbeeSo: It's available in dockerHub.

examples/customfuse/1-baked-in/entrypoint.sh
```diff
@@ -0,0 +1,28 @@
...
+    "$META_URL" \
+    myjfs
+
+# foreground: required by mount-proxy; no-update: skip version check (pinned by image).
+MOUNT_OPTS="foreground,no-update,cache-size=$CACHE_SIZE"
+[ "$readOnly" = "true" ] && MOUNT_OPTS="${MOUNT_OPTS},ro"
+
+if [ -n "$CAPACITY" ]; then
+    echo "Setting quota: capacity=${CAPACITY}GiB"
+    juicefs quota set "$META_URL" --path / --capacity "$CAPACITY"
```
mowangdk, line 24: The Dockerfile and entrypoint should be generalized as templates rather than being specific to JuiceFS. We can add a script to render the template into JindoFS version.

pkg/customfuse/controllerserver.go
```diff
@@ -0,0 +1,134 @@
...
+	client         kubernetes.Interface
+	metadata       metadata.MetadataProvider
+	fusePodManager *fpm.FusePodManager
+	common.GenericControllerServer
+}
+
+func (*controllerServer) ControllerGetCapabilities(context.Context, *csi.ControllerGetCapabilitiesRequest) (*csi.ControllerGetCapabilitiesResponse, error) {
+	return &csi.ControllerGetCapabilitiesResponse{
+		Capabilities: common.ControllerRPCCapabilities(
+			csi.ControllerServiceCapability_RPC_CREATE_DELETE_VOLUME,
```
mowangdk, line 33: We didn't implements the CreateVolume & DeleteVolume in this driver

pkg/features/features.go
```diff
@@ -72,6 +72,20 @@ const (
 	// instance type does not support high density incur no extra call. When disabled,
 	// the behavior is byte-identical to before (no extra API call).
 	DiskHighDensityMode featuregate.Feature = "DiskHighDensityMode"
+
+	// CustomFuseAutoCapacity enables the customfuse controller to read
+	// pv.spec.capacity.storage via K8s API and pass it as $capacity env var
+	// to the FUSE entrypoint.
```
mowangdk: What is the purpose of the CustomFuseAutoCapacity feature? It seems a bit odd to add capacity messages to the context. and i'm not sure if this specific to juicefs

mowangdk (commented):
Additional findings from a code review pass (complementing the earlier inline comments):

### 1. Node capability vs. behavior mismatch (STAGE_UNSTAGE)
`NodeGetCapabilities` advertises `STAGE_UNSTAGE_VOLUME`, but `NodeStageVolume` is a no-op — the actual global mount to `attachPath` happens in `NodePublishVolume`, while `NodeUnstageVolume` cleans up `attachPath`. This is asymmetric: the shared mount is created in Publish but torn down in Unstage. Recommend either performing the global mount in `NodeStageVolume` (proper CSI semantics) or dropping the STAGE_UNSTAGE capability and cleaning `attachPath` in Unpublish.

### 2. Security: privileged fuse pods + secret-as-env passthrough
`buildPodSpec` sets `Privileged: true` and `HostNetwork: true`, and `buildEnvVars` passes **all** secret entries as environment variables to a user-supplied entrypoint. Secrets in env are visible via `/proc/<pid>/environ` to anything in the host PID namespace; combined with privileged + hostNetwork this is a meaningful exposure. Please document the threat model prominently. The `akId→accessKeyId` / `akSecret→accessKeySecret` auto-aliasing also silently duplicates credentials into more env vars — worth a doc note.

### 3. `otherOpts` command-injection surface — validation only lives in examples
The commit message claims "validate each argument starts with `--` to prevent command injection," but that validation (`validate_opts`) lives only in the example entrypoints, not in the driver. `otherOpts` flows verbatim into `$otherOpts` and is expanded by the customer-authored entrypoint's shell, so injection safety is entirely the entrypoint author's responsibility. That's defensible, but the driver-side docs/comments should not imply the driver enforces it. Consider rejecting `otherOpts` containing newline / `;` / `$()` in `parseOptions` as defense-in-depth.

### 4. `hasUnitSuffix` is fragile
`hasUnitSuffix` returns true if the last rune isn't a digit. This works for `100Gi`/`100`, but the ad-hoc suffix check is easy to get subtly wrong. Consider simply running `resource.ParseQuantity` whenever the value is non-integer, rather than the custom suffix heuristic.

### 5. Capacity precedence split across controller and node paths
`resolveAutoCapacity` (controller) checks `req.GetVolumeContext()["capacity"]`, while `parseOptions` (node) also merges `PublishContext["capacity"]`. The precedence between `volumeAttributes.capacity`, `PublishContext` capacity, and auto-capacity is split across two files and hard to follow. A single documented precedence table (and a test) would help.

### 6. ConfigMap informer lifecycle
The single-ConfigMap field-selector informer is low overhead — good. Please double-check that the informer's `DeleteFunc` clearing `configMapData` to nil does not cause in-flight mounts to lose a valid config; the fallback to `defaultConfig` looks correct, just calling it out.

### Nits
- `mountProxySocket` const in `controllerserver.go` duplicates `mounterutils.MountProxySocketKey`.
- The intentional wire-format typo `mountPorxySocket` is preserved (documented) — ugly to carry forward but understood.
- Several `klog.Infof` calls at default verbosity in the hot Publish path; consider `V(2)` for success lines to reduce log spam at scale.

deploy/charts/alibaba-cloud-csi-driver/templates/controller.yaml
AlbeeSo: Done — CustomFuse ships as its own component. The shared provisioner loop now skips it explicitly (`templates/controller.yaml:839`, `(ne $key "customfuse")`), and it gets dedicated workloads in `templates/customfuse-controller.yaml` and `templates/customfuse-plugin.yaml`, with their own ServiceAccounts (`csi-customfuse-controller`, `csi-customfuse-node`) rather than sharing the other drivers'.

`csi.customfuse.enabled` is off in every values file, so nothing renders unless it is opted into. `hack/check-helm.sh` validates both the standalone and the alongside-other-drivers configurations.

examples/customfuse/1-baked-in/entrypoint.sh
AlbeeSo: I'd push back on this one. The Dockerfile and the entrypoint *are* the compatibility layer — they are the part the customer writes, because only they know how their FUSE client is invoked, what its options mean, and how it wants to be given a credential. If that layer could be rendered from a template, the logic would belong in the driver instead, and there would be nothing left to bring your own.

So the examples are deliberately concrete rather than generic: one real client end to end, so the shape is copyable. The driver side is already client-agnostic — juicefs appears in non-test Go code in three illustrative comments only (`pkg/customfuse/options.go:18`, `:35`, `pkg/mounter/proxy/server/customfuse/driver.go:221`) and nowhere in the chart.

On the JindoFS suggestion specifically: `examples/customfuse/README.md` has a "JindoFS example" section showing the same four steps. It differs from the JuiceFS one in exactly the two things that cannot be templated — the install stage and the mount command.

pkg/customfuse/controllerserver.go
AlbeeSo: Both are implemented now: `CreateVolume` at `controllerserver.go:109`, `DeleteVolume` at `:139`.

Neither provisions anything — a customfuse volume is whatever the customer's entrypoint mounts, which the driver does not model. But `CreateVolume` is where the CSI protocol hands over the requested size, so it converts `capacity_range` into the volume context; a dynamically provisioned volume then needs no capacity restated anywhere. Because nothing is provisioned, a reclaim policy promising deletion is rejected rather than accepted and quietly ignored, and `DeleteVolume` succeeds without deprovisioning.

pkg/features/features.go
AlbeeSo: The gate is gone — your instinct that it was odd was right.

It existed because the controller had to `Get` the PV to read `spec.capacity.storage`, and `ControllerPublishVolume` never receives the PV name, only the volumeHandle. So it used volumeHandle as the PV name, which holds only when the two happen to be equal, and it cost an API call per publish. Gating that did not make it sound.

Capacity now travels the standard CSI route instead: `CreateVolume` returns `CapacityBytes` and sets `capacity` in `VolumeContext`, external-provisioner writes that into the PV, and `NodePublishVolume` reads it back from `req.VolumeContext`. No API call, no assumption about the PV name, and a single source for capacity rather than a precedence split across the controller and node paths.

On whether it is juicefs-specific: nothing in the path is. The driver validates that the value parses as a `resource.Quantity` and passes the string through as `$capacity`. What a capacity means, and whether to enforce it, is the entrypoint's decision — the examples show one client's interpretation, the driver has none.

AlbeeSo (commented):
Thanks for the thorough pass — all six findings and the nits are addressed or answered below. Head is now `789e43f1`.

**1. STAGE_UNSTAGE vs. behaviour.** This mirrors the oss driver line for line: oss advertises the capability at `pkg/oss/nodeserver.go:76`, no-ops `NodeStageVolume` at `:388`, does the global mount inside `NodePublishVolume` at `:306`, and tears `attachPath` down in `NodeUnstageVolume` at `:395`/`:405`. customfuse does the same at `:38` / `:156` / `:105` / `:160` / `:167`. I agree the split is not ideal CSI semantics, but moving it for customfuse alone would leave it the only fuse driver here with different stage behaviour, and oss's Unstage carries an extra legacy-cleanup block (`:414+`) that would have to move with it. I'd rather fix both drivers in one follow-up than introduce the asymmetry here — happy to file an issue so it is not lost.

**2. Privileged fuse pods + secrets as env.** `docs/customfuse.md` now has a `## Security model` section stating plainly that mounting requires a privileged container, that every Secret key becomes an environment variable because the driver cannot know how an arbitrary client wants its credential, who can read it, and that an entrypoint echoing its environment will write credentials to the pod log. The `akId`→`accessKeyId` / `akSecret`→`accessKeySecret` aliasing is documented there too, including the fact that both spellings end up present. It also covers unsetting the variables before `exec`ing the client.

That said, documenting the exposure is not the end state — it is what this PR can honestly claim, and the follow-up removes it at the root. `authType` already reserves both alternatives: `rrsa` and `agent-identity` are named in `options.go:54` and `:234-235`, and `precheckAuthConfig` (`:224`) rejects them today with an explicit error rather than silently falling back to secret passthrough.

The split is by deployment path, and it is a PR boundary rather than an open question:

- **agent-identity** for the sandbox path — the credential is exchanged per mount, delivered as files and masked mount options instead of environment variables, and rotated before it expires. NAS already does exactly this and it is documented upstream in `docs/nas-agent-identity.md`, so the mechanism is established in this repo; customfuse follows the same shape.
- **RRSA** for the runc path, via mount-proxy.

One piece of the RRSA path is already in this PR: the `serviceAccountName` volume attribute added for private-registry pulls is spelled as in oss (`pkg/oss/utils.go:211`), where the same field carries the RRSA ServiceAccount, and oss overrides it from `authCfg.RrsaConfig.ServiceAccountName` (`ossfs/manager.go:439`, `ossfs2/manager.go:373`). So the follow-up plugs into an existing seam instead of adding a new one.

I left the agent-identity documentation out of this PR deliberately — an earlier revision of `docs/customfuse.md` described it and linked a document that is not in this tree, which would have pointed readers at an `authType` that errors out. It comes back with the code that implements it.

**3. `otherOpts` injection.** Taken — the driver now refuses `\n`, `\r`, `;`, a backtick and `$(` in `otherOpts` (`pkg/customfuse/options.go:102`). One more than you named: rejecting `$(` but not a backtick would be a half-measure that reads as a guarantee. It checks `pv.spec.mountOptions` entries as well, because `buildEnvVars` maps both through the same path, so an `otherOpts=<payload>` mount flag lands in `$otherOpts` too — refusing only the attribute would just move the payload.

Deliberately *not* checked: `source`, `path` and Secret values, which can legitimately contain these characters, so quoting them stays the entrypoint's job. Also deliberately not policed: the remaining metacharacters (`|`, `&`, `>`). The docs call this "a floor, not a validation of your options", and `TestParseOptions_LeavesRemainingMetacharsToTheEntrypoint` pins the boundary so widening it later is a visible policy change rather than a quiet one.

**4. `hasUnitSuffix`.** Gone, replaced by `resource.ParseQuantity` as you suggested. The old heuristic indexed a byte as a rune and let a bare integer past unvalidated; now every non-empty value has to parse (`options.go:150-159`). `TestParseOptions_CapacityPassthrough` covers plain integers, `Gi`/`Ti`/`Mi`, an invalid unit and a non-numeric value.

**5. Capacity precedence.** The table is now a doc comment on the `Capacity` field itself (`options.go:56-65`), and the split you flagged is gone with the removal of `resolveAutoCapacity` — see the reply on `pkg/features/features.go`. Each leg is tested: `TestCreateVolumeTurnsRequestedSizeIntoCapacity`, `TestCreateVolumeWithoutCapacityRange`, `TestCreateVolumeKeepsExplicitCapacityParameter`, and `TestMakeMountOptionsLetsMountOptionsOverrideAttributes` for the node-side override.

**6. Informer `DeleteFunc`.** Checked, and the fallback is load-bearing rather than incidental: `resolveConfig` returns `defaultConfig` when `configMapData` is nil, and `defaultConfig` is parsed once at construction from the ConfigMap the process started with — so a deleted ConfigMap degrades to the startup image, never to an empty config that would fail a mount in flight. `TestResolveConfig` asserts exactly that path (`clearConfigMap()` → `defaultConfig`).

Related, since it is the same informer: the initial sync is now bounded at 30s instead of waiting on the caller's context. A list that RBAC will never authorize does not start succeeding, so the reflector only retries, and the controller would previously have hung before serving anything. It now exits naming the cause. `docs/customfuse.md` documents the Kubernetes 1.32 floor behind it — `resourceNames` scopes a list or a watch only from 1.32, where `AuthorizeWithSelectors` defaulted on — and the chart comment points at that section.

**Nit: `mountProxySocket` duplication.** customfuse has no local const; it uses `mounterutils.MountProxySocketKey` directly (`pkg/customfuse/controllerserver.go:75`). While there I also collapsed oss's hardcoded literal — `pkg/oss/controllerserver.go:45` was `mountProxySocket = "mountPorxySocket"`, now `= mounterutils.MountProxySocketKey`. So the wire-format spelling has a single definition at `pkg/mounter/utils/utils.go:46`.

**Nit: the preserved `mountPorxySocket` typo.** Understood, and it is now contained to that one line rather than repeated.

**Nit: log verbosity in the Publish path.** Most are now `V(2)`/`V(4)` (`nodeserver.go:46`, `:62`, `:101`, `:125`, `:136`, `:142`, `:152`). Three stay at default and I'd argue for keeping them: the two "already mounted" idempotency returns (`:85`, `:129`) and `NodeUnstageVolume` (`:161`). oss keeps the same three at default in the same places — it has three "already mounted" lines at `:178`/`:212`/`:251` and an equivalent Unstage log — and customfuse is now quieter than oss at the points they share, since oss logs "Starting Mount volume" (`:117`) and "successfully mounted volume" (`:336`) at default where customfuse uses `V(2)`. The controller-side lines match oss's `:124`/`:159`/`:164`/`:219` verbatim. Say the word if you'd rather have them all at `V(2)` and I'll drop them plus oss's for consistency.

mowangdk (changes requested):
Reviewed commit `789e43f141541aae0cebbe67bb2bcfc202f441a2`, focusing on whether the FUSE abstraction is genuinely reusable across clients. This supplements the earlier review and the responses to it.

**Request changes.** The image-per-fuseType and client-owned adapter approach is a sound starting point, but several OSS assumptions still cross the abstraction boundary, and there are concrete correctness issues to address before merging.

## Must fix

### 1. [P1] Opaque mount sources can leak credentials through metrics

[pkg/customfuse/nodeserver.go:88,112](https://github.com/kubernetes-sigs/alibaba-cloud-csi-driver/blob/789e43f141541aae0cebbe67bb2bcfc202f441a2/pkg/customfuse/nodeserver.go#L88) passes `opts.Source` as the filesystem identifier to the metrics helpers. [pkg/utils/metrics.go:158-175](https://github.com/kubernetes-sigs/alibaba-cloud-csi-driver/blob/789e43f141541aae0cebbe67bb2bcfc202f441a2/pkg/utils/metrics.go#L158) writes that value into the metadata file, and the collector uses it as `BucketName`/the `bucket_name` metric label.

A bucket name is normally a reasonable public identifier; an arbitrary FUSE source is not. For example, a JuiceFS source can be `redis://user:password@host/1`, exposing credentials to metric files and downstream monitoring. The V(4) option logs also print the source.

Please treat the source as opaque and potentially sensitive. Use the volume ID or a separately defined non-sensitive identifier for metrics, and avoid logging the raw source. This is separate from the already documented Secret-as-environment exposure.

### 2. [P1] The production writeback example has no durable cache volume

[examples/customfuse/4-configmap/configmap-production.yaml:29-31](https://github.com/kubernetes-sigs/alibaba-cloud-csi-driver/blob/789e43f141541aae0cebbe67bb2bcfc202f441a2/examples/customfuse/4-configmap/configmap-production.yaml#L29) enables `writeback,cache-dir=/var/jfsCache`, but [pkg/mounter/fuse_pod_manager/customfuse/manager.go:188-232](https://github.com/kubernetes-sigs/alibaba-cloud-csi-driver/blob/789e43f141541aae0cebbe67bb2bcfc202f441a2/pkg/mounter/fuse_pod_manager/customfuse/manager.go#L188) mounts only the target, metrics and optional entrypoint. The cache is therefore in the container writable layer.

Once the example script errors below are fixed, replacing the container before asynchronous uploads complete can discard acknowledged writes. JuiceFS writeback commits metadata before uploading the locally cached data; this is not merely a cache-performance issue.

Please remove writeback from the shipped examples until durable cache storage and recovery are demonstrated. More generally, a reusable FUSE runtime needs an administrator-controlled way to provide client configuration files and state/cache volumes, not just one executable ConfigMap key.

### 3. [P1] Windows main-binary compilation is broken

[pkg/bmcpfs/bmcpfs_windows.go:9](https://github.com/kubernetes-sigs/alibaba-cloud-csi-driver/blob/789e43f141541aae0cebbe67bb2bcfc202f441a2/pkg/bmcpfs/bmcpfs_windows.go#L9) adds a fourth constructor argument, while [main.go:326](https://github.com/kubernetes-sigs/alibaba-cloud-csi-driver/blob/789e43f141541aae0cebbe67bb2bcfc202f441a2/main.go#L326) still passes three. The non-Windows constructor also remains three-argument.

Locally reproduced with a Windows cross-build of the root package: `not enough arguments in call to bmcpfs.NewServers`. Please remove the unrelated argument addition to the Windows stub, or consistently implement the intended change.

### 4. [P1] Disabling or uninstalling the chart deletes live FUSE pods

[deploy/charts/alibaba-cloud-csi-driver/templates/customfuse-rbac.yaml:1-7](https://github.com/kubernetes-sigs/alibaba-cloud-csi-driver/blob/789e43f141541aae0cebbe67bb2bcfc202f441a2/deploy/charts/alibaba-cloud-csi-driver/templates/customfuse-rbac.yaml#L1) creates the FUSE namespace without `helm.sh/resource-policy: keep`. Removing that resource on uninstall, or an upgrade setting `csi.customfuse.enabled=false`, deletes the namespace and its on-demand FUSE pods and user configuration, disrupting existing consumer mounts.

Please retain the namespace, consistent with the existing protection in `templates/fuse.yaml:5-11` for OSS.

## Correctness and genericity

### 5. [P2] Arbitrary volumeAttributes are silently dropped

[pkg/customfuse/options.go:124-171](https://github.com/kubernetes-sigs/alibaba-cloud-csi-driver/blob/789e43f141541aae0cebbe67bb2bcfc202f441a2/pkg/customfuse/options.go#L124) recognizes a fixed set of fields and discards the rest. Client-specific attributes such as `region`, `configFile` or `formatOptions` do not reach the entrypoint, contrary to the README's promise that all non-control volumeAttributes are passed through.

Routing them through `mountOptions` is a workaround, not an equivalent contract: a client configuration map and an ordered mount-flag list are different things. Flattening flags into environment variables also does not preserve repeated keys as a list.

Please parse only driver control fields and preserve client configuration separately, retaining mount flags as an array. OSS-specific conventions such as constructing `bucket:path` and aliasing AK/SK keys belong in the OSS-compatible adapter rather than the generic Go layer.

### 6. [P2] Client options and Secrets can overwrite driver-owned variables

[pkg/mounter/proxy/server/customfuse/driver.go:232-253](https://github.com/kubernetes-sigs/alibaba-cloud-csi-driver/blob/789e43f141541aae0cebbe67bb2bcfc202f441a2/pkg/mounter/proxy/server/customfuse/driver.go#L232) appends options and then Secrets after driver-generated variables. With last-wins environment semantics, these can overwrite `mountpoint` and `readOnly`, as well as inherited process environment such as `PATH`.

For example, a Secret containing `readOnly=false` overrides a read-only request; a `mountpoint` collision makes the adapter mount at a different target while the driver waits for the requested one.

Please separate driver-owned fields, client configuration and credentials. A structured request plus restricted-permission credential files is one option. If environment transport is retained, at minimum protect reserved variables and explicitly define client environment mappings.

### 7. [P2] Shared-volume read-only behavior depends on publication order

[pkg/customfuse/nodeserver.go:105-134](https://github.com/kubernetes-sigs/alibaba-cloud-csi-driver/blob/789e43f141541aae0cebbe67bb2bcfc202f441a2/pkg/customfuse/nodeserver.go#L105) uses the first NodePublish request's options to establish the shared FUSE mount, then publishes subsequent consumers with only `bind`.

If a read-only consumer of an RWX PVC publishes first, an adapter honoring `$readOnly` creates a read-only shared mount. A later read-write consumer receives a successful publish response but cannot write. Conversely, the driver does not make the per-consumer CSI bind target read-only when the shared mount is writable; container-runtime restrictions may mask that direction.

Please distinguish shared-volume access from per-consumer read-only publication, enforce the latter on the bind mount, and test both publication orders. This is a concrete issue separate from the earlier Stage/Publish placement discussion.

### 8. [P2] The second-client adaptation recipe copies the wrong script; shipped scripts also fail

The README tells users to copy an example directory and change the client image and entrypoint. However, [examples/customfuse/2-oss-compatible/Dockerfile:28](https://github.com/kubernetes-sigs/alibaba-cloud-csi-driver/blob/789e43f141541aae0cebbe67bb2bcfc202f441a2/examples/customfuse/2-oss-compatible/Dockerfile#L28) still copies the original JuiceFS entrypoint from its fixed repository path. Following the s3fs/JindoFS recipe therefore packages the wrong adapter.

Additional executable-example defects:

- `bash -n examples/customfuse/2-oss-compatible/entrypoint.sh` fails at line 26, and the same check for `3-standard/entrypoint.sh` fails at line 16: the literal regex contains an unescaped semicolon and produces `syntax error in conditional expression`.
- Both ConfigMap scripts put spaces and a comment after the continuation backslash on the `--trash-days` line. This does not continue the command, so `juicefs format` does not receive its source/name arguments.

Please parameterize the adapter COPY path, fix the scripts, and add shell syntax/argument tests. At least two complete, tested client adapters would demonstrate genericity more convincingly than four configuration styles for JuiceFS.

## Recommended abstraction boundary

The goal should be to generalize the integration contract and runtime plumbing, not to teach Go every client's CLI. No large plugin framework is necessary.

| Layer | Responsibilities |
| --- | --- |
| Generic CSI / mount-proxy | Mount target, per-consumer read-only semantics, process lifecycle, timeout, unmount, mount health, non-sensitive metrics |
| Administrator-controlled backend profile | Image, entrypoint, configuration/credential file projections, cache/state volumes, resources and necessary runtime settings |
| Client adapter | Client-specific arguments, credential format, initialization and other backend semantics |

On the earlier template discussion: keeping the actual client mount command in an adapter is correct. It does not follow that common image assembly cannot be templated. A shared Dockerfile scaffold can assemble mount-proxy, the selected client image and the selected adapter without understanding any client's command-line semantics.

Foreground execution is a reasonable initial contract; there is no need to support every daemonization mode. Require adapters to exec a foreground process, or correctly forward signals and wait for children. Also document the RunC/sandbox boundary: controller-side image/ConfigMap resolution does not automatically apply to the externally injected sandbox runtime.

Suggested acceptance criterion: adding a new client changes only its image, adapter and profile, not Go code; at least two clients pass the same mount, read-only, timeout, abnormal-exit and unmount tests.

## Verification limits

The existing test checks were passing when inspected. The Bash syntax failures and Windows compilation failure above were reproduced locally. The relevant Go unit tests could not run on the macOS review host because the packages depend on Linux-specific mount APIs; no live FUSE mount integration test was performed. The remaining findings are based on source tracing and configuration inspection.

AlbeeSo (commented):
Went through all eight. Everything below is in [`f0e9488`](https://github.com/kubernetes-sigs/alibaba-cloud-csi-driver/commit/f0e94880120adcba053854dcb861a168dbfa37c6), which replaces the previous three commits — same split, so the line links below are stable.

### 1. Opaque mount sources leaking through metrics — fixed

`Source` no longer reaches the metrics helpers. [`metricsBucketName()`](https://github.com/kubernetes-sigs/alibaba-cloud-csi-driver/blob/f0e94880120adcba053854dcb861a168dbfa37c6/pkg/customfuse/options.go#L438-L444) returns the `bucket` field, or the literal `unknown` when the volume has none, and both call sites use it ([nodeserver.go#L90](https://github.com/kubernetes-sigs/alibaba-cloud-csi-driver/blob/f0e94880120adcba053854dcb861a168dbfa37c6/pkg/customfuse/nodeserver.go#L90), [#L114](https://github.com/kubernetes-sigs/alibaba-cloud-csi-driver/blob/f0e94880120adcba053854dcb861a168dbfa37c6/pkg/customfuse/nodeserver.go#L114)). `Source` still goes into `MountOperation.Source`, which is what the mount needs, and [its field comment](https://github.com/kubernetes-sigs/alibaba-cloud-csi-driver/blob/f0e94880120adcba053854dcb861a168dbfa37c6/pkg/customfuse/options.go#L21-L32) now records why it is neither logged nor used as a label. The V(4) line at [nodeserver.go#L64](https://github.com/kubernetes-sigs/alibaba-cloud-csi-driver/blob/f0e94880120adcba053854dcb861a168dbfa37c6/pkg/customfuse/nodeserver.go#L64) prints `fuseType`, `mountOptions` and `otherOpts` only. `TestMetricsBucketNameNeverCarriesSource` asserts a credential-bearing source produces `unknown`, that `bucket` wins when both are set, and that no case contains the password.

### 2. Writeback without a durable cache volume — removed; the general mechanism is not in this PR

`writeback` is gone from every shipped example. What remains is a read cache only, and [the README says why](https://github.com/kubernetes-sigs/alibaba-cloud-csi-driver/blob/f0e94880120adcba053854dcb861a168dbfa37c6/examples/customfuse/README.md#L634-L644): anything the client writes beyond `entrypointConfig` lands in the container's writable layer, which costs re-reads for a read cache and silently loses acknowledged writes for a mode that commits metadata before uploading data.

The second half of the ask is not done. There is no administrator-controlled way to project client configuration files or state/cache volumes; the fuse pod gets the target, the metrics directory and, with `entrypointConfig`, one ConfigMap key. That is a chart-and-API surface of its own and I would rather add it as a follow-up than widen this change.

### 3. Windows main-binary compilation — fixed

`pkg/bmcpfs` is byte-identical to the merge base again. The fourth constructor argument was unrelated to this PR, so it is gone rather than propagated, and `main.go` and both constructors agree. `GOOS=windows GOARCH=amd64 go build .` passes.

### 4. Namespace deleted on uninstall — fixed

The namespace carries `helm.sh/resource-policy: keep` ([customfuse-rbac.yaml#L13-L16](https://github.com/kubernetes-sigs/alibaba-cloud-csi-driver/blob/f0e94880120adcba053854dcb861a168dbfa37c6/deploy/charts/alibaba-cloud-csi-driver/templates/customfuse-rbac.yaml#L13-L16)), matching `templates/fuse.yaml`.

### 5. volumeAttributes silently dropped — contract corrected, OSS conventions moved out; flags are still env vars

The README no longer promises pass-through. It states the actual contract at [README.md#L248-L298](https://github.com/kubernetes-sigs/alibaba-cloud-csi-driver/blob/f0e94880120adcba053854dcb861a168dbfa37c6/examples/customfuse/README.md#L248-L298): `volumeAttributes` is a whitelist, because it feeds the driver's own logic — the `bucket_name` label — so a key that logic does not know cannot be forwarded safely, and one is dropped with a log line. `spec.mountOptions` is the open channel, and exists precisely to reach an entrypoint the driver knows nothing about. The asymmetry is deliberate and the precedence table spells out each combination, including which bare entries are refused rather than passed through.

The OSS conventions are out of the Go layer. The driver composes nothing — `Source`'s comment says the format is opaque and whatever the volume said is what arrives. The `bucket:path` join and the AK/SK aliasing now live in the adapter that needs them, [2-oss-compatible/entrypoint.sh#L38-L45](https://github.com/kubernetes-sigs/alibaba-cloud-csi-driver/blob/f0e94880120adcba053854dcb861a168dbfa37c6/examples/customfuse/2-oss-compatible/entrypoint.sh#L38-L45).

What I did not do is keep mount flags as an array all the way to the entrypoint. `fuseOptions.MountOptions` is a `[]string` internally, but the transport is the environment, so each entry becomes one variable and a repeated key is claimed by its first occurrence — the second is dropped and logged. A structured request would fix that properly; it is the same change as the credential-file work in #6, and I have not made it here.

### 6. Options and Secrets overwriting driver-owned variables — fixed

[`buildEnvVars`](https://github.com/kubernetes-sigs/alibaba-cloud-csi-driver/blob/f0e94880120adcba053854dcb861a168dbfa37c6/pkg/mounter/proxy/server/customfuse/driver.go#L268-L334) now:

- emits `mountpoint`, `source` and `readOnly` itself, before any volume data is considered;
- refuses those three names case-insensitively from both legs ([`driverOwnedNames`](https://github.com/kubernetes-sigs/alibaba-cloud-csi-driver/blob/f0e94880120adcba053854dcb861a168dbfa37c6/pkg/mounter/proxy/server/customfuse/driver.go#L222-L226));
- seeds `taken` from the environment the entrypoint would otherwise inherit, so `PATH` and anything the fuse image sets in its own `ENV` cannot be redefined by a PV field or a Secret key;
- applies Secrets before options, so a name is claimed by whichever leg reaches it first and a volume parameter cannot pass its own value off as a credential the volume's Secret carries.

The plugin enforces the same reserved names when it parses the PV, since the two are separate images built and rolled out independently. Dropped names are logged together with the list of names the entrypoint will actually see, values omitted because secrets travel through the same list.

A Secret containing `readOnly=false` now changes nothing: `readOnly` is read off the driver's own option spelling before that leg is treated as volume data, and `fuseOptions.ReadOnly` is derived from the VolumeCapability and the publish request only.

### 7. Read-only depending on publication order — fixed on the bind

The per-consumer bind carries this consumer's own flag ([nodeserver.go#L135-L147](https://github.com/kubernetes-sigs/alibaba-cloud-csi-driver/blob/f0e94880120adcba053854dcb861a168dbfa37c6/pkg/customfuse/nodeserver.go#L135-L147)): `bind,ro` when the publish request is read-only, `bind` otherwise. mount-utils implements that as a bind followed by a remount, so the shared mount is left as it was. `TestNodePublishVolume_BindCarriesReadOnly` seeds `attachPath` as already mounted — the second-consumer path, which is the one that was wrong — and asserts the recorded options for both a read-write and a read-only consumer. It needs its own recorder because `mount-utils`' `FakeMounter` records the action but not the options, and the options are the whole point here.

The other direction is real and this does not fix it: if a read-only consumer mounts first and the client honours `$readOnly`, the shared mount is read-only and a later read-write consumer gets a successful publish over a mount it cannot write to. That follows from mounting once per node, so [README.md#L300-L309](https://github.com/kubernetes-sigs/alibaba-cloud-csi-driver/blob/f0e94880120adcba053854dcb861a168dbfa37c6/examples/customfuse/README.md#L300-L309) tells readers to give the two access modes separate volumes. The same paragraph notes that where the node plugin is injected as a sidecar there is no bind to make read-only, so honouring `$readOnly` is the entrypoint's job there.

### 8. Recipe copies the wrong script; shipped scripts fail — fixed, with tests

- [2-oss-compatible/Dockerfile](https://github.com/kubernetes-sigs/alibaba-cloud-csi-driver/blob/f0e94880120adcba053854dcb861a168dbfa37c6/examples/customfuse/2-oss-compatible/Dockerfile) takes its own directory as the whole build context and does `COPY entrypoint.sh /entrypoint.sh`, so a copy of it anywhere builds the same image and the s3fs/JindoFS recipe replaces the file instead of inheriting one from a fixed repository path.
- `bash -n` passes on all three entrypoints and on both scripts embedded in `4-configmap`'s ConfigMaps. The construct that broke it was a `[[ $x =~ ... ]]` with an unescaped `;` inside the regex; it is a `case` pattern now ([entrypoint.sh#L46-L64](https://github.com/kubernetes-sigs/alibaba-cloud-csi-driver/blob/f0e94880120adcba053854dcb861a168dbfa37c6/examples/customfuse/2-oss-compatible/entrypoint.sh#L46-L64)).
- The `--trash-days` lines no longer have anything after the continuation backslash.
- [`pkg/customfuse/examples_test.go`](https://github.com/kubernetes-sigs/alibaba-cloud-csi-driver/blob/f0e94880120adcba053854dcb861a168dbfa37c6/pkg/customfuse/examples_test.go) runs in `go test ./pkg/...` and checks what `bash -n` does not: a shebang, `exec` as the last act, nothing after a continuation backslash, and — for the ConfigMap demos — that the manifest parses and the script inside it gets the same treatment. It skips when the examples directory is absent, so the Go package is not coupled to the tree layout.

Not done: only JuiceFS is exercised. The s3fs and JindoFS adapters are documented and their scripts pass the checks above, but nothing builds or runs them, so "at least two clients pass the same mount, read-only, timeout, abnormal-exit and unmount tests" is not met by this PR.

### On the abstraction boundary

The table is right, and the layering matches what is here: the generic side owns the mount target, per-consumer read-only semantics, process lifecycle, timeout, unmount, mount health and non-sensitive metrics; the adapter owns client arguments, credential format and initialization. `build/mount-proxy/Dockerfile`'s `customfuse-base` stage is the shared image scaffold — the two proxy binaries and nothing client-specific — and a customer's Dockerfile is that plus their client image plus their script.

Foreground execution is a documented requirement now rather than an assumption, with both ways to get it wrong spelled out: [docs/customfuse.md#L149-L177](https://github.com/kubernetes-sigs/alibaba-cloud-csi-driver/blob/f0e94880120adcba053854dcb861a168dbfa37c6/docs/customfuse.md#L149-L177). The RunC/sandbox boundary is the table under [Where the driver runs](https://github.com/kubernetes-sigs/alibaba-cloud-csi-driver/blob/f0e94880120adcba053854dcb861a168dbfa37c6/docs/customfuse.md#L75), including that on the injected-sidecar path nothing resolves the image from the ConfigMap and there is no bind to make read-only.

The suggested acceptance criterion holds for the mount path: adding a client today means an image, an adapter and a ConfigMap entry, with no Go change. It does not hold for the profile row — there is no administrator-controlled configuration-file or state-volume projection yet, which is #2's second half.

pkg/mounter/fuse_pod_manager/oss/utils.go
```diff
@@ -23,9 +23,12 @@ var (
 )
 
 func SetDefaultImage(fuseType string, m metadata.MetadataProvider, config *fpm.FuseContainerConfig) {
-	// deprecated
+	// An ossfs image is derived from the release registry below, so one named in the
+	// configmap is not honoured. Discarded here rather than in the shared parser,
+	// which other fuse types use to name an image they bring themselves.
 	if config.Image != "" {
-		return
+		klog.Warning("'image' config in configmap no longer supported")
```
mowangdk, line 30: Are there any potential breaking changes here?

mowangdk:
Two findings from reviewing the customfuse mount-proxy process lifecycle at `8cc4dcce2511485392b2c1d1eaee78231682cf39`:

### 1. [P2] Bound startup stderr buffering to avoid OOM

[**pkg/mounter/proxy/server/customfuse/driver.go:123–134**](https://github.com/kubernetes-sigs/alibaba-cloud-csi-driver/blob/8cc4dcce2511485392b2c1d1eaee78231682cf39/pkg/mounter/proxy/server/customfuse/driver.go#L123-L134)

Every byte of entrypoint stderr is copied into an unbounded `bytes.Buffer` until mounting succeeds or the request times out. A noisy FUSE client or pre-mount preparation command can exhaust the mount-proxy container's memory before that deadline.

Moreover, the `cmd.Wait()` goroutine captures `stderrBuf`, so switching the writer after a successful mount does not release the accumulated startup output: it can remain allocated for the client's lifetime.

**Suggested fix:** Use a bounded diagnostic buffer (for example, retaining only the last N KiB), while continuing to stream the full output to `os.Stderr`.

### 2. [P2] Timeout cleanup leaves child processes alive — reproduced

[**pkg/mounter/proxy/server/customfuse/driver.go:194–206**](https://github.com/kubernetes-sigs/alibaba-cloud-csi-driver/blob/8cc4dcce2511485392b2c1d1eaee78231682cf39/pkg/mounter/proxy/server/customfuse/driver.go#L194-L206)

Both SIGTERM and SIGKILL target only the entrypoint PID, not its children. This can affect scripts that correctly end with `exec`: the timeout can happen earlier, while running a preparation command such as the examples' `juicefs format` or quota setup.

I reproduced this in a disposable Linux container with an entrypoint that starts a long-running preparation subprocess, waits for it, and would then `exec` the final client. After the mount deadline and the two-second shutdown grace:

- The preparation subprocess was still alive.
- Its inherited stderr pipe kept `cmd.Wait()` and the driver's wait group blocked until the child was explicitly killed.

Retries can accumulate processes, and driver shutdown can hang.

**Suggested fix:** Start the entrypoint in its own process group and signal the entire group during cleanup, with bounded shutdown/reaping. Apply the same process-group handling to `Terminate()` as well.

**Validation:** The four focused Linux test packages (`pkg/customfuse`, `pkg/mounter/fuse_pod_manager/customfuse`, `pkg/mounter/proxy/server/customfuse`, and `pkg/mounter/utils`) passed. An additional timeout-cleanup reproduction failed as described in finding 2. No live-cluster validation was performed.

pkg/mounter/fuse_pod_manager/oss/utils.go
AlbeeSo: No behaviour change for ossfs/ossfs2 — that early `return` was already unreachable on master.

`e5fafea9a` ("refactor!(oss): move fuse pod operations to controller", 2024-08-20) replaced the parser's `config.Image = value` with the `'image' config in configmap no longer supported` warning, and added `image-tag` as the replacement. Since then `ExtractFuseContainerConfig` never sets `Image`, and the only assignment to `config.Image` anywhere in the non-test tree is the `config.Image = path.Join(prefix, image)` at the end of this same function, after the guard. So `config.Image` is always empty on entry and the `return` cannot fire; that is what the `// deprecated` above it was pointing at.

What changes here is *where* the discard happens. The warning used to sit in the shared parser, which customfuse now also goes through, and customfuse has no release registry to derive an image from — `image=` in `fuse-customfuse` is the only way to name its client. Leaving the discard in the parser meant every customfuse mount logged its own mandatory config as unsupported. The parser now returns what it finds, and ossfs discards it here, where the rule lives.

Same warning text, same discarded value, same derived `<prefix>/csi-<fuseType>:<tag>`. Both managers call this immediately after parsing (`ossfs/manager.go:60`, `ossfs2/manager.go:45`), and the only reads of `config.Image` are in `PodTemplateSpec` (`ossfs/manager.go:222`, `ossfs2/manager.go:273`) — well after, so what they see is unchanged.

AlbeeSo:
Re [the two findings above](https://github.com/kubernetes-sigs/alibaba-cloud-csi-driver/pull/1722#issuecomment-5708009666): both fixed in 18f6fdf25.

### Finding 1

The capture is now a fixed 64 KiB ring (`stderr_tail.go`) that retains the tail of the stream. `os.Stderr` still receives every byte, unchanged.

On the retained-allocation point specifically: the ring is allocated once at its full size and never grows, so what the `cmd.Wait()` goroutine still holds after a successful mount is a constant rather than a function of how chatty the client was during startup. It remains reachable for the client's lifetime, but it is 64 KiB now, not the whole stream.

On choosing N: I did not want to derive it from a guess about how much a client might print, since the entrypoint is a script the customer wrote and nothing about its output is knowable from here. The number comes from where the captured text actually goes. It rides back inside the mount error as `Response.Error`, and the client reads that response through the same `ReadMsg` that reads requests, which caps at `MaxMsgSize` and returns `message too large` past it. So an unbounded capture does not merely waste memory: past a certain volume of output it destroys the diagnosis it exists to deliver, and the failure mode is the caller getting a protocol error instead of the client's last words. 64 KiB keeps that message sendable with a lot of room.

When bytes are dropped, the retained text carries a leading line saying how many were dropped and that the full stream is in this container's log. That is because the message is subject to further truncation downstream, and a reader who never sees the note would take a partial stream for everything the client said.

### Finding 2

The entrypoint now runs with `Setpgid`, and both the timeout path and `Terminate()` signal the process group rather than the pid. `Terminate()`'s wait is bounded: `proxy.MountShutdownGrace`, then SIGKILL to the groups, then a final wait. The pid map entry is deleted by the same goroutine that had been blocked in `cmd.Wait()`, so it no longer leaks and a later shutdown no longer signals a stale pid.

I stopped short of the reaping half of your suggestion and want to say so explicitly rather than leave it implied. What the group kill leaves behind is zombies reparented to PID 1 of the fuse pod container, and nothing collects them.

The obvious fix, a SIGCHLD handler calling `Wait4(-1, ...)`, races `os/exec` for the driver's own children and turns a clean exit into `ECHILD`, which the mount monitor reads as a failure. The other usual answer is an init in front. That is available to `alinas` because the alinas image is entirely ours, and `build/mount-proxy/Dockerfile` does put `tini` there. It is not available here: the customfuse image is the customer's. All four examples build `FROM` the client's own image, copy the two mount-proxy binaries in, and write their own `ENTRYPOINT`. The Dockerfile ships a `customfuse-base` stage as a convenience, but nothing enforces that a given fuse pod uses it, and the ConfigMap delivery path changes only the script, not the image, so an init cannot be relied on to be PID 1.

The kernel releases a zombie's descriptors before it is reaped, including the stderr pipe that the wait was blocked on, so what remains costs a pid slot and nothing that grows. The reasoning is in the comment on `signalGroup`. Happy to revisit if you read that trade differently.

### Validation

Your reproduction is now a committed test. `TestExtendedMountTimeoutKillsTheWholeProcessGroup` starts an entrypoint that forks a long-running preparation subprocess and then waits on it, and asserts that the subprocess is gone, that the driver's wait group drains, and that the pid entry was released. `TestTerminateEscalatesPastAnEntrypointThatIgnoresSIGTERM` covers the shutdown half against an entrypoint that traps SIGTERM away, and asserts it returns within the grace plus the escalation rather than hanging.

The package had no test reaching `ExtendedMount` before, which is how both of these shipped. They now drive the mount through an injected command rather than a script at one of the two fixed paths the container reads. `go test -race ./pkg/...` passes. As in your review, no live-cluster validation was performed for this change.

PR #1724, opened 2026-06-16
Add efc default arguments for bmcpfs

mowangdk:
/kind cleanup

Fixes #

After merging this PR, can we use the same EFC & Alinas RPM versions for both x86 and arm64

```release-note

```

```docs

```

### DIFF
--- a/pkg/bmcpfs/nodeserver.go
+++ b/pkg/bmcpfs/nodeserver.go
@@ -24,6 +24,8 @@ import (
 	"errors"
 	"fmt"
 	"os"
+	"slices"
+	"strings"
 
 	"github.com/container-storage-interface/spec/lib/go/csi"
 	"github.com/kubernetes-sigs/alibaba-cloud-csi-driver/pkg/cloud/metadata"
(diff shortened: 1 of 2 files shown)


pkg/bmcpfs/nodeserver.go
```diff
@@ -159,3 +165,14 @@ func (ns *nodeServer) NodeUnpublishVolume(ctx context.Context, req *csi.NodeUnpu
 	klog.InfoS("NodeUnpublishVolume: succeeded to umount", "targetPath", req.TargetPath)
 	return &csi.NodeUnpublishVolumeResponse{}, nil
 }
+
+// hasMountOption checks whether any option in the slice has the given key prefix (e.g. "g_lease_Enable").
+func hasMountOption(options []string, key string) bool {
+	prefix := key + "="
+	for _, opt := range options {
```
huww98: Can be simplified by `slices.Contains`.

Why EFC itself does not default this to false? What does this option actually do?

huww98:
> can we use the same EFC & Alinas RPM versions for both x86 and arm64

Do you want to update the dockerfile too?

mowangdk:
> > can we use the same EFC & Alinas RPM versions for both x86 and arm64
> 
> Do you want to update the dockerfile too?

No, rpm will update in next PR

PR #1728, opened 2026-06-23
metric/nas: remove StorageMonitorClient

huww98:
#### What type of PR is this?

/kind cleanup
/kind deprecation

Removes `StorageMonitorClient` and the NAS volume capacity metrics
(`node_volume_capacity_bytes_total` / `_used` / `_available`) that depended on
the in-cluster `storage-monitor-service`.

We no longer install this service on new clusters, and existing clusters will
stop serving it after `storage-operator` is upgraded to v1.33.1, so the client
would only ever fail.

This also drops the now-dead supporting code that the removal left behind:
the three capacity `prometheus.Desc` definitions, the `nfsCapacityInfo` type,
the unused event recorder wiring in the NFS collector, the `GiBSize` constant,
and the `NFS_CAPACITY_THRESHOLD_PERCENTAGE` parsing. If that env var is still
set, a warning is now logged explaining it no longer has any effect instead of
silently ignoring it. The unused RBAC `get` rule on `storage-monitor-service`
is removed as well.

This should avoid issuing useless get service request to APIServer, and making the audit log clean.

To avoid losing these metrics, users should install `cnfs-controller` and
upgrade `ack-arms-prometheus` to v1.1.37 from ACK addons **before** upgrading
CSI; these provide the NAS capacity metrics going forward.

```release-note
action required: The NAS volume capacity metrics (node_volume_capacity_bytes_total/used/available) served via storage-monitor-service have been removed. Install cnfs-controller and upgrade ack-arms-prometheus to v1.1.37 from ACK addons before upgrading CSI to keep these metrics. The NFS_CAPACITY_THRESHOLD_PERCENTAGE setting no longer has any effect.
```


mowangdk:
TODO: we need to update the component configuration to include these new dependencies.

PR #1740, opened 2026-06-30
disk: don't send empty DiskCategory/PerformanceLevel in ModifyDiskSpec

huww98:
#### What type of PR is this?

/kind bug
/kind regression

`buildModifySpecRequest` unconditionally set `DiskCategory` and `PerformanceLevel` on the `ModifyDiskSpec` request, even when the corresponding `ModifyParameters` fields were empty.

This was fine with the legacy v1 SDK (`alibaba-cloud-sdk-go`), which stored these as plain string values and omitted empty-string query parameters. After the migration to the v2 SDK (`ecs-20140526/v7`) in a112567, the fields became pointers and `new(string(""))` is a non-nil pointer to an empty string, which the v2 SDK *does* serialize.

As a result, modifying a `cloud_auto` disk (which does not support performance levels) with only `provisionedIops` still sent `PerformanceLevel=`, and ECS rejected the request:

```
OpenAPI returned error: The specified disk category is not supported. (InvalidDiskCategory.NotSupported)
```

This broke the csi-sanity `ModifyVolume` tests:
- `should modify a volume created without a volume attribute class`
- `should modify a volume created with a volume attribute class`

The fix only sets `DiskCategory`/`PerformanceLevel` when non-empty, restoring the v1 behavior.

Fixes #

Tests added/strengthened, both verified to fail against the buggy version and pass with the fix:
- `TestBuildModifySpecRequest`: directly covers the empty `DiskCategory`/`PerformanceLevel` cases.
- `TestModify_HappyPath`: tightened the previously `gomock.Any()` `ModifyDiskSpec` expectation to assert the exact request, so an empty `PerformanceLevel` leaking through is caught at the `Modify()` level.

No release note, this feature is not released yet.
```release-note
NONE
```

### DIFF
--- a/pkg/disk/modify.go
+++ b/pkg/disk/modify.go
@@ -37,9 +37,13 @@ func (p *ModifyParameters) buildModifySpecRequest(diskID string) *ecs20140526.Mo
 		return nil
 	}
 	req := &ecs20140526.ModifyDiskSpecRequest{
-		DiskId:           new(diskID),
-		DiskCategory:     new(string(p.Category)),
-		PerformanceLevel: new(string(p.PerformanceLevel)),
+		DiskId: &diskID,
+	}
+	if p.Category != "" {
+		req.DiskCategory = new(string(p.Category))
+	}
+	if p.PerformanceLevel != "" {
+		req.PerformanceLevel = new(string(p.PerformanceLevel))
 	}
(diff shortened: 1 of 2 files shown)


mowangdk:
/lgtm
/approve

Which versions are affected?

huww98:
No released version affected

PR #1742, opened 2026-06-30
Implement automatic mount point cleanup for NAS-backed sandboxes.

mowangdk:
#### What type of PR is this?
/kind feature

Implement automatic mount point cleanup for NAS-backed sandboxes.

Fixes #

```release-note

```

```docs

```

### DIFF
--- a/pkg/mounter/proxy/server/alinas/driver.go
+++ b/pkg/mounter/proxy/server/alinas/driver.go
@@ -8,6 +8,7 @@ import (
 	"os/exec"
 	"path/filepath"
 	"strings"
+	"sync"
 	"syscall"
 	"time"
 
@@ -27,16 +28,22 @@ const (
 )
 
 func init() {
-	server.RegisterDriver(&Driver{
-		Mounter: mounter.NewForMounter(
-			&extendedMounter{Interface: mount.New("")},
-			interceptors.AlinasSecretInterceptor,
-		),
-	})
+	server.RegisterDriver(NewDriver())
+}
+
+func NewDriver() *Driver {
+	driver := &Driver{}
+	driver.Mounter = mounter.NewForMounter(
+		&extendedMounter{driver: driver, Interface: mount.New("")},
+		interceptors.AlinasSecretInterceptor,
+	)
+	return driver
 }
 
 type Driver struct {
 	mounter.Mounter
+	targets       sync.Map
+	ResetFlagPath string
 }
 
 func (h *Driver) Name() string {
@@ -65,7 +72,40 @@ func (h *Driver) Init() {
 	setupDefaultConfigs()
 }
 
-func (h *Driver) Terminate() {}
+// defaultResetFlagPath is the path to the reset flag file written by envd.
+// When this file exists at termination time, the driver will unmount all
+// tracked NAS mount points before exiting.
+var defaultResetFlagPath = "/etc/aliyun/alinas/reset"
+
+func (h *Driver) resetFlagPath() string {
+	if h.ResetFlagPath != "" {
+		return h.ResetFlagPath
+	}
+	return defaultResetFlagPath
+}
+
+func (h *Driver) Terminate() {
+	if !server.CleanupNASMountsOnExit() {
+		return
+	}
+	if !h.shouldCleanup() {
+		klog.InfoS("Reset flag not found, skipping NAS mount cleanup", "path", h.resetFlagPath())
+		return
+	}
+	h.targets.Range(func(key, _ any) bool {
+		target := key.(string)
+		klog.InfoS("Unmounting NAS mount point on exit", "target", target)
+		if err := h.Unmount(target); err != nil {
+			klog.ErrorS(err, "Failed to unmount NAS mount point", "target", target)
+		}
+		return true
+	})
+}
+
+func (h *Driver) shouldCleanup() bool {
+	_, err := os.Stat(h.resetFlagPath())
+	return err == nil
+}
 
 // ApplyOptionDefaults applies driver-specific option defaults.
 // alinas does not apply any environment-detected defaults.
@@ -176,6 +216,7 @@ func copyFile(src, dst string) error {
 }
 
 type extendedMounter struct {
+	driver *Driver
 	mount.Interface
 }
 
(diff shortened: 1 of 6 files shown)


mowangdk (commented):
## Review

### Design
The approach is reasonable — tracking mounts in a `sync.Map` and cleaning up on termination is straightforward. A few concerns below.

### Issues

**1. Global `cleanupNASMountsOnExit` is a plain bool with no synchronization (`handler.go`)**
`SetCleanupNASMountsOnExit` writes from `main()` and `CleanupNASMountsOnExit()` reads from the `Terminate()` goroutine. This is a data race under the Go memory model. Use `atomic.Bool` or move it into the `Driver` struct.

**2. `Terminate()` — consider `MNT_DETACH` for shutdown unmount (`driver.go`)**
`syscall.Unmount(target, 0)` will fail if the target is busy, and there's no retry. Since this runs on SIGTERM with limited time, `syscall.Unmount(target, syscall.MNT_DETACH)` (lazy unmount) is safer for a shutdown path.

**3. `Unmount` override may not cover all unmount paths (`driver.go`)**
`extendedMounter.Unmount` removes targets from the map, but only if callers go through this wrapper. If anything calls `m.Interface.Unmount()` directly, the map leaks stale entries. Worth verifying no other code path bypasses `extendedMounter`.

**4. `TestShouldCleanup_FileExists` doesn't call `shouldCleanup()` (`driver_test.go`)**
The test reimplements the `os.Stat` logic manually instead of exercising the real method. Since `resetFlagPath` is a const, it can't be redirected. Consider making it a field on `Driver` so the test covers the actual code.

**5. Missing test for the happy-path unmount (`driver_test.go`)**
No test verifies that `Terminate()` actually unmounts tracked targets when both the flag is enabled and the reset file exists. The current tests only cover the "skip" branches — the most important behavior is untested.

### Minor
- `extendedMounter` → `Driver` back-pointer creates a circular reference. Works with GC but makes ownership harder to follow.
- PR body has an empty `Fixes #` — should reference an issue or be removed.
- Release note block is empty.

mowangdk:
## Review Update

The following issues from the initial review have been addressed in the latest force-push:

**Fixed:**
1. ~~Global `cleanupNASMountsOnExit` is a plain bool with no synchronization~~ — Changed to `atomic.Bool`.
2. ~~`TestShouldCleanup_FileExists` doesn't call `shouldCleanup()`~~ — `resetFlagPath` is now a configurable field on `Driver`. Test exercises the real method.
3. ~~Missing test for the happy-path unmount~~ — Added `TestTerminate_CleansUpMounts` which verifies both tracked targets are unmounted when cleanup is enabled and the reset flag exists (via injectable `unmountFunc`).
4. ~~`TestTerminate_CleanupEnabledButNoResetFlag` relies on `/etc/...` not existing~~ — Now uses an explicit nonexistent temp path.

**Verified not an issue:**
- `Unmount` bypass risk — confirmed the only `.Interface.Unmount()` call is inside `extendedMounter.Unmount` itself. `utils.IsNotMountPoint` operates on the passed interface, which would be `extendedMounter` for alinas.

**Still open:**
- PR body has an empty `Fixes #` — should reference an issue or be removed.
- Release note block is empty.

pkg/mounter/proxy/server/alinas/driver.go
```diff
@@ -65,7 +73,41 @@ func (h *Driver) Init() {
...
+		klog.InfoS("Unmounting NAS mount point on exit", "target", target)
+		if err := unmount(target, 0); err != nil {
+			klog.ErrorS(err, "Failed to unmount NAS mount point", "target", target)
+		}
+		return true
+	})
+}
+
+func (h *Driver) shouldCleanup() bool {
+	_, err := os.Stat(h.resetFlagPath())
```
huww98, line 106: Who will write to this path?

pkg/mounter/proxy/server/alinas/driver.go
```diff
@@ -65,7 +73,41 @@ func (h *Driver) Init() {
 	setupDefaultConfigs()
 }
 
-func (h *Driver) Terminate() {}
+var defaultResetFlagPath = "/etc/aliyun-defaults/alinas/reset"
```
huww98:
```suggestion
var defaultResetFlagPath = "/etc/aliyun/alinas/reset"
```
Don't use /etc/aliyun-defaults, which we may set readonly in the future.

pkg/mounter/proxy/server/alinas/driver.go
mowangdk: Written by envd

pkg/mounter/proxy/server/alinas/driver.go
mowangdk: this is an empty-dir for sandbox scenario, which should be fine

pkg/mounter/proxy/server/alinas/driver.go
huww98: That is very implicit. Can you add a comment?

pkg/mounter/proxy/server/alinas/driver.go
huww98: That is wrong, we should mount the emptydir on `/etc/aliyun`. From code we specifically want to preserve data from `/etc/aliyun-defaults`

pkg/mounter/proxy/server/alinas/driver_test.go
```diff
@@ -71,3 +78,165 @@ func TestAddAutoFallbackNFSMountOptions(t *testing.T) {
...
+	driver.targets.Store("/mnt/nas1", struct{}{})
+	driver.targets.Store("/mnt/nas2", struct{}{})
+
+	driver.Terminate()
+
+	// After Terminate, targets should have been cleaned up via Unmount
+	_, loaded := driver.targets.Load("/mnt/nas1")
+	assert.False(t, loaded, "target /mnt/nas1 should be removed after terminate")
+	_, loaded = driver.targets.Load("/mnt/nas2")
+	assert.False(t, loaded, "target /mnt/nas2 should be removed after terminate")
```
huww98, line 173: Better to also assert that Unmount is really called? fakeMounter should enable this.

huww98:
/lgtm
mostly fine, up to you if you want to fix my above comment.

huww98:
forgot to /hold ...

PR #1763, opened 2026-07-04
disk: report HighDensityDisk mode limit as NodeGetInfo max volumes

huww98:
#### What type of PR is this?

/kind feature

When an ECS instance runs in **HighDensityDisk mode (云盘高密模式)**, it can attach far more
cloud disks than its default `DiskQuantity`. This PR makes the disk driver report that raised
limit in `NodeGetInfo.MaxVolumesPerNode` (and the centralized labeler's
`node.csi.alibabacloud.com/max-disk` annotation) so the scheduler packs disks correctly.

The feature relies on non-public ECS OpenAPI fields, so it is behind an **alpha feature gate
`DiskHighDensityMode` (off by default)** with **zero overhead when off** — no extra OpenAPI
calls and no non-standard request fields are sent.

Details:
- High-density state is **per-instance** — `DescribeInstances`
  `AdditionalInfo.EnableHighDensityMode` (needs `AdditionalAttributes=DISK_HIGH_DENSITY_MODE`).
- The raised limit is **per-instance-type** — the `DensityDiskQuantity` attribute of
  `DescribeInstanceTypes` (needs `MaxResults` set, otherwise `Attributes` is omitted).
- `metadata.EffectiveDiskQuantity` composes them: it only makes the per-instance
  `DescribeInstances` call for types that actually advertise a higher `DensityDiskQuantity`,
  so non-capable types take the normal path with no extra API call.
- The gate is consulted at a single point in the metadata layer; the node plugin and the
  labeler need no gate check of their own, and when the gate is off `DescribeInstances`
  carries no `AdditionalAttributes`.

Validated end-to-end on a test node (`ecs.c7a.4xlarge`, high-density on): the
`allocatable-disk` annotation and the CSINode allocatable count flip from 8 to 32.

- Also included: a small dev-env tooling fix (`init-dev-env` skill) so the generated image
  pull secret also authorizes the VPC ACR endpoint, for clusters without public internet.
- `DensityDiskQuantity` / `EnableHighDensityMode` are non-public ECS OpenAPI surface, hence
  the gate.

```release-note
Add an alpha `DiskHighDensityMode` feature gate (off by default). When enabled, the disk driver reports the raised disk attach limit of ECS instances running in HighDensityDisk mode (云盘高密模式) as NodeGetInfo MaxVolumesPerNode.
```

```docs

```

### DIFF
--- a/pkg/cloud/metadata/ecs_instance_type.go
+++ b/pkg/cloud/metadata/ecs_instance_type.go
@@ -2,9 +2,11 @@ package metadata
 
 import (
 	"fmt"
+	"strconv"
 
 	ecs20140526 "github.com/alibabacloud-go/ecs-20140526/v7/client"
 	"github.com/kubernetes-sigs/alibaba-cloud-csi-driver/pkg/cloud"
+	"k8s.io/utils/ptr"
 )
 
 type ECSInstanceTypeMetadata struct {
@@ -14,6 +16,10 @@ type ECSInstanceTypeMetadata struct {
 func NewEcsInstanceTypeMetadata(c cloud.ECSv2Interface, instanceType string) (*ECSInstanceTypeMetadata, error) {
 	req := ecs20140526.DescribeInstanceTypesRequest{
 		InstanceTypes: []*string{&instanceType},
+		// MaxResults is required for the response to include the Attributes array
+		// (which carries DensityDiskQuantity). A single-type query returns one
+		// result with an empty NextToken, so no pagination is needed.
+		MaxResults: new(int64(100)),
 	}
 	resp, err := c.DescribeInstanceTypes(&req)
 	if err != nil {
@@ -36,6 +42,8 @@ func (m *ECSInstanceTypeMetadata) GetAny(_ *mcontext, key MetadataKey) (any, err
 		if m.t.DiskQuantity != nil {
 			return *m.t.DiskQuantity, nil
 		}
+	case diskQuantityHighDensity:
+		return m.DiskQuantityHighDensity(), nil
 	}
 	return nil, ErrUnknownMetadataKey
 }
@@ -48,6 +56,22 @@ func (m *ECSInstanceTypeMetadata) DiskQuantity() int32 {
 	return *m.t.DiskQuantity
 }
 
+// DiskQuantityHighDensity returns the DensityDiskQuantity attribute value (the disk
+// count attachable in HighDensityDisk mode), or 0 if the type does not advertise it.
+func (m *ECSInstanceTypeMetadata) DiskQuantityHighDensity() int32 {
+	if m.t.Attributes == nil {
+		return 0
+	}
+	for _, a := range m.t.Attributes.Attribute {
+		if a != nil && ptr.Deref(a.Name, "") == "DensityDiskQuantity" {
+			if v, err := strconv.Atoi(ptr.Deref(a.Value, "")); err == nil {
+				return int32(v)
+			}
+		}
+	}
+	return 0
+}
+
 type ECSInstanceTypeFetcher struct {
 	ecsClient cloud.ECSv2Interface
 	mPre      middleware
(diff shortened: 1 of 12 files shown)


pkg/cloud/metadata/metadata.go
```diff
@@ -185,6 +202,51 @@ func getT[T any](m *Metadata, key MetadataKey) (T, error) {
...
+		return 0, err
+	}
+	density, err := m.DiskQuantityHighDensity()
+	if err != nil {
+		if errors.Is(err, ErrUnknownMetadataKey) {
+			return normal, nil // type has no high-density support -> no DescribeInstances call
+		}
+		return 0, err
+	}
+	if density <= normal {
```
mowangdk: Is this needed? Dense patterns might have lower counts than usual?

pkg/cloud/metadata/ecs_instance_type.go
```diff
@@ -14,6 +16,10 @@ type ECSInstanceTypeMetadata struct {
 func NewEcsInstanceTypeMetadata(c cloud.ECSv2Interface, instanceType string) (*ECSInstanceTypeMetadata, error) {
 	req := ecs20140526.DescribeInstanceTypesRequest{
 		InstanceTypes: []*string{&instanceType},
+		// MaxResults is required for the response to include the Attributes array
+		// (which carries DensityDiskQuantity). A single-type query returns one
+		// result with an empty NextToken, so no pagination is needed.
+		MaxResults: new(int64(100)),
```
mowangdk, line 22: Where does the number 100 come from?

pkg/cloud/metadata/metadata.go
```diff
@@ -185,6 +202,51 @@ func getT[T any](m *Metadata, key MetadataKey) (T, error) {
...
+	return getT[int32](m, diskQuantityHighDensity)
+}
+func (m *Metadata) IsHighDensityMode() (bool, error) {
+	if !diskHighDensityModeEnabled() {
+		return false, nil
+	}
+	return getT[bool](m, isHighDensityMode)
+}
+
+func diskHighDensityModeEnabled() bool {
```
mowangdk, line 218: I don't think we need this featuregate, it’s already controlled by the OpenAPI.

mowangdk (approved):
## Review

Well-designed feature behind an Alpha gate (`DiskHighDensityMode`, off by default). The approach is sound: per-type data (normal + density disk quantities) is cached, per-instance mode is queried fresh, and non-HD instance types incur zero extra API calls.

### What looks good

- **Zero cost when off:** No extra API calls, no `AdditionalAttributes` sent, no behavioral change. Tests explicitly verify this (`TestGetOpenAPINoAttrWhenGateOff`, `TestHighDensityAccessorsGateOff`, labeler "gate off" subtest).
- **Call avoidance:** Both `EffectiveDiskQuantity` and the labeler skip the per-instance `DescribeInstances` when `density <= normal`, so non-HD types pay nothing even with the gate on.
- **Cache correctness in labeler:** `diskQuantities{normal, highDensity}` is per-type and cached; `IsHighDensityMode` is per-instance and queried fresh per reconcile. Correct because mode can change per instance while type limits are static.
- **Test coverage is thorough:** gate on/off, mode on/off, no-density-support, error propagation, labeler end-to-end patch assertions, and `hdCalls` counter to verify DescribeInstances is skipped when unnecessary.
- **`OpenAPIMetadata` migrated from `Get` to `GetAny`**, removing the `strProvider` indirection cleanly.

### Suggestions (non-blocking)

1. **Labeler duplicates `EffectiveDiskQuantity` decision logic** (`reconcile.go:resolveECS`). The caching reason is valid, but consider extracting the pure decision into a shared helper like `func pickDiskQuantity(normal, density int32, isHD bool) int32` so the two call sites can't diverge.

2. **Unrelated commit bundled:** The first commit (`skill(init-dev-env): authorize VPC registry endpoint in pull secret`) is independent of the disk feature. Splitting it would keep history cleaner.

pkg/cloud/metadata/metadata.go
huww98: We return 0 for instance types that don't support high density mode.

pkg/cloud/metadata/ecs_instance_type.go
huww98: Random. Any value <=100 should work.

pkg/cloud/metadata/metadata.go
huww98: This is to ensure users who don't use this feature not pay for the extra DescribeInstances OpenAPI call.

pkg/cloud/metadata/metadata.go
huww98: Changed to `density <= 0 || density == normal`

PR #1770, opened 2026-07-08
bmcpfs: check fuse_dev_ioctl_recover kernel symbol in init container

mowangdk:
#### What type of PR is this?
/kind feature

Mount host /proc to /host/proc in the init container and check /proc/kallsyms for fuse_dev_ioctl_recover symbol after modprobe fuse. This indicates whether the kernel supports fuse connection recovery.

Also add bmcpfs.enabled to the init container rendering condition so the init container is included when only bmcpfs is enabled.

Fixes #

```release-note

```

```docs

```

### DIFF
--- a/deploy/charts/alibaba-cloud-csi-driver/templates/plugin.yaml
+++ b/deploy/charts/alibaba-cloud-csi-driver/templates/plugin.yaml
@@ -54,7 +54,7 @@ spec:
       hostIPC: true
       hostPID: true
       dnsPolicy: ClusterFirst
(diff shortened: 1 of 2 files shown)


deploy/charts/alibaba-cloud-csi-driver/templates/plugin.yaml
```diff
@@ -84,6 +84,9 @@ spec:
               mountPath: /var/log/
             - name: ossconnectordir
               mountPath: /host/usr/
+            - name: host-proc
+              mountPath: /host/proc
+              readOnly: true
```
huww98: I think /proc is mounted by default

PR #1772, opened 2026-07-13
disk: fix nil deref in CreateSnapshot when snapshot disappears

huww98:
#### What type of PR is this?

/kind bug

`CreateSnapshot` panicked with a nil pointer dereference when a just-created snapshot was deleted (e.g. by the user, or by retention) before the status poll observed it:

```
panic: runtime error: invalid memory address or nil pointer dereference
github.com/.../pkg/disk.formatCSISnapshot(0x0)
    pkg/disk/controllerserver.go:703
github.com/.../pkg/disk.(*controllerServer).CreateSnapshot(...)
    pkg/disk/controllerserver.go:508
```

The root cause is that `waitstatus.WaitFor` returned `(nil, nil)` when the resource was not found. That is easy to mishandle: `CreateSnapshot` checked only `err != nil` and passed the nil snapshot straight into `formatCSISnapshot`, which dereferenced it.

This PR introduces a `waitstatus.ErrNotFound` sentinel so that the *default* behavior — just propagating the returned `err` — is the *safe* behavior. A caller that forgets to special-case "not found" now fails the RPC (which the external-snapshotter retries idempotently) instead of panicking.

- `Simple` and `Batched` now return `ErrNotFound` instead of `(nil, nil)`.
- `CreateSnapshot` handles it explicitly with a clear message.
- `waitForDiskDetached` opts into treating "gone" as success via `errors.Is`.
- `waitForDiskAttached` drops its now-dead nil check (a missing disk now arrives as `err`).

`batcher.LowLatency.Describe` intentionally keeps `(nil, nil)` semantics: it is a plain point lookup, all four of its callers already branch on `disk == nil`, and two of them (DeleteVolume, DetachDisk) treat "not found" as success — which reads more naturally as a nil check than as unwrapping an error.

Fixes #

Tests in `pkg/disk/waitstatus` (`TestWaitForUnfound`, `TestPoll`) are updated to assert `ErrorIs(err, ErrNotFound)`. Full `pkg/disk/...` suite passes.

```release-note
Fixed a panic in the disk controller's CreateSnapshot when a snapshot is deleted immediately after creation.
```

### DIFF
--- a/pkg/disk/cloud.go
+++ b/pkg/disk/cloud.go
@@ -574,24 +574,18 @@ func (ad *DiskAttachDetach) waitForDiskAttached(ctx context.Context, diskID, nod
 	if err != nil {
 		return nil, err
 	}
-	if disk == nil {
-		return nil, fmt.Errorf("waitForDiskAttached: disk %s not found", diskID)
-	}
 	return disk, nil
 }
 
 func (ad *DiskAttachDetach) waitForDiskDetached(ctx context.Context, diskID, nodeID string) error {
-	disk, err := ad.waiter.WaitFor(ctx, diskID, func(disk *ecs.Disk) bool {
+	_, err := ad.waiter.WaitFor(ctx, diskID, func(disk *ecs.Disk) bool {
 		return !waitstatus.IsInstanceAttached(disk, nodeID)
 	})
-	if err != nil {
-		return err
-	}
-	if disk == nil {
+	if errors.Is(err, waitstatus.ErrNotFound) {
 		klog.Infof("waitForDiskDetached: disk %s not found", diskID)
 		return nil
 	}
-	return nil
+	return err
 }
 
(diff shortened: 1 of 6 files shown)


pkg/disk/controllerserver.go
```diff
@@ -512,6 +512,9 @@ func (cs *controllerServer) CreateSnapshot(ctx context.Context, req *csi.CreateS
 	klog.Infof("CreateSnapshot:: Snapshot create successful: snapshotName[%s], sourceId[%s], snapshotId[%s]", req.Name, req.GetSourceVolumeId(), snapshotResponse.SnapshotId)
 
 	snap, err := cs.snapshotWaiter.WaitFor(ctx, snapshotResponse.SnapshotId, waitstatus.SnapshotCut)
+	if errors.Is(err, waitstatus.ErrNotFound) {
```
mowangdk: I didn't see much difference between this and the one below, so we can merge these judgments.

huww98: merged.

PR #1774, opened 2026-07-14
disk: declare ONLINE volume expansion capability

huww98:
#### What type of PR is this?

/kind bug
/kind cleanup

The disk driver advertised `OFFLINE` volume expansion in `GetPluginCapabilities`, but it has actually supported **online** expansion end-to-end since 2019:

- `ControllerExpandVolume` issues an `"online"` ECS `ResizeDisk` when the disk status is `In_use` (`pkg/disk/controllerserver.go`), and never rejects an attached disk.
- The node side grows the partition on a mounted disk with `sfdisk --no-reread --no-tell-kernel` + `partx` (`pkg/disk/sfdisk/expand.go`) and then resizes the filesystem in place.

This PR flips the advertised capability from `OFFLINE` to `ONLINE` so the declaration matches the implementation, per the CSI spec's block-storage ONLINE example (a plugin with both controller and node `EXPAND_VOLUME` that supports expanding published volumes).

#### Why this was safe before, and remains safe:

Nothing actually consumed the `OFFLINE` declaration, which is why online resize already worked in practice:

- `external-resizer` only reads the `CONTROLLER_SERVICE` plugin capability; it does **not** inspect `ONLINE`/`OFFLINE` at all.
- Its "cannot expand in-use volume" gating is driven solely by an *in-use error returned by the driver* — which this driver never returns.
- We additionally run the resizer with `--handle-volume-inuse-error=false`, disabling that machinery entirely.
- Kubernetes core (kubelet CSI plugin, controllers) does not consume this capability either.

So this is purely a truth-in-advertising fix. The only behavioral effect is positive: any CO that *did* honor `OFFLINE` would have forced a detach before resize; declaring `ONLINE` removes that unnecessary restriction. There is no path where `ONLINE` causes an expansion that `OFFLINE` permitted to now be rejected.

Fixes #

Only the identity server capability and its unit test change; the expand implementation is unchanged.

```release-note
disk: advertise ONLINE volume expansion capability, matching the driver's long-standing support for online (in-use) disk resize.
```

### DIFF
--- a/pkg/disk/identityserver.go
+++ b/pkg/disk/identityserver.go
@@ -62,7 +62,7 @@ func (iden *identityServer) GetPluginCapabilities(ctx context.Context, req *csi.
 			{
 				Type: &csi.PluginCapability_VolumeExpansion_{
 					VolumeExpansion: &csi.PluginCapability_VolumeExpansion{
-						Type: csi.PluginCapability_VolumeExpansion_OFFLINE,
+						Type: csi.PluginCapability_VolumeExpansion_ONLINE,
 					},
 				},
 			},
--- a/pkg/disk/identityserver_test.go
+++ b/pkg/disk/identityserver_test.go
@@ -22,7 +22,7 @@ func TestGetPluginCapabilities(t *testing.T) {
 		{
 			name:                       "volume group snapshots disabled",
 			enableVolumeGroupSnapshots: false,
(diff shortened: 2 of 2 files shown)


mowangdk:
If we make these changes, what happens if a PVC is expanded while no pod is using it?

huww98:
Still works. As said in PR desc, no one actually read this currently in Kubernetes.

PR #1777, opened 2026-07-15
Update path validation

mowangdk:
#### What type of PR is this?

/kind cleanup

Change path validation for hidden dir

Fixes #

```release-note

```

```docs

```

### DIFF
--- a/pkg/utils/auth.go
+++ b/pkg/utils/auth.go
@@ -17,9 +17,9 @@ limitations under the License.
 package utils
 
 import (
-	"errors"
 	"fmt"
 	"os"
+	"path/filepath"
 	"strings"
 	"time"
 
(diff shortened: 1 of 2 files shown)


pkg/utils/auth.go
```diff
@@ -72,16 +72,67 @@ type AccessControl struct {
...
+
+	// Also reject if the resolved path falls under a sensitive prefix.
+	for _, prefix := range sensitivePathPrefixes {
+		if cleaned == prefix || strings.HasPrefix(cleaned, prefix+string(os.PathSeparator)) {
+			return false, fmt.Errorf("path %s is under sensitive system path %s", path, prefix)
+		}
+	}
+
+	// Reject paths that are under any directory in the PATH environment variable.
+	for _, dir := range getSensitiveDirsFromPATH() {
```
huww98: Why bother so many round of checks? I'd just EvalSymlinks (implies Clean), then check once with the merged prefix list.

mowangdk: note: paths like /proc/self/fd/0 resolve to actual devices (e.g., /dev/pts/0), allowing them to bypass /proc prefix checks.

pkg/utils/auth.go
```diff
@@ -72,16 +72,55 @@ type AccessControl struct {
...
-		msg = msg + fmt.Sprintf("Path %s has illegal access.", path)
-		return false, errors.New(msg)
+	if !filepath.IsAbs(path) {
+		return false, fmt.Errorf("path %s must be an absolute path", path)
+	}
+
+	cleaned := filepath.Clean(path)
+	if isUnderProc(cleaned) {
+		return false, fmt.Errorf("path %s is under sensitive path /proc", path)
+	}
```
huww98, line 131: IMO, we should also reject symlinks pointing to /proc

pkg/utils/auth_test.go
```diff
@@ -166,28 +136,86 @@ func TestValidatePath(t *testing.T) {
...
+	prefixDir := filepath.Join(base, "binary") // shares a prefix with binDir
+	for _, d := range []string{binDir, otherDir, prefixDir} {
+		if err := os.MkdirAll(d, 0o755); err != nil {
+			t.Fatal(err)
+		}
+	}
+
 	origPath := os.Getenv("PATH")
 	defer func() { _ = os.Setenv("PATH", origPath) }()
+	_ = os.Setenv("PATH", binDir)
```
huww98: Use `t.Setenv()`, here and several other places.
otherwise
/lgtm

/hold
For you to fix the setenv.

PR #1781, opened 2026-07-18
utils: resolve symlinks recursively for non-existent path suffixes

mowangdk:
#### What type of PR is this?

/kind bug

The agent scenario involves multiple non-existence paths. resolveSymlinks previously resolved only one missing trailing component by evaluating the parent directory once. When multiple trailing components do not exist, resolution failed and ValidatePath rejected the path outright.

Now it recursively walks up to the deepest existing ancestor (only on os.IsNotExist errors), resolves its symlinks, and re-appends the full non-existent suffix, so symlinks in any existing ancestor are followed and exact-match sensitive-path checks remain effective. Other resolution errors are propagated instead of triggering the recursion.

Add TestResolveSymlinks covering the exact output contract, and extend ValidatePath tests for multi-level missing paths through symlinked ancestors and non-ENOENT error propagation.

Fixes #

```release-note

```

```docs

```

### DIFF
--- a/pkg/utils/auth.go
+++ b/pkg/utils/auth.go
@@ -115,7 +115,7 @@ func resolveSymlinks(cleaned string) (string, error) {
 	parent := filepath.Dir(cleaned)
 	if parent == cleaned {
 		// Reached the root without resolving any ancestor.
-		return "", fmt.Errorf("cannot resolve any ancestor of %s", cleaned)
+		return "", fmt.Errorf("cannot resolve any ancestor of %s: %w", cleaned, err)
 	}
 	resolvedParent, err := resolveSymlinks(parent)
 	if err != nil {
--- a/pkg/utils/auth.go
+++ b/pkg/utils/auth.go
@@ -126,9 +126,13 @@ func resolveSymlinks(cleaned string) (string, error) {
 
 // ValidatePath checks that path is a safe mount path.
 //
-// It requires the path to be absolute, then:
-//  1. Rejects paths literally under /proc (even if symlinks resolve elsewhere).
-//  2. Resolves symlinks and rejects paths whose real location is under a PATH directory.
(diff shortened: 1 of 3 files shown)


pkg/utils/auth.go
```diff
@@ -95,22 +95,33 @@ func isUnderPATH(resolved string) (string, bool) {
...
 		return resolved, nil
 	}
-	resolvedDir, err := filepath.EvalSymlinks(filepath.Dir(cleaned))
+	if !os.IsNotExist(err) {
+		return "", err
+	}
+	parent := filepath.Dir(cleaned)
+	if parent == cleaned {
+		// Reached the root without resolving any ancestor.
+		return "", fmt.Errorf("cannot resolve any ancestor of %s", cleaned)
```
huww98: I think this is mostly unreachable, root path should always resolvable. If we do reach here, we should pass along the error for diagnosis?

```suggestion
		return "", fmt.Errorf("cannot resolve any ancestor of %s: %w", cleaned, err)
```

pkg/utils/auth.go
```diff
@@ -95,45 +95,69 @@ func isUnderPATH(resolved string) (string, bool) {
...
 		return false, fmt.Errorf("path %s must be an absolute path", path)
 	}
 
-	cleaned := filepath.Clean(path)
-	if isUnderProc(cleaned) {
+	for _, seg := range strings.Split(path, "/") {
+		if seg == ".." {
+			return false, fmt.Errorf("path %s contains '..' which is not allowed", path)
+		}
+	}
```
huww98:
```suggestion
	if slices.Contains(strings.Split(path, "/"), "..") {
		return false, fmt.Errorf("path %s contains '..' which is not allowed", path)
	}
```

huww98: /lgtm
/hold
if you also want to simplify this.

PR #1784, opened 2026-07-21
feat(mounter): jwtauth credential support with pluggable credential sinks

mowangdk:
### Summary

Adds jwtauth (agent-identity) credential support for mount-proxy drivers and extracts the credential flow into a reusable `pkg/mounter/jwtauth` package with pluggable credential delivery, so every mount-proxy driver (ossfs, ossfs2, alinas, future customfuse) can consume it. A jwtauth mount exchanges the sandbox agent-identity token for a scoped STS credential and keeps it fresh for the lifetime of the mount, so alinas/CPFS mounts can be authorized without a long-lived AccessKey stored in a Secret or on disk.

### What's included

**`pkg/mounter/jwtauth` — reusable credential flow**

- `Opts`/`STSToken` plus a TLS-strict token exchange (`FetchSTSToken`): TLS verification is never disabled; when a CA file is configured it must be readable and parseable, otherwise the system root pool is used.
- `Refresher` — sink-generic refresh loop that renews before expiry. `Start(ctx)` fetches + applies the initial credential then launches the loop; `StartWith(cred)` supports consumers that deliver the initial credential themselves (alinas injects it as mount options).
- `Manager` — keyed by mount target: `StopByTarget` on unmount, `StopAll` in driver `Terminate`. Shutdown is bounded (mirroring `MountMonitorManager`); the package-level `StopAll` budget is derived from the per-refresher `stopWaitTimeout` so the documented bound actually holds and no refresh goroutine outlives it.
- **`CredentialSink` strategy** — `Apply(cred)` on initial fetch and every rotation, `Cleanup()` after stop:
  - `ExecSink`: pushes rotated credentials to a live alinas mount by executing `alinas-tls-cert-refresh --mount-point <t> --ak --sk --token`; nothing is written to disk, argv is never logged.
  - A file-based sink (in the `interceptors` package) for FUSE clients that read AK/SK/token files, using atomic directory-symlink rotation.

**`interceptors` — driver wiring**

- `AlinasJWTAuthInterceptor` (wired into the alinas driver): exchanges the STS credential in memory and injects the STS triple as `SensitiveOptions` (masked via `MountSensitive`), forces `tls`+`ram`, strips `authType` so it never reaches `mount.nfs`. After a successful mount it starts an `ExecSink` refresher so the mount stays authorized past STS expiry; the refresher is stopped on unmount (`StopByTarget`) and driver `Terminate` (`StopAll`).
- `JWTAuthInterceptor` (customfuse-style FUSE mounts): resolves config, starts a background refresher backed by the file sink, rewrites mount options so the FUSE entrypoint only sees `credentialDir` + `authType`; refresher lifetime is bound to the mount process via `OssfsMountResult.ExitChan`. The file sink shares the existing `rotateTokenFiles`/`cleanupTokenFiles` helpers with `OssfsSecretInterceptor` by living in the same package.
- Both recognize `authType=agent-identity` (canonical, matching OSS) with `authType=jwtauth` as a backward-compatible alias.
- Defaults reuse the shared mount-proxy env vars via `pkg/mounter/utils/agentidentity`: `AGENT_IDENTITY_ENDPOINT`, `AGENT_IDENTITY_CERT_FILE`, `AGENT_IDENTITY_TOKEN_DIR` (no hardcoded fallbacks).

**Security fix** — the STS triple is carried in the new `MountOperation.SensitiveOptions` and mounted via `MountSensitive`, so secrets are masked in mount logs and error strings (previously leaked in plain `Options` on mount failure).

**NAS driver** — the NAS nodeserver forwards agent-identity options (`authType` / `sandboxId` / `sandboxCredProviderName`, with `credentialProviderName` alias) from the volume context into the alinas mount options via `appendJWTAuthOptions`, aligning the NAS driver with OSS so mounts work without hardcoding credentials in the PV.

**Docs** — adds `docs/nas-agent-identity.md` (linked from `docs/nas.md`) describing the option contract and prerequisites.

### Scope

The consuming customfuse driver that wires `JWTAuthInterceptor` is intentionally out of scope and will land separately.

### Testing

- `go test -count=1 ./pkg/mounter/...` — all pass (run in Linux container).
- `go test -race ./pkg/mounter/jwtauth/... ./pkg/mounter/interceptors/...` — race-clean.
- Coverage: `jwtauth` 94.9%, `interceptors` 92.3%.
- `go build ./...`, `go vet`, `gofmt` — clean.
- Tests cover: `FetchSTSToken` success/bad-CA, `exchangeSTSToken` error branches, `ExecSink` command contract (and no secret in errors), rotation via `StartWith`, refresher start-failure/stop/retry paths, manager stop-by-target/idempotency/`StopAll`, sensitive-options passthrough, refresher stop on unmount, and handler-error cleanup.

### TODO before ready for review

- [ ] Wire the interceptor into the customfuse mount-proxy driver (follow-up PR).
- [ ] End-to-end validation on a live cluster (including `alinas-tls-cert-refresh` rotation).
- [ ] Confirm `ram` mount option should be auto-added by the interceptor (current behavior) vs. supplied via PV options.


AlbeeSo:
The `refreshMargin` in `pkg/mounter/jwtauth/refresher.go` is hardcoded to 5 minutes with no way to configure it externally. Should we make this configurable?

Use cases where 5 minutes may not be ideal:
- Very short-lived STS tokens (e.g. 10-15 min) — 5 min margin eats a large fraction of the validity window
- Environments with high network latency or flaky connectivity — a larger margin provides more retry headroom

mowangdk:
> The `refreshMargin` in `pkg/mounter/jwtauth/refresher.go` is hardcoded to 5 minutes with no way to configure it externally. Should we make this configurable?
> 
> Use cases where 5 minutes may not be ideal:
> 
> * Very short-lived STS tokens (e.g. 10-15 min) — 5 min margin eats a large fraction of the validity window
> * Environments with high network latency or flaky connectivity — a larger margin provides more retry headroom

Fixed

PR #1769, opened 2026-07-08
feat(ossfs2): support FUSE fd-passing and Crash Recovery

AlbeeSo:
#### What type of PR is this?
/kind feature

#### What this PR does / why we need it:

# ossfs2 FUSE fd-passing and Crash Recovery

## 1. Overview

In RunC mode, the ossfs2 daemon runs inside a shared fuse pod. When the daemon crashes (OOM, panic, SIGKILL), the FUSE mount becomes stale and all pods using that volume suffer IO failures until a manual `NodeUnstageVolume` cleanup.

This PR introduces two complementary features:

1. **FUSE fd-passing**: The CSI client performs the kernel mount and passes the FUSE file descriptor to the fuse pod via SCM_RIGHTS, decoupling kernel mount ownership from daemon lifecycle.
2. **Automatic recovery**: On non-SIGTERM exit, the supervisor flushes the FUSE connection (interrupting in-flight requests with `-EINTR`), then restarts ossfs2 with exponential backoff (up to 5 attempts).

## 2. Feature Gates

| Feature Gate | Default | Stage | Description |
|---|---|---|---|
| `EnableFUSEFdPassing` | false | Alpha | Client performs kernel mount; fd passed to server |
| `EnableOssfs2Recovery` | false | Alpha | Auto-restart on crash (implies fd-passing) |

Both gates must be configured **identically** on csi-provisioner (controller) and csi-plugin (node). Mismatched configurations lead to undefined behavior.

## 3. Architecture and Key Components

```
CSI Client ──SCM_RIGHTS(fd)──► Fuse Pod (Server)
  1. open /dev/fuse              1. recvMsgWithFd parses OOB
  2. mount -t fuse (kernel)      2. dup fd to prevent GC finalizer close
  3. WriteMsgUnix(msg, oob)      3. start ossfs2 (ExtraFiles)
  4. wait for response           4. wait for readiness (stat/poll)
  5. SafeCleanup on failure      5. superviseProcess monitors lifecycle
```

| Package / File | Responsibility |
|---|---|
| `pkg/features/features.go` | Feature gate definitions |
| `pkg/utils/os/kernel.go` | Kernel/OS prerequisite checks (alinux3, >=5.10.134-18, x86_64) |
| `proxy/client/client.go` + `fuse_options.go` | Client: open /dev/fuse, kernel mount, SCM_RIGHTS send, FUSE/daemon option separation |
| `proxy/server/handler.go` | Server: recvMsgWithFd, fd lifecycle management |
[...]

PR #1786, opened 2026-07-23
feat: support overlayfs on top of FUSE in RunD/Sandbox

AlbeeSo:
Add overlay filesystem support for OSS volumes in RunD/Sandbox environments. When overlay is enabled, mount-proxy-server mounts the underlying FUSEas a read-only lower dir, then creates an overlayfs with a writable upper layer, exposing the merged view to the business container.

Constraints:
- Only supported in RunD/Sandbox (requires mount-proxy-server in VM)
- Underlying volume must be read-only (overlay provides writable layer)
- Forward-compatible: old mount-proxy-server ignores unknown JSON fields

#### What type of PR is this?
/kind feature

### Scope

- **OSS on Sandbox**: supported and tested
- **NAS**: intentionally not included. The alinas mount-proxy driver has additional complexity (subdir lazy creation with retry, mount-proxy restart preservation via reset flags) that requires careful handling before overlay can be safely integrated. The overlay interceptor is storage-type agnostic and ready for NAS integration in a follow-up.
- **RunC**: not yet implemented. Requires different isolation model (per-pod overlay dirs) since multiple pods share the host filesystem.

#### Basic Functionality

| Test Case | Result |
|-----------|--------|
| Write to overlay (upper layer writable despite read-only PV) | PASS |
| Read original data through overlay lower | PASS |
| Delete masking (overlay whiteout hides lower layer files) | PASS |
| Writes do NOT propagate to OSS (verified via ossutil) | PASS |
| Shallow hibernate (5 min) → wake-up → upper dir data persisted | PASS |
| New writes work after wake-up | PASS |
| Checkpoint clone preserves upper dir data | PASS |

##### Checkpoint Design Note

Overlay dir path is computed from `sha256(target_path)` where target path = `/run/csi/mount-root/<driver>/md5(<mountPath>)`. This is deterministic and stable across checkpoint clone because:
- `mountPath` is preserved in the sandbox annotation across checkpoint/clone
- The target path derivation is deterministic (md5 of mountPath)
- Same mountPath → same target → same overlay dir → upper data restored from checkpoint filesystem

We intentionally do NOT use volumeID as the key because volumeID contains a random suffix that changes on each mount (across checkpoint clone and deep hibernate resume).

##### Deep Hibernate Note

Sandbox has two hibernate modes:
- **Shallow hibernate**: VM frozen in place, hostPath data preserved → upper dir survives (tested, PASS)
- **Deep hibernate**: VM destroyed, Pod recreated → upper dir data **lost** without checkpoint

Deep hibernate destroys the VM, so the hostPath volume (containing upper dir) is gone. If the platform performs a checkpoint before deep hibernate, the upper dir is included in the filesystem snapshot and can be restored on resume (same mechanism as checkpoint clone, verified PASS).

Note: deep hibernate cannot be triggered manually — it occurs when the platform proactively releases the VM during shallow hibernate. The data loss scenario (deep hibernate without checkpoint) has not been directly tested, but follows from the fact that VM destruction erases all local state.

#### Remote/Local Change Interaction (per overlay instance)

Each overlay mount is an independent instance with its own upper dir. The lower layer is a live FUSE proxy to OSS — not a frozen snapshot.

| Scenario | What container sees | OSS affected? |
|----------|-------------------|---------------|
| Remote file added (not in upper) | New file visible | N/A |
| Remote file deleted (not in upper) | File disappears | N/A |
| Remote file updated (not in upper) | New content visible | N/A |
| Local write to existing file (triggers copy-up) | Local version in upper | No |
| After copy-up, remote updates same file | Local version wins (upper shadows lower) | No |
| Local delete (whiteout) | File hidden | No |
| Local create (new file in upper) | Visible only locally | No |
| Direct modification of upper dir on VM host | **Undefined behavior** — overlay dentry cache becomes inconsistent | N/A |

**Important**: Do NOT directly modify files under the overlay upper/lower dirs on the VM host. Overlay relies on kernel dentry cache consistency; bypassing the merged view causes stale cache issues (e.g. `ls` not showing files that `cat` can read).

#### Multi-Volume / Multi-Container Tests

| Configuration | Result | Notes |
|---------------|--------|-------|
| 2 different PVs → 1 container, 2 mountPaths | **PASS** — fully isolated | Different target path → different overlay dir |
| 1 PV → 1 container, 2 mountPaths | **PASS** — fully isolated | Same PV but different mountPath → different target → different overlay dir |
| 1 PV → 2 containers | **NOT SUPPORTED** | Sandbox controller only mounts to mainContainer (platform limitation, not overlay limitation) |

For the "1 PV → 2 mountPaths" case, verified that:
- Writes in `/data-a` are invisible in `/data-b` and vice versa
- Whiteout in `/data-a` does not affect `/data-b`
- Remote update is visible in the mount that has no copy-up, invisible in the mount that does

Fixes #

### Deployment Notes

A dedicated hostPath volume MUST be added to the `csi-agent-sidecar` container at `/run/csi-overlay`, backed by a persistent ext4/xfs partition on the VM host. Without this hostPath volume:
1. Overlay mount fails with `"filesystem not supported as upperdir"` because the container rootfs is itself an overlayfs, and Linux kernel rejects nested overlay-on-overlay.
2. Even if mount succeeded (e.g. on tmpfs), upper dir data would be lost on container restart or hibernate/wake-up since non-hostPath volumes are ephemeral.

The hostPath must support `d_type` (ext4/xfs). In the `sandbox-injection-config` ConfigMap, add to `csi-agent-sidecar` volumeMounts and volumes:
```json
{"name": "csi-overlay", "mountPath": "/run/csi-overlay"}
{"name": "csi-overlay", "hostPath": {"path": "/run/csi-overlay", "type": "DirectoryOrCreate"}}
```

### NAS Overlay: Not Supported (Future Work)

NAS overlay is by design not included in this PR. The overlay interceptor is storage-type agnostic and ready for NAS integration, but the alinas driver has additional complexity that must be addressed first:

1. **Subdir lazy creation with retry**: `doMount` retries with a fresh mount after creating subpath directories. The retry must carry the `Overlay` flag, otherwise the second mount silently produces a non-overlay volume.
2. **Mount-proxy restart preservation**: alinas deliberately preserves NFS kernel mounts across mount-proxy restarts (controlled by `CleanupNASMountsOnExit` / reset-flag gates). `TerminateOverlays()` must respect the same gates to avoid tearing down overlay during upgrades while leaving the NFS mount alive.
3. **NFS-specific behavior (from early testing)**: When a file is deleted remotely from NAS while the container has previously accessed it, `cat`/`open` returns `Stale file handle` (ESTALE) instead of `No such file or directory`. This is standard NFS client behavior (cached file handle invalidated by server). `ls` on the parent directory works normally — the deleted file disappears from listing. This does not affect correctness but may surprise users expecting consistent error messages across OSS and NAS.

```

<!--
If no, just write "NONE" in the release-note block below.
If yes, a release note is required:
Enter your extended release note in the block below. If the PR requires additional action from users switching to the new release, include the string "action required".

For more information on release notes see: https://git.k8s.io/community/contributors/guide/release-notes.md
-->
```release-note

```

```docs

```


pkg/mounter/proxy/server/ossfs/driver.go
```diff
@@ -34,18 +34,22 @@ type Driver struct {
 	pids           sync.Map
 	monitorManager *server.MountMonitorManager
 	wg             sync.WaitGroup
+	overlay        *server.OverlayManager
 }
 
 func NewDriver() *Driver {
+	rawMounter := mount.NewWithoutSystemd("")
```
mowangdk, line 42: We should avoid bringing overlay to NAS for the time being, as mount-proxy already contains subdir creation and several other pieces of logic.

pkg/oss/nodeserver.go
```diff
@@ -165,6 +166,13 @@ func (ns *nodeServer) NodePublishVolume(ctx context.Context, req *csi.NodePublis
 		return ns.publishDirectVolume(ctx, req, opts)
 	}
 
+	// Only Sandbox and Rund support overlay
+	if opts.Overlay {
```
mowangdk, line 177: Let’s keep this limited to the sandbox scenario for now. There doesn’t seem to be any current requirement or traffic for runD, so we can avoid adding extra logic for it until there is a clear need.

pkg/mounter/proxy/server/overlay.go
```diff
@@ -0,0 +1,136 @@
...
+	options := []string{
+		fmt.Sprintf("lowerdir=%s", lowerDir),
+		fmt.Sprintf("upperdir=%s", upperDir),
+		fmt.Sprintf("workdir=%s", workDir),
+	}
+	if err := m.mounter.Mount("overlay", mergedDir, "overlay", options); err != nil {
+		return fmt.Errorf("overlay mount failed: %w", err)
+	}
+
+	m.overlays.Store(mergedDir, true)
```
mowangdk: Will the overlays be cleaned up if mount-proxy crashes or gets OOM-killed?

pkg/nas/utils.go
```diff
@@ -149,6 +149,7 @@ func doMount(m mounter.Mounter, opt *Options, targetPath, volumeId, podUid strin
 		Options:  combinedOptions,
 		Secrets:  secrets,
 		VolumeID: volumeId,
+		Overlay:  opt.Overlay,
```
mowangdk: **Bug (silent wrong mount):** `Overlay` is only set on this first mount attempt. The retry mount at the end of `doMount` (after lazy subpath creation, ~L195) does **not** carry `Overlay: opt.Overlay`.

So when overlay is enabled and the subpath does not exist yet (extreme/CPFS create it lazily via that fallback), the first publish mounts NFS **directly at the merged target with no overlay on top** — and returns success. The pod silently gets a read-only NFS mount instead of the writable merged view.

The temp root mount in between is correctly overlay-free, but the final retry is the same logical mount as this one and must carry the flag. This interaction is also a concrete argument for keeping NAS out of the initial scope as suggested above.

pkg/mounter/proxy/server/alinas/driver.go
```diff
@@ -85,6 +102,9 @@ func (h *Driver) resetFlagPath() string {
 }
 
 func (h *Driver) Terminate() {
+	// First unmount all overlay mounts (must happen before NFS cleanup)
+	h.overlay.TerminateOverlays()
```
mowangdk: **Restart safety:** this runs unconditionally, *before* the `CleanupNASMountsOnExit()` / reset-flag gates below. The alinas driver deliberately preserves kernel NFS mounts across a mount-proxy restart (upgrade) and only cleans up when the reset flag exists — but overlayfs mounts (also kernel mounts, which would equally survive the restart) now get torn down on every graceful terminate.

Net effect: a cnfs-nas-daemon/mount-proxy upgrade inside the VM unmounts every merged view under running pods while the underlying NFS mounts intentionally stay alive. Suggest moving `TerminateOverlays()` behind the same gates for alinas (ossfs/ossfs2 are fine as-is since their Terminate kills the FUSE processes anyway).

pkg/nas/nodeserver.go
```diff
@@ -738,6 +750,12 @@ func (ns *nodeServer) NodeUnpublishVolume(ctx context.Context, req *csi.NodeUnpu
 	}
 	klog.Infof("NodeUnpublishVolume: unmount volume on %s successfully", targetPath)
 
+	// Best-effort cleanup of overlay lower dir (NFS mount).
+	// In RunD/Sandbox with overlay, mount-proxy-server mounted NFS to a lower dir.
+	// After overlay unmount above, the lower dir mount may still be alive.
+	// VM destruction will reclaim all resources regardless.
+	server.CleanupOverlayLowerDir(targetPath)
```
mowangdk: **Layering / skew risk:** the node server (proxy *client*) now imports `pkg/mounter/proxy/server` and hard-codes the server's private dir scheme (`/run/csi-overlay/<sha256(target)>/lower`) to unmount something the server mounted. Same applies to `pkg/oss/nodeserver.go`.

Two concerns:
1. The PR description advertises protocol forward-compatibility, but this creates an out-of-band coupling: if the server-side dir scheme ever changes, this cleanup silently becomes a no-op (the `IsLikelyNotMountPoint` guard swallows it) and lower mounts leak with no signal.
2. It bakes in the assumption that client and server share a mount namespace — true only inside the VM, and nothing enforces it here.

Consider an explicit unmount/cleanup message in the proxy protocol so the server owns lower-dir teardown; failing that, at least move the path scheme into the shared `proxy` protocol package instead of `server`.

pkg/oss/nodeserver.go
```diff
@@ -256,6 +264,8 @@ func (ns *nodeServer) NodePublishVolume(ctx context.Context, req *csi.NodePublis
 			Options:     mountOptions,
 			Secrets:     authCfg.Secrets,
 			MetricsPath: metricsPath,
+			VolumeID:    req.VolumeId, // Required for overlay: mount-proxy-server uses VolumeID to compute overlay dir paths
```
mowangdk: Stale comment: overlay dir paths are keyed by the **target path** (see `overlayDirs` — the doc there explicitly explains why volumeID must NOT be used, since it changes across checkpoint clone / hibernate resume). Passing `VolumeID` here may still be useful for other purposes, but the justification in this comment contradicts the final design and will mislead the next reader.

pkg/mounter/interceptors/overlay.go
```diff
@@ -0,0 +1,68 @@
...
+func NewOverlayInterceptor(manager *server.OverlayManager) mounter.MountInterceptor {
+	return func(ctx context.Context, op *mounter.MountOperation, handler mounter.MountHandler) error {
+		if !op.Overlay {
+			return handler(ctx, op)
+		}
+
+		// Token rotation: if overlay merged is already mounted, pass through to
+		// downstream handlers (secret interceptor handles credential update).
+		// Do NOT attempt overlay mount again.
+		notMnt, err := raw.IsLikelyNotMountPoint(op.OverlayMerged)
```
mowangdk: The trickiest behaviors in this PR live in this interceptor — the token-rotation passthrough (which the comment itself calls critical: a second `MountOverlay` would tear down a working mount) and the cleanup-of-underlying-mount on overlay failure — but neither has a test. The new tests only cover pure path computation.

`OverlayManager` takes an injected `mountutils.Interface`, so both paths are directly testable with `FakeMounter` (rotation: merged already in MountPoints -> handler called, no second overlay mount; failure: overlay Mount errors -> lower gets unmounted). Given interceptor ordering subtleties across three drivers, these tests would be the main regression guard.

pkg/utils/util.go
```diff
@@ -677,7 +677,51 @@ func parseUdevadmInfoSerial(out string) string {
...
+
+// ValidateOverlayRequirements checks that overlay prerequisites are met.
+// Callers should only invoke this when overlay is enabled.
+//
+// Requirements:
+//   - readOnly: the underlying volume must be read-only (overlay provides the writable upper layer).
+//   - inVM: the CSI process must be running inside a VM (RunD or Sandbox). In RunC, the mount path
+//     goes through global mount + bind mount on the host, which is incompatible with overlay
+//     (overlay needs mount-proxy-server inside the VM to manage lower/upper/merged dirs together).
+func ValidateOverlayRequirements(readOnly, inVM bool) error {
```
mowangdk, line 724: Nit: the "am I in a VM" decision is now expressed with two different formulas — NAS uses `ns.config.AgentMode || utils.GetSkipGlobalMount(false)` while OSS uses `runtimeType == RuntimeTypeRunD || ns.skipGlobalMount` — and `GetSkipGlobalMount`'s doc overloads a mount-behavior env var into a VM indicator. These will drift the next time a runtime type is added, and a stray `SKIP_GLOBAL_MOUNT=true` on a RunC node would unlock the overlay gate. Consider one explicit helper (e.g. `utils.RunningInSandbox(agentMode, runtimeType)`), especially if the scope narrows to sandbox-only per the earlier comment.

pkg/oss/nodeserver.go
AlbeeSo: Sandbox and RunD cannot be further distinguished at the nodeserver level. Explicitly distinguishing them would require introducing an additional variable with no clear benefit.
In practice, overlay **requires a dedicated hostPath volume at /run/csi-overlay backed by ext4/xfs on the VM host**. We can only add this hostPath only to the Sandbox deployment (sandbox-injection-config), not to RunD.**If RunD attempts overlay without this hostPath, the mount will fail with "filesystem not supported as upperdir" (kernel rejects overlay-on-overlay since the container rootfs is itself overlayfs), effectively blocking the operation at runtime.**
Given this, there is no need to add explicit code-level gating for RunD — the missing hostPath already prevents it from working.

pkg/nas/utils.go
AlbeeSo: already remove NAS overlayfs support

pkg/mounter/proxy/server/overlay.go
AlbeeSo: Overlay does not introduce new failure modes beyond the non-overlay case:
**- mount-proxy crash (container restart):** all processes die, FUSE mount breaks — same behavior with or without overlay. The merged view becomes stale along with the underlying FUSE mount.
**- ossfs crash (mount-proxy alive):** without recovery, the FUSE mount breaks regardless of overlay. With fd-passing recovery (#1769), the FUSE kernel mount survives the process crash; after ossfs2 restarts, the lower dir recovers and the overlay merged view resumes automatically (overlay is a kernel-level layer on top of lower — if lower is functional, overlay is functional).
Currently recovery is RunC-only and overlay is Sandbox-only, so they don't intersect. Technically they are compatible — recovery support for Sandbox is blocked by a deployment constraint (two separate images), not a code limitation.

pkg/mounter/proxy/server/alinas/driver.go
AlbeeSo: NAS overlay has been removed from this PR — the alinas driver is untouched, no TerminateOverlays() call exists there now.
For context on the ossfs side (where TerminateOverlays() does run unconditionally): verified that overlay umount succeeds even after ossfs is killed (kill -9 → lower becomes "Transport endpoint is not connected" → umount overlay merged still returns exit code 0). umount is a VFS-level operation on the mount structure, independent of lower layer data availability. So the ordering is a cleanliness preference, not a strict requirement.
When NAS overlay is added in a follow-up, agreed that TerminateOverlays() should respect the same CleanupNASMountsOnExit() / reset-flag gates to preserve overlay mounts across mount-proxy restarts.

pkg/nas/nodeserver.go
AlbeeSo: 1. Dir scheme coupling: resolved — moved path computation to shared pkg/mounter/utils package.
2. Mount namespace sharing: this is not overlay-specific. The existing non-overlay flow has the same pattern — NodeUnpublish in csi-sidecar calls CleanupMountPoint(targetPath) to unmount a FUSE mount created by mount-proxy-server. Both rely on mountPropagation: Bidirectional on the shared mount-root volume. Overlay cleanup is just one more umount under the same guarantee.

pkg/utils/util.go
AlbeeSo: Using ns.skipGlobalMount directly — it indicates the VM environment (true for both Sandbox and RunD via csi-agent default). These two cannot be distinguished at code level since both use the same nodeserver with skipGlobalMount=true. The actual Sandbox-only enforcement is the hostPath volume at runtime: without it, overlay mount fails with "filesystem not supported as upperdir". This is the same layered defense discussed in comment 2 — no explicit RunningInSandbox helper needed for a single call site.
NAS removed, so only one formula remains.

pkg/mounter/proxy/server/overlay.go
mowangdk: I mean that m.overlays is an in-memory sync.Map that is never reconstructed after a process restart.

mowangdk (commented):
Reviewed with focus on maintainability/decay risks. Overall the design is clean: the overlay interceptor has a single responsibility, the protocol change is forward-compatible, and interceptor branches are well tested. Leaving 3 warnings + 4 suggestions inline; the highest-leverage fix is unifying the overlay unmount path (it resolves the tracking-drift warning and the dead-code suggestion together).

pkg/mounter/proxy/server/overlay.go
```diff
@@ -0,0 +1,80 @@
...
+	}
+	m.overlays.Delete(mergedDir)
+	klog.InfoS("Overlay unmounted", "merged", mergedDir)
+	return nil
+}
+
+// TerminateOverlays unmounts all tracked overlay mounts.
+// Errors are logged but not returned (best-effort cleanup on exit).
+// Safe to call on nil receiver.
+func (m *OverlayManager) TerminateOverlays() {
```
mowangdk: 🟡 **Warning — lifecycle asymmetry: tracking state is guaranteed to drift**

Mounts are registered into `m.overlays` here in mount-proxy-server, but the normal unmount happens in `NodeUnpublishVolume` (csi-agent unmounts the merged dir directly and calls `CleanupOverlayLowerDir`) without ever notifying the OverlayManager. So under the normal lifecycle the map is append-only, and `TerminateOverlays` will attempt cleanup on long-unmounted paths.

The reverse gap is more concrete: after a mount-proxy-server restart the map is empty, so overlays established before the restart are **not** torn down in the promised "overlay before FUSE cleanup" order — the ordering guarantee only holds for mounts made within the current process lifetime.

Suggestion: derive state from the system instead of memory — scan `/proc/mounts` for overlay mounts whose dirs live under `OverlayBaseDir` — or route unmount through the proxy via `UnmountOverlay` so tracking stays accurate.

pkg/mounter/proxy/server/overlay.go
```diff
@@ -0,0 +1,80 @@
...
+	}
+
+	m.overlays.Store(mergedDir, true)
+	klog.InfoS("Overlay mounted", "merged", mergedDir, "lower", lowerDir, "upper", upperDir)
+	return nil
+}
+
+// UnmountOverlay unmounts the overlay at mergedDir and removes it from tracking.
+// Safe to call on nil receiver (returns nil).
+func (m *OverlayManager) UnmountOverlay(mergedDir string) error {
```
mowangdk: 🟢 **Suggestion — dead code + speculative nil guards**

`UnmountOverlay` is exported but has no callers anywhere in the PR, and the `if m == nil` guards here and in `TerminateOverlays` defend a state that never occurs (the manager is always constructed in `NewDriver`). An unexercised exported API carries no real coverage and the next caller will trust code that has never been validated.

Either remove it until needed, or wire it into the unpublish path — the latter would also fix the tracking-drift issue noted on `TerminateOverlays`.

pkg/mounter/proxy/server/ossfs2/driver.go
```diff
@@ -60,14 +65,26 @@ func (h *Driver) Fstypes() []string {
 }
 
 func (h *Driver) Mount(ctx context.Context, req *proxy.MountRequest) error {
+	// When overlay is enabled, the interceptor chain handles:
+	//   1. FUSE mount to Target (= lower dir)
+	//   2. Overlay mount from lower to OverlayMerged (= original target)
+	fuseTarget := req.Target
+	overlayMerged := ""
+	if req.Overlay {
+		fuseTarget = mounterutils.OverlayLowerDir(req.Target)
```
mowangdk: 🟡 **Warning — overlay wiring duplicated line-by-line across ossfs and ossfs2**

The three new blocks (rawMounter/OverlayManager assembly in `NewDriver`, this merged→lower target rewrite in `Mount`, overlay-first teardown in `Terminate`) are identical in both drivers, so the decision "when overlay is on, swap Target for the lower dir" now lives in two places — and NAS integration will make it three.

Notably, the alinas `doMount` retry path is exactly the hazard the PR description calls out ("the retry must carry the Overlay flag, otherwise the second mount silently produces a non-overlay volume") — keeping the rewrite in each driver's hands is the breeding ground for that silent omission.

Suggestion: push the rewrite into `NewOverlayInterceptor` itself. The op already carries both `Overlay` and `OverlayMerged`, so the interceptor can do `op.Target = OverlayLowerDir(op.OverlayMerged)` and drivers only need to set two fields.

pkg/mounter/utils/utils.go
```diff
@@ -289,6 +289,58 @@ func GetPasswdHashDir(target string) string {
 	return filepath.Join("/tmp", ComputeMountPathHash(target))
 }
 
+const OverlayBaseDir = "/run/csi-overlay"
```
mowangdk: 🟡 **Warning — hard-coded base dir makes unit tests write to the host's `/run/csi-overlay`**

`TestOverlayInterceptor_FirstMount_Success` and `TestOverlayInterceptor_OverlayMountFails_CleansUpLower` reach `MountOverlay` → `OverlayDirs` → `os.MkdirAll` and create real directories under this constant — `FakeMounter` intercepts the mount syscall but not `MkdirAll`. On a non-root environment `/run` is not writable and the tests fail outright; as root they leave garbage dirs on the host and share an implicit global path across cases.

Suggestion: make the base dir injectable (a field on `OverlayManager` or a package-level `var` defaulting to `/run/csi-overlay`) and point it at `t.TempDir()` in tests.

pkg/mounter/interceptors/overlay.go
```diff
@@ -0,0 +1,69 @@
...
+//
+// Token rotation: if the merged dir is already mounted, passes through to downstream
+// handlers without attempting a second overlay mount (secret interceptor handles
+// credential update, overlay stays intact).
+//
+// This interceptor is storage-type agnostic and can be plugged into any driver's
+// interceptor chain. Currently integrated with ossfs and ossfs2.
+// NAS (alinas) is not yet integrated due to additional driver-side complexity
+// (subdir lazy creation with retry, mount-proxy restart preservation).
+func NewOverlayInterceptor(manager *server.OverlayManager) mounter.MountInterceptor {
```
mowangdk: 🟢 **Suggestion — generic interceptor layer now depends on a concrete proxy/server type**

`pkg/mounter/interceptors` was previously a generic layer consumed by the proxy server drivers; taking `*server.OverlayManager` here reverses that direction and ties the reusable pipeline to the proxy implementation (no import cycle, but it weakens the stated "storage-type agnostic, pluggable into any driver's chain" property).

Suggestion: accept a narrow interface defined in this package instead, e.g. `type overlayMounter interface { MountOverlay(merged string) error }`, so the dependency arrow points back toward the abstraction.

pkg/oss/utils.go
```diff
@@ -237,6 +237,8 @@ func parseOptions(ctx context.Context, cnfsGetter cnfsv1beta1.CNFSGetter, volOpt
 				klog.Warning(WrapOssError(ParamError, "the value(%q) of %q is invalid, only support %v, %v, %v",
 					v, k, corev1.DNSClusterFirst, corev1.DNSClusterFirstWithHostNet, corev1.DNSDefault).Error())
 			}
+		case "overlay":
+			opts.Overlay, _ = strconv.ParseBool(value)
```
mowangdk: 🟢 **Suggestion — parse error silently swallowed, inconsistent with neighboring branches**

The adjacent `dnsPolicy` case logs a warning on invalid values, but here `overlay: "yes"` silently becomes `false`: the user believes overlay is enabled, actually gets a plain read-only volume, and only finds out on the first write — with no log trail.

Suggestion: log a warning on `ParseBool` failure, matching the `dnsPolicy` style.

.gitignore
```diff
@@ -14,4 +14,5 @@ build/lvm/lvmplugin.csi.alibabacloud.com
 build/Dockerfile
 /alibaba-cloud-csi-driver
 .devcontainer/
+.agents/skills/debug-csi/.local/
```
mowangdk, line 17: 🟢 **Suggestion — local dev tooling entry unrelated to this PR**

`.agents/skills/debug-csi/.local/` looks like a personal workspace artifact; it dilutes review signal in the upstream repo. Consider dropping it from this PR and using `.git/info/exclude` or a personal global gitignore instead.

pkg/mounter/proxy/server/overlay.go
AlbeeSo: In the Sandbox scenario, NodeUnpublishVolume is never called — the VM is destroyed or frozen directly. The only unmount path is TerminateOverlays() on SIGTERM, which is symmetric with MountOverlay. So m.overlays does not drift in practice.
The CleanupOverlayLowerDir in NodeUnpublish is defensive programming for forward compatibility (future RunC where NodeUnpublish is the normal path), not exercised in the current scope.

mowangdk: NodeUnpublishVolume is never called in the current scenario is not a sufficient reason to leave the code path incomplete. If an upper layer does invoke NodeUnpublishVolume, the driver must still handle it correctly. If we do not clean up the overlay state there, entries in m.overlays may never be removed, which can lead to memory/state leakage and stale overlay references.

Even if CleanupOverlayLowerDir in NodeUnpublishVolume is defensive or forward-looking, the implementation should still be semantically correct: every publish/mount path should have a matching unpublish/unmount/cleanup path. Otherwise, the code assumes an internal invariant that is not guaranteed by the CSI interface contract.

PR #1791, opened 2026-07-29
feat(alinas): always overwrite alinas default configs on startup

myccccccc:
setupDefaultConfigs only copied a default config file when the target
  was absent (copy-if-absent). As a result, config files already present
  under /etc/aliyun/alinas (persisted on the host via the bind-mounted
  /etc/cnfs/alinas) were never refreshed, so config updates shipped in a
  new image never took effect after a restart.

  Always overwrite files under the alinas config dir with the defaults
  shipped in the image so image config updates apply on restart. cpfs
  keeps the existing copy-if-absent behavior.

Fixes #

```release-note

```

```docs

```

### DIFF
--- a/pkg/mounter/proxy/server/alinas/driver.go
+++ b/pkg/mounter/proxy/server/alinas/driver.go
@@ -163,6 +163,10 @@ func setupDefaultConfigs() {
 	for _, name := range []string{"cpfs", "alinas"} {
 		srcDir := filepath.Join(defaultConfigDir, name)
 		dstDir := filepath.Join(configDir, name)
+		// The alinas config is always overwritten with the default shipped in
+		// the image so that config updates from a new image take effect on
+		// restart. cpfs keeps the copy-if-absent behavior.
+		overwrite := name == "alinas"
 		if err := filepath.WalkDir(srcDir, func(path string, d fs.DirEntry, err error) error {
 			if err != nil {
 				return err
@@ -179,11 +183,13 @@ func setupDefaultConfigs() {
 
 			dstPath := filepath.Join(dstDir, relPath)
 
-			if _, err := os.Stat(dstPath); err == nil {
-				// File already exists, skip
-				return nil
-			} else if !os.IsNotExist(err) {
-				return err
+			if !overwrite {
+				if _, err := os.Stat(dstPath); err == nil {
+					// File already exists, skip
+					return nil
+				} else if !os.IsNotExist(err) {
+					return err
+				}
 			}
 
 			klog.InfoS("Copying default config file", "path", dstPath)


huww98:
Can user also modify these config, and we risk overwrite user's custom config?

huww98:
I dug into what actually lives under `/etc/aliyun/{alinas,cpfs}` at runtime (inspected the `aliyun-alinas-utils` and `alinas-efc` RPM payloads plus the watchdog/mount scripts) to check whether this overwrite is safe. Summary: **the alinas overwrite is safe**, but there are two things worth addressing.

**The overwrite is safe for alinas.** The files seeded into `/etc/aliyun-defaults/alinas` are exactly the RPM-shipped ones (`alinas-utils.conf`, `alinas-utils.crt`, the `*.json`, the `log_conf.efc.*` templates). All of them are *read-only* at runtime — the watchdog only ever `read_config()`s `alinas-utils.conf`, and the log_conf files are just copy-source templates. The genuinely dynamic files the watchdog writes (`privateKey.pem`, `.credentials`, `last-mountpoint`, `nas-agent-commands-remote/`) are generated directly into the live dir and are **not** present in the defaults dir, so `WalkDir` never touches them. So blanket-overwriting the alinas defaults can't clobber TLS keys / credentials / state.

One semantic note: `alinas-utils.conf` is marked `%config(noreplace)` in the RPM, i.e. the vendor intends host-side edits to survive upgrades. Overwriting it unconditionally is the opposite of that — fine here because nothing but the image writes it, but it does mean any operator edit to alinas configs on the host would be silently reverted on restart. Probably worth a one-line comment that host-side edits to these configs aren't supported.

**cpfs is the interesting asymmetry.** I couldn't find *any* writer to `/etc/aliyun/cpfs` anywhere in the payload — `aliyun-cpfs-mount-watchdog`, `mount.cpfs-nfs`, and `switch_server` all only *read* `cpfs-utils.conf` / `alinas-utils.crt`; cpfs runtime state goes to `/var/run/cpfs` and logs to `/var/log/aliyun/cpfs`. So cpfs is fully read-only under `/etc/aliyun`. That makes the Dockerfile comment ("alinas-utils generates some dynamic configuration files ... under /etc/aliyun/{alinas,cpfs}") inaccurate for cpfs, and it means leaving cpfs as copy-if-absent reproduces the exact stale-config bug this PR fixes for alinas — cpfs config updates shipped in a new image will silently never apply, since `/etc/aliyun/cpfs` is a persisted host subPath in the cnfs-nas-daemon manifest.

There are two coherent end-states; the PR currently sits between them:
- **(A)** stop persisting `/etc/aliyun/cpfs` (leave it in the container rootfs, so it's always the image version and needs no seeding), keep persisting `/etc/aliyun/alinas` and overwrite it — but this needs the cnfs-nas-daemon daemonset manifest changed to drop the cpfs mount.
- **(B)** keep both mounts and overwrite both — smallest change here, fixes stale config for both, and it's safe since neither dir's shipped files are runtime-written.

Since cpfs is read-only, overwriting it carries the same low risk as alinas. Could you clarify whether cpfs config is shipped/updated the same way alinas is? If yes, I'd overwrite cpfs too (B) for consistency; if cpfs is deliberately host-managed, the split is intentional but deserves a code comment explaining *why* `overwrite := name == "alinas"`, because right now it reads as an arbitrary distinction.

**Copy is non-atomic while a reader is polling.** `Init()` starts the two watchdogs *before* `setupDefaultConfigs()`, and the watchdog re-reads `alinas-utils.conf` every poll. `copyFile` does `os.Create` (truncate) → `io.Copy` → `Sync`, so there's a brief window where the watchdog can read a truncated/half-written config. Copy-if-absent only exposed this on first boot; overwriting every restart makes it recur on each restart. Rather than making `copyFile` atomic, the simplest fix is to **swap the order in `Init()`** — run `setupDefaultConfigs()` first, then spawn the watchdogs:

```go
func (h *Driver) Init() {
	setupDefaultConfigs()
	go runCommandForever("aliyun-alinas-mount-watchdog")
	go runCommandForever("aliyun-cpfs-mount-watchdog")
}
```

The watchdogs are the only concurrent readers (mount requests can't arrive until after `server.Init()` returns and `srv.Serve()` starts), so seeding before they launch closes the window entirely — and it also fixes the pre-existing first-boot race, and makes the `panic(err)` path abort startup cleanly instead of leaving orphaned watchdog goroutines.

huww98:
/close
#1792 merged

PR #1792, opened 2026-07-29
Alinas always overwrite default config

bingrenlyyyyy:
setupDefaultConfigs only copied a default config file when the target
was absent (copy-if-absent). As a result, config files already present
under /etc/aliyun/alinas (persisted on the host via the bind-mounted
/etc/cnfs/alinas) were never refreshed, so config updates shipped in a
new image never took effect after a restart.

Always overwrite files under the alinas config dir with the defaults
shipped in the image so image config updates apply on restart. cpfs
keeps the existing copy-if-absent behavior.


huww98:
could replace #1791

PR #1795, opened 2026-07-30
feat(bmcpfs): delay detach on unstage for possible reuse

huww98:
#### What type of PR is this?

/kind feature

The bmcpfs node plugin previously did not implement STAGE_UNSTAGE_VOLUME. This PR advertises the capability and adds a NodeUnstageVolume that, when the `BMCPFS_DETACH_DELAY` environment variable is set, defers reporting success until the delay has elapsed since the first unstage attempt. This keeps the volume attached for a window so that a returning pod can reuse the existing mount instead of paying a full detach + re-attach + re-mount cycle, which is expensive for CPFS.

The delay is enforced by a persisted per-volume start time rather than by parking the RPC. Each NodeUnstageVolume call returns after at most 1s with `DeadlineExceeded`, so a returning pod is never blocked behind an in-flight unstage — kubelet serializes mount and unmount on the same per-volume operation key, so a long in-flight unstage would otherwise delay the reuse it exists to enable. The start time survives across kubelet's retries, so the total delay stays bounded to `detachDelay` regardless of how many retries it takes. `NodeStageVolume` clears any leftover start time so a reused volume waits a full fresh delay on its next detach.

Fixes #

`BMCPFS_DETACH_DELAY` is unset by default, so behavior is unchanged unless explicitly enabled; an unparseable value fails node-server startup rather than silently disabling the feature. It is intended for delays on the order of minutes. Because the design relies on kubelet's retry loop, the actual detach happens somewhere in `[t0 + detachDelay, t0 + detachDelay + kubelet_retry_interval]`, and each teardown logs `UnmountDevice failed: DeadlineExceeded` periodically for the duration of the delay — this is expected operation, not an error.

The timing-sensitive behavior is covered by unit tests using `testing/synctest` (fake clock), so they verify the full multi-minute delay deterministically and instantly.

```release-note
bmcpfs: added STAGE_UNSTAGE_VOLUME support with an optional `BMCPFS_DETACH_DELAY` to defer detach after unmount so a returning pod can reuse the existing mount.
```

```docs

```

### DIFF
--- a/pkg/bmcpfs/nodeserver.go
+++ b/pkg/bmcpfs/nodeserver.go
@@ -26,6 +26,8 @@ import (
 	"os"
 	"slices"
 	"strings"
+	"sync"
+	"time"
 
 	"github.com/container-storage-interface/spec/lib/go/csi"
 	"github.com/kubernetes-sigs/alibaba-cloud-csi-driver/pkg/cloud/metadata"
@@ -42,6 +44,9 @@ type nodeServer struct {
 	common.GenericNodeServer
 	mounter mount.Interface
 	locks   *utils.VolumeLocks
+
+	detachDelay      time.Duration
+	unstageStartTime sync.Map // map[volumeID]time.Time
 }
 
 const (
@@ -49,7 +54,7 @@ const (
 	metricsPathPrefix             = "/run/cnfs/efc/"
 )
 
-func newNodeServer(meta *metadata.Metadata) (*nodeServer, error) {
+func newNodeServer(meta metadata.MetadataProvider) (*nodeServer, error) {
 	var nodeID string
 	data, err := os.ReadFile(metadata.LingjunConfigFile)
 	if err != nil {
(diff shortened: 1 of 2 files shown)


pkg/bmcpfs/nodeserver.go
```diff
@@ -83,13 +88,61 @@ func newNodeServer(meta *metadata.Metadata) (*nodeServer, error) {
 	}
 	klog.Infof("bmcpfsplugin nodeId: %s", nodeID)
 	mounter := mounter.NewProxyMounter(defaultAlinasMountProxySocket, mount.NewWithoutSystemd(""))
+
+	var detachDelay time.Duration
```
mowangdk, line 92: Please add unittest for this

huww98: Added

huww98:
Manually verified end-to-end on a live cluster (cn-beijing, plain-ECS `common:` nodes → VPC mount path, `BMCPFS_DETACH_DELAY=2m`, EFC/`fuse.aliyun-alinas-efc` mount via cnfs-nas-daemon). All three behaviors observed in the node-plugin logs:

**Fast return** — each `NodeUnstageVolume` returned `DeadlineExceeded` after exactly ~1s and kubelet retried with backoff; the mount stayed attached, no detach.

**Reuse within the window** — deleting the pod and recreating it before the delay elapsed triggered `NodeStageVolume` (which drops the persisted start time), then `NodePublishVolume` re-mounted in ~0.4s. The file written before deletion was still present, confirming no real detach/re-attach/re-mount happened. Cross-checked the controller logs: across the whole reuse cycle there was **no `ControllerUnpublishVolume` and no second `ControllerPublishVolume`** — only node-side RPCs. Because the delayed node-side unstage never reports success, the VolumeAttachment stays `true`, kubelet never sees the volume as detached, and the attach/detach controller is never involved. That's the whole point: a returning pod re-stages against the still-attached volume with zero controller round-trip.

**Full delay expiry** — with no reuse, the first unstage recorded the start time at `08:44:40`, so the deadline was `08:46:40`. Retries kept returning `DeadlineExceeded`; the first retry landing *after* the deadline (`08:46:52`) returned success and only then did `ControllerUnpublishVolume` fire (`08:46:54`) and the VolumeAttachment detach (`08:46:55`). Total delay stayed bounded to `detachDelay` regardless of the ~10 retries. The small gap between nominal deadline and actual success is kubelet's exponential retry backoff (retries space out to ~1min near the end), not driver latency — the driver succeeds on the first retry past the deadline as designed.

One operational note surfaced during setup (not a code issue): the bmcpfs init container hard-fails (`ERROR: fuse connection recovery not supported`) on nodes without the enhanced `kmod-fuse` (checks `/proc/kallsyms` for `fuse_dev_ioctl_recover`). Installing kmod-fuse per the ECS/EGS CPFS mount guide resolves it. Worth keeping in mind before enabling bmcpfs fleet-wide.
https://help.aliyun.com/zh/cpfs/bmcpfs/user-guide/mount-cpfs-for-lingjun-on-ecs-egs

PR #1812, opened 2026-08-12
feat(oss): support AgenticBucket/BucketSpace mounting

AlbeeSo:
#### What type of PR is this?

/kind feature
### Summary

Add CSI support for OSS AgenticBucket two-layer architecture, enabling per-sandbox isolated storage via BucketSpace.

### Changes

- **New volumeAttributes**: `agenticBucket`, `bucketSpace`, `bucketSpacePrefix`
- **MountBucket()**: resolves effective bucket (BucketSpace or legacy Bucket)
- **resolveBucketSpacePrefix**: composes full BucketSpace name from prefix + accountId + region
- **Mutual exclusion validation**: agenticBucket/bucketSpace and bucket cannot coexist
- **ossfs/ossfs2 mount options**: `auto_create_bucket` + `agentic_bucket=<name>`
- **volumeAs=bucketspace**: dynamic provisioning blocked (63-char naming limit)

### How it works

1. PV declares `agenticBucket` (the parent AgenticBucket name) and `authType: agent-identity`
2. SandboxClaim dynamically injects `bucketSpace` (full name) or `bucketSpacePrefix` (resolved at mount time)
3. CSI NodePublish passes `auto_create_bucket` and `agentic_bucket=<name>` to ossfs
4. ossfs creates BucketSpace if not exists (PutBucket with AgenticBucketName header), then mounts it

### Testing

E2E verified in sandbox environment (cluster `c8bc78a08095d4d8bbf8e723468333b6a`, cn-shanghai):

| Case | Result |
|------|--------|
| name mode (bucketSpace=full name) | ✅ mount + read/write |
| prefix mode (bucketSpacePrefix) | ✅ resolves full name, mount + read/write |
| prefix + region | ✅ |
| name + subPath | ✅ mount to subdirectory, read/write |
| prefix + subPath | ✅ |
| pause/resume (hibernate/wake) | ✅ data persists |
| remount (BucketSpace already exists) | ✅ |

### Dependencies

- ossfs `ack/feat/agentic-bucket` branch (commit `a1229db`): adds `auto_create_bucket` + `agentic_bucket` mount options
- sandbox-controller `feat/agentic-bucket` branch: enricher injects `bucketSpace`/`bucketSpacePrefix` into volumeAttributes and storage-auth annotation

Fixes #

```release-note

```

```docs

```

### DIFF
--- a/pkg/mounter/fuse_pod_manager/oss/oss_fuse_pod_manager.go
+++ b/pkg/mounter/fuse_pod_manager/oss/oss_fuse_pod_manager.go
@@ -72,9 +72,15 @@ type Options struct {
 	CNFSName       string
 
 	// oss options
+	// normal bucket
 	Bucket string `json:"bucket"`
-	URL    string `json:"url"`
-	Path   string `json:"path"`
+	// agentic bucket
+	AgenticBucket     string `json:"agenticBucket"`
+	BucketSpace       string `json:"bucketSpace"`
+	BucketSpacePrefix string `json:"bucketSpacePrefix"`
+
+	URL  string `json:"url"`
+	Path string `json:"path"`
 
 	// authorization options
 	// accesskey
@@ -114,4 +120,16 @@ type Options struct {
 
 	// pod template
 	DnsPolicy corev1.DNSPolicy `json:"dnsPolicy"`
+}
+
+// MountBucket returns the effective bucket name for mounting.
+// BucketSpace and Bucket are mutually exclusive (enforced by validation).
+// In AgenticBucket mode, BucketSpace is used; otherwise Bucket.
+// If BucketSpacePrefix was specified, it has already been resolved into BucketSpace
+// by resolveBucketSpacePrefix before this method is called.
+func (o *Options) MountBucket() string {
+	if o.BucketSpace != "" {
+		return o.BucketSpace
+	}
+	return o.Bucket
 }
--- a/pkg/mounter/fuse_pod_manager/oss/ossfs2/manager.go
+++ b/pkg/mounter/fuse_pod_manager/oss/ossfs2/manager.go
@@ -141,7 +141,7 @@ func (f *fuseOssfs) MakeAuthConfig(o *ossfpm.Options, m metadata.MetadataProvide
 func (f *fuseOssfs) MakeMountOptions(o *ossfpm.Options, m metadata.MetadataProvider) (mountOptions []string, err error) {
 	mountOptions = append(mountOptions,
 		"oss_endpoint="+o.URL,
-		"oss_bucket="+o.Bucket,
+		"oss_bucket="+o.MountBucket(),
 		"oss_bucket_prefix="+o.Path,
 	)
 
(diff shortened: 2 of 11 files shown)


pkg/mounter/fuse_pod_manager/oss/ossfs2/manager.go
```diff
@@ -163,6 +163,14 @@ func (f *fuseOssfs) MakeMountOptions(o *ossfpm.Options, m metadata.MetadataProvi
 		mountOptions = append(mountOptions, fmt.Sprintf("oss_region=%s", region))
 	}
 
+	// AgenticBucket / BucketSpace support
+	if o.AgenticBucket != "" {
+		mountOptions = append(mountOptions,
+			"auto_create_bucket=true",
```
mowangdk, line 169: Do the parameters require an additional RAM policy? If they do, how are users granted the required permissions?

pkg/mounter/fuse_pod_manager/oss/oss_fuse_pod_manager.go
```diff
@@ -115,3 +121,15 @@ type Options struct {
...
 }
+
+// MountBucket returns the effective bucket name for mounting.
+// BucketSpace and Bucket are mutually exclusive (enforced by validation).
+// In AgenticBucket mode, BucketSpace is used; otherwise Bucket.
+// If BucketSpacePrefix was specified, it has already been resolved into BucketSpace
+// by resolveBucketSpacePrefix before this method is called.
+func (o *Options) MountBucket() string {
+	if o.BucketSpace != "" {
+		return o.BucketSpace
```
mowangdk, line 132: If this is a bucket space, do we no longer need a regular bucket in the following steps?

pkg/mounter/fuse_pod_manager/oss/ossfs2/manager.go
AlbeeSo: Yes, `auto_create_bucket=true` will cause ossfs to call `oss:PutBucket` to create the BucketSpace. The required permissions need to be configured in the **CredentialProvider** CR (SessionPolicy). Specifically, a scoped `oss:PutBucket` permission is needed, where the Resource is rendered to the specific BucketSpace name by the identity provider (the consumer of the CredentialProvider CR), and the `oss:AgenticBucketName` condition further scopes it to the AgenticBucket. The RAM role policy also needs the corresponding PutBucket permission as the upper bound.

pkg/mounter/fuse_pod_manager/oss/oss_fuse_pod_manager.go
AlbeeSo: `Bucket` and `BucketSpace` are functionally equivalent at the mount API level. All downstream code uses a unified `MountBucket()` accessor that returns `BucketSpace` if set, otherwise `Bucket`. There is no "subsequent step that still needs a regular bucket" — once `BucketSpace` is set, it fully replaces `Bucket` for the entire mount flow.

The reason we use separate parameters is to avoid user confusion and enable clear validation: `bucket` and `bucketSpace`/`bucketSpacePrefix` are mutually exclusive, and `agenticBucket` requires one of the BucketSpace variants.

PR #1817, opened 2026-08-16
Fix/nas unmount via broker by target

mowangdk:
#### What type of PR is this?
/kind bug

##### Commit 1 — route alinas unmount through mount broker by ownership
Why NAS umount takes ~3 seconds (root cause)
NAS AccessPoint volumes are NFSv3-over-TLS. At mount time mount.alinas runs mount.nfs, and nfs-utils/libmount records the mount in the userspace mount table /run/mount/utab. That entry stores the info the kernel mount table does not keep but that umount.nfs needs to send the NFSv3 UMNT RPC to mountd — notably addr, port=xxxx, mountport, proto, clientaddr.

Crucially:
- /run/mount/utab is per-mount-namespace (each container has its own /run tmpfs); it is not shared across namespaces.
- Mount is delegated to the broker (mount-proxy-server): NasMounter.ExtendedMount sends the mount over the proxy socket, so mount.nfs runs inside the broker's mount namespace, and the utab entry is written there.
- The old unmount ran locally in the csi-plugin container's mount namespace (NasMounter inherited the embedded mount-utils Unmount). That namespace's /run/mount/utab never had the entry (it didn't perform the mount, and utab isn't shared).

So umount.nfs finds no utab record, falls back to contacting mountd on 127.0.1.x:xxx to send the UMNT RPC. The socket is opened non-blocking; the connect returns EINPROGRESS and umount.nfs waits in select() for the socket to become writable. The RPC gets no usable response, so select() returns only on its built-in 3×1s timeout ≈ 3.0s (the kernel umount(2) itself is microseconds; sendto is never even reached).

This isolates the cause to utab, independent of any network/nftables factor.
Note on the csi_mount_proxy nftables rule (oif lo tcp dport xxxx meta cgroup != 0 drop): in cgroup-v2 deployments where the plugin container is in a non-root cgroup, that rule also drops the fallback UMNT RPC and produces the same 3s stall. But it is a secondary amplifier, not the primary cause — the utab-namespace mismatch alone reproduces the 3s even when the rule does not match (e.g. root cgroup).

Compatibility / fallback
- Old mount-proxy-server (no unmount RPC): replies invalid method → local unmount (no regression).
- Plain NFS / non-broker mounts: broker has no record → ErrTargetNotManaged → local unmount (correct).
- Agent/connector mode (no ProxyUnmounter): always local.
- Broker restart (in-memory targets lost): affected old mounts fall back to local unmount; acceptable edge case.
Validation
- End-to-end: NodePublishVolume's subpath-creation cleanup now logs unmounted successfully via mount broker and completes in ~9ms (was ~3010ms via the local path).
- Unit tests added for broker-side routing (owned / not-owned / error / empty target / driver-without-Unmounter) and NasMounter routing (broker success / fallback-to-local / non-proxy mode).

##### Commit 2 — chmod auto-created subpath to 0777
The subpath directory auto-created during NodePublishVolume (via the temporary subpath-creation mount) was made with os.MkdirAll(..., os.ModePerm), which is subject to the process umask and therefore usually ended up as 0755. Workloads running as arbitrary/non-root UIDs then could not write into their own subpath.
Fix: after MkdirAll, explicitly os.Chmod(subDir, 0o777) (chmod is not affected by umask) so the auto-created subpath is world-writable by default.

Fixes #

```release-note
NAS: fix NodeUnpublishVolume taking ~3s per alinas/AccessPoint mount by routing the unmount through the mount broker (the namespace that performed the mount and holds the libmount utab entry); falls back to a local unmount when the broker does not own the target or is unavailable.

Action required / behavior change: the directory auto-created for a NAS subpath volume is now chmod'd to 0777 (previously it was created with MkdirAll under the process umask and typically ended up 0755). Only newly auto-created subpath directories are affected; existing directories are not modified.
```

```docs

```

### DIFF
--- a/pkg/mounter/mounter.go
+++ b/pkg/mounter/mounter.go
@@ -19,6 +19,24 @@ type Mounter interface {
 	ExtendedMount(ctx context.Context, op *MountOperation) error
 }
 
+// ProxyUnmounter is implemented by mounters that can unmount through the mount
+// broker (mount-proxy-server) instead of unmounting locally. This is required
+// for NAS AccessPoint mounts: the broker runs in cgroup 0, the only cgroup
+// allowed to reach the mount broker port (tcp 12049) under the csi_mount_proxy
+// nftables rule. A local umount from a container cgroup would be dropped.
+//
+// ExtendedUnmount returns ErrTargetNotManagedByBroker when the broker has no
+// record of target (it was not mounted through the broker), so the caller can
+// fall back to a local unmount.
+type ProxyUnmounter interface {
+	ExtendedUnmount(ctx context.Context, target string) error
+}
+
+// ErrTargetNotManagedByBroker is returned by ProxyUnmounter.ExtendedUnmount when
+// the mount broker reports it does not own the target (proxy.ErrTargetNotManaged)
+// or when the broker predates the unmount RPC. Callers fall back to local unmount.
+var ErrTargetNotManagedByBroker = errors.New("target not managed by mount broker")
+
 type MountOperation struct {
 	Source      string
 	Target      string
--- a/pkg/mounter/proxy/client/client.go
+++ b/pkg/mounter/proxy/client/client.go
@@ -101,6 +101,15 @@ func (c *client) Mount(ctx context.Context, req *proxy.MountRequest) (*proxy.Res
 	})
 }
 
+func (c *client) Unmount(ctx context.Context, req *proxy.UnmountRequest) (*proxy.Response, error) {
+	return c.doRequest(ctx, &proxy.Request{
+		Header: proxy.Header{
+			Method: proxy.Unmount,
+		},
+		Body: req,
+	})
+}
+
 func (c *client) Ping(ctx context.Context) (*proxy.Response, error) {
 	return c.doRequest(ctx, &proxy.Request{
 		Header: proxy.Header{
--- a/pkg/mounter/proxy/protocol.go
+++ b/pkg/mounter/proxy/protocol.go
@@ -17,9 +17,17 @@ const (
 type Method string
 
 const (
-	Mount Method = "mount"
-	Ping  Method = "ping"
+	Mount   Method = "mount"
+	Unmount Method = "unmount"
+	Ping    Method = "ping"
 )
+
+// ErrTargetNotManaged is the error string returned by mount-proxy-server when an
+// Unmount request targets a mount point that was NOT mounted through the broker
+// (i.e. the broker has no record of it). The client uses this sentinel to fall
+// back to a local unmount. It is intentionally matched by string because the
+// error crosses the proxy socket as plain text (see Response.Error).
+const ErrTargetNotManaged = "target not managed by mount broker"
 
 type Header struct {
 	Method Method `json:"method,omitempty"`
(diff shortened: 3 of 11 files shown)


mowangdk:
### End-to-end test results

Tested on an ACS Agent Sandbox cluster with `--enable-nftables=true`, images built from this PR:
- `csi-plugin` (csi-sidecar) — contains both commits (unmount-via-broker + subpath chmod 0777)
- `csi-agent` (mount-proxy-server) — contains the broker-side `Unmount` RPC handler

#### 1) Mounting a non-existent subPath (triggers subpath auto-creation)

Fresh `subPath` that does not exist on the NAS filesystem. Full NodePublishVolume timeline from the csi-sidecar log:

| time | step | delta |
|------|------|-------|
| 03:52:10.593 | NodePublishVolume start, mount `server:/sp-new-52328` | — |
| 03:52:12.574 | first mount fails `exit status 32` (subpath missing) | ~1.98s |
| 03:52:13.416 | root mounted (temporary subpath-creation mount) | ~0.84s |
| 03:52:13.416 → .430 | mkdir + chmod 0777 the subpath | ~14ms |
| 03:52:13.430 → **.439** | **unmount temp mount → `unmounted successfully via mount broker`** | **~9ms** |
| 03:52:14.246 | re-mount `server:/sp-new-52328` succeeds | ~0.8s |
| 03:52:14.247 | `Mount success` | — |

**Total ≈ 3.65s.** The key point: the temporary-mount cleanup now unmounts **via the broker in ~9ms**. Before this PR the same step ran a local `umount.nfs` whose `/run/mount/utab` had no entry (mount was performed in the broker's mount namespace, utab is per-namespace), so it fell back to the mountd UMNT RPC and its `select()` timed out ~3s — i.e. this path used to cost ~3s more.

Relevant log lines:
```
utils.go:167  trying to create subpath /sp-new-52328 in ap-...nas.aliyuncs.com
mounter.go:39 "mounted successfully" target=".../subpath-creation_..._1565715693" fstype="alinas"
mounter.go:81 "unmounted successfully via mount broker" target=".../subpath-creation_..._1565715693"
mounter.go:39 "mounted successfully" source="...:/sp-new-52328" ...
nodeserver.go:564 NodePublishVolume:: ... Mount success
```

#### 2) Permission of the auto-created directory

```
$ stat -c '%A %a %U:%G' <mountpoint>
drwxrwxrwx 777 root:root
$ touch <mountpoint>/write-test.txt   # OK
```
The auto-created subpath is **0777** (chmod is not affected by umask). Before this PR it was created via `MkdirAll(..., os.ModePerm)` and, subject to the process umask, typically ended up 0755.

#### Control experiment confirming the root cause (utab, not nftables)

On a live AccessPoint mount, in the **same mount namespace and same cgroup**, the only variable changed was the utab entry:

| utab state | `umount` wall-clock |
|---|---|
| entry present (written by broker at mount time) | ~9 ms |
| entry removed (`: > /run/mount/utab`) | **~3011 ms** |

This isolates the ~3s to the missing utab record (umount.nfs falling back to the mountd UMNT RPC), independent of nftables/cgroup. In this ECI environment the containers run in the root cgroup, so the `csi_mount_proxy` `cgroup != 0` rule does not even match — yet the 3s still reproduces purely from the utab-namespace mismatch, which is what this PR fixes by unmounting in the same namespace that performed the mount.

PR #1822, opened 2026-08-19
Chmod to 777 for agent expecially

mowangdk:
#### What type of PR is this?
/kind cleanup

MkdirAll honors the process umask, so the created subpath would typically end up as 0755. Explicitly chmod to 0777 for agent mode (rootless container are required)

Fixes #

```release-note

```

```docs

```


huww98:
I'd still stand against chmod to 777, even only for agent.

If a normal user mount the same NAS (root path, in other Pods, or on ECS directly), then this is a behaviour change (may be security hole) for him. previously he can only read the file written to that dir, now he can write.

But agent only is better than unconditional.
/lgtm

PR #1826, opened 2026-08-24
fix(mount-proxy): make FUSE mount failures diagnosable

AlbeeSo:
#### What type of PR is this?

/kind feature
A mount that never completes used to surface as a bare deadline: ossfs kept running until the connection timeout in Handle expired, and by then the caller had already given up, so the process stderr never reached it.

- Bound a single ossfs/ossfs2 attempt with FuseMountTimeout, short enough that the mount error reaches the caller before its own deadline. The connection timeout is left alone because alinas mounts ignore ctx, so shortening the shared value would only shrink their write window.
- Report the timeout with elapsed time, pid and mountpoint instead of a bare wait error.
- Fix the inverted os.ErrProcessDone check that muted every real kill failure in the ossfs2 fallback.
- Keep the signal/kill failure cause in the log instead of discarding it.
- Cover ExtendedMount with tests for the stderr, timeout, success and startup-failure paths, via a mount command seam. ossfs2 had none.

Fixes #

```release-note

```

```docs

```

### DIFF
--- a/cmd/mount-proxy-server/main.go
+++ b/cmd/mount-proxy-server/main.go
@@ -15,6 +15,7 @@ import (
 	flag "github.com/spf13/pflag"
 	"golang.org/x/sys/unix"
 
+	"github.com/kubernetes-sigs/alibaba-cloud-csi-driver/pkg/mounter/proxy"
 	"github.com/kubernetes-sigs/alibaba-cloud-csi-driver/pkg/mounter/proxy/server"
 	_ "github.com/kubernetes-sigs/alibaba-cloud-csi-driver/pkg/mounter/proxy/server/alinas"
 	_ "github.com/kubernetes-sigs/alibaba-cloud-csi-driver/pkg/mounter/proxy/server/ossfs"
@@ -34,7 +35,8 @@ var (
 func main() {
 	flag.StringVar(&socketPath, "socket", "/var/run/csi/mounter.sock", "socket path")
 	flag.StringSliceVar(&drivers, "driver", nil, "drivers to enable (e.g. 'ossfs,alinas')")
-	flag.DurationVar(&handleTimeout, "timeout", time.Second*30, "timeout for connection")
+	flag.DurationVar(&handleTimeout, "timeout", proxy.DefaultConnectionTimeout,
+		"how long a connection has to deliver a request and receive its answer; also caps a mount when shorter than what the client asks for")
 	flag.BoolVar(&enableNftables, "enable-nftables", false, "enable nftables rules to restrict mount proxy port access (default: false)")
 	flag.BoolVar(&cleanupNASMounts, "cleanup-nas-mounts-on-exit", false, "unmount all NAS mount points inside the pod on SIGTERM")
 	utils.AddKlogFlags(flag.CommandLine)
--- a/pkg/mounter/proxy/client/client.go
+++ b/pkg/mounter/proxy/client/client.go
@@ -12,11 +12,6 @@ import (
 	"k8s.io/klog/v2"
 )
 
-const (
-	// this should be longer than default timeout in server
-	defaultTimeout = time.Second * 35
-)
-
 type client struct {
 	timeout time.Duration
 	raddr   net.UnixAddr
@@ -26,7 +21,7 @@ type client struct {
 func NewClient(socketPath string) *client {
 	return &client{
 		raddr:   net.UnixAddr{Name: socketPath, Net: "unix"},
-		timeout: defaultTimeout,
+		timeout: proxy.ClientFallbackTimeout,
 	}
 }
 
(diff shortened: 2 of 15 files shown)


pkg/mounter/proxy/server/fuse_mount_timeout.go
```diff
@@ -0,0 +1,25 @@
...
+// It is deliberately shorter than the connection timeout in Handle so the
+// caller receives the mount error instead of a bare deadline. The sandbox
+// caller (agents storage-cli gRPC and CSIMount) allows 30s end to end, and a
+// timed-out attempt spends up to 2 more seconds on the SIGTERM grace period
+// before the response is written, so the budget has to leave room for both.
+//
+// The connection timeout is intentionally left untouched: alinas mounts run
+// through mount-utils and ignore ctx entirely, so shortening the shared
+// timeout would only shrink their response write window.
+const FuseMountTimeout = 25 * time.Second
```
mowangdk: The caller timeout should be configurable and dynamically adjusted based on the remaining time in the context (ctx).

pkg/mounter/proxy/server/fuse_mount_timeout_test.go
```diff
@@ -0,0 +1,43 @@
...
+import (
+	"context"
+	"testing"
+	"time"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+func TestWithFuseMountTimeout(t *testing.T) {
```
mowangdk: With dynamic timeouts based on ctx.Deadline() and can be resued among different driver, we can decouple this into a separate configuration file.

pkg/mounter/proxy/server/fuse_mount_timeout.go
AlbeeSo: Done — the caller's deadline is now what drives it, and there is no separate timeout for a driver to be given.

`client.Send` states the deadline it is working to on the request, taking it from the context when the caller set one:

```go
if deadline, ok := ctx.Deadline(); ok {
    req.Header.DeadlineUnixNano = deadline.UnixNano()
    ...
}
```

The server derives the mount's deadline from that rather than from a constant of its own (`proxy.ResolveWorkDeadline`):

```go
if stated.Sub(now) <= DeadlineReserve {
    // Reporting is the point, so when there is not enough time for both,
    // spend none of it mounting rather than shorten the reserve.
    return now
}
if deadline := stated.Add(-DeadlineReserve); deadline.Before(connDeadline) {
    return deadline
}
return connDeadline
```

`DeadlineReserve` is what makes this worth doing: the mount has to end early enough to still say why it ended. It is `MountShutdownGrace + 1s` — the time the driver needs between SIGTERM and SIGKILL, plus a second to write the response — so a mount that runs out of time is reported as a timeout rather than the caller seeing its connection close on it.

A caller with no deadline falls back to `ClientFallbackTimeout`, so the server always has something to subtract from.

`pkg/mounter/proxy/integration/deadline_test.go` holds this: a driver that hangs still produces `ossfs2 mount timeout after ...` at the caller. I also ran it against a real cluster by pointing the credential endpoint at a blackhole address, which gave `ossfs2 mount timeout after 26s, pid=463, mountpoint=..., process terminated with SIGTERM, check network connectivity` — 26s of the caller's 30s, the rest being the reserve.

pkg/mounter/proxy/server/fuse_mount_timeout_test.go
AlbeeSo: Done — `fuse_mount_timeout.go` and this test are gone. Everything about deadlines now lives in `pkg/mounter/proxy/deadline.go`, which both drivers and the server share:

```go
DefaultConnectionTimeout = 30 * time.Second
ResponseWriteWindow      = 5 * time.Second
MountShutdownGrace       = 2 * time.Second
DeadlineReserve          = MountShutdownGrace + time.Second
ClientFallbackTimeout    = 35 * time.Second
```

Used by `server/handler.go`, `server/ossfs/driver.go` and `server/ossfs2/driver.go`, so ossfs and ossfs2 no longer each carry their own idea of how long a mount may take.

They are in one file for a reason worth stating: they only make sense relative to each other. `DeadlineReserve` has to cover `MountShutdownGrace`, and `ClientFallbackTimeout` has to exceed `DefaultConnectionTimeout` or the server would cut a mount short that the client still considers live. Defining them together makes those relations expressible — `DeadlineReserve = MountShutdownGrace + time.Second` rather than two numbers that must be kept in step by hand — and the file carries the four-hop timeline they belong to.

The tests split the same way: `deadline_test.go` for the arithmetic, `client/deadline_test.go` for what the client states, `server/deadline_test.go` for what the server derives, and `integration/deadline_test.go` for a hanging driver end to end.

PR #1828, opened 2026-08-25
ens: allow overriding the OpenAPI endpoint via ENS_ENDPOINT

zyjhtangtang:
The endpoint cannot be derived from RegionID since RegionID is itself resolved through DescribeInstance, so the international site needs an explicit override.

Fixes #

```release-note

```

```docs

```


mowangdk:
nit: We need update the ENS YAML to align with these changes.

zyjhtangtang:
> nit: We need update the ENS YAML to align with these changes.

done

PR #1832, opened 2026-09-01
feat(mount-proxy): add --init-nas-rsa-pem to generate NAS RSA private key on startup

mowangdk:
## What this PR does

Adds a new `mount-proxy-server` flag `--init-nas-rsa-pem` (default `false`). When enabled, the `alinas` driver generates the NAS RSA private key during `Init()`, equivalent to:

```
mkdir -p /etc/aliyun/alinas
openssl genpkey -algorithm RSA -out /etc/aliyun/alinas/privateKey.pem -pkeyopt rsa_keygen_bits:3072
chmod 400 /etc/aliyun/alinas/privateKey.pem
```

This lets the key be provisioned by the mount proxy itself at pod init time instead of requiring an external init step.

## Implementation notes

- Flag plumbed through `server.SetInitNASRSAPEM()` / `server.InitNASRSAPEM()` (`atomic.Bool`), following the existing `--cleanup-nas-mounts-on-exit` pattern.
- Generation runs in `alinas.Driver.Init()` after `setupDefaultConfigs()`; a failure panics, consistent with the existing `setupDefaultConfigs()` error handling.
- The key is regenerated on every start when the flag is on. The existing file is removed first, because a previous run leaves it at mode `0400`, which would otherwise make `openssl` fail to write it.
- Added `Driver.ConfigDir` to override `/etc/aliyun` in tests.

## Behavior when disabled

Default `false` → no change to existing behavior.

## Tests

- `TestInitRSAPrivateKey`: key file created, permissions are `0400`, content is a PRIVATE KEY PEM, and a second run successfully regenerates it over the read-only file.
- `TestConfigDir`, `TestInitNASRSAPEMFlag`.

### DIFF
--- a/build/mount-proxy/Dockerfile
+++ b/build/mount-proxy/Dockerfile
@@ -107,7 +107,7 @@ ENTRYPOINT ["csi-mount-proxy-server", "--driver=ossfs2"]
 FROM alinux-install AS alinas
 
 ARG TARGETPLATFORM
-ARG ALINAS_UTILS_VERSION=2.4-0.20260813143608.fcc94d.generic
+ARG ALINAS_UTILS_VERSION=2.4-0.20260825141903.98d7f7.generic
 ARG EFC_VERSION=2.4-20260817135527.cb17f4.5b35e56.release
 ARG ALINAS_RPM_BASE_URL=https://aliyun-alinas-eac.oss-cn-beijing.aliyuncs.com/efc-release/2.4
 
@@ -119,7 +119,7 @@ RUN --mount=type=bind,from=rpm-builder,source=/root/rpmbuild/RPMS/noarch,target=
         linux/arm64) ARCH="aarch64" ;; \
         *) echo "unsupported platform"; exit 1 ;; \
     esac; \
-    yum install -y /tmp/rpms/python3-provider-*.rpm util-linux procps tini; \
+    yum install -y /tmp/rpms/python3-provider-*.rpm util-linux procps tini openssl; \
     yum install -y ${ALINAS_RPM_BASE_URL}/aliyun-alinas-utils-${ALINAS_UTILS_VERSION}.${ARCH}.rpm; \
     yum install -y ${ALINAS_RPM_BASE_URL}/alinas-efc-${EFC_VERSION}.${ARCH}.rpm; \
     yum clean all; \
@@ -129,6 +129,9 @@ RUN --mount=type=bind,from=rpm-builder,source=/root/rpmbuild/RPMS/noarch,target=
 # alinas-utils generates some dynamic configuration files (such as TLS keys, etc.) under /etc/aliyun/{alinas,cpfs}.
 # So we will mount a hostPath to prevent the loss of these files when the container restarts.
 # On the first launch, /etc/aliyun-defaults will be used to initialize the configurations under /etc/aliyun/{alinas,cpfs}.
+# The NAS RSA private key is generated at pod startup by csi-mount-proxy-server
+# via the alinas-keygen script (shipped by aliyun-alinas-utils), so it is not
+# pre-generated here.
 RUN cp -r /etc/aliyun /etc/aliyun-defaults
 
 COPY --link --from=builder /out/csi-mount-proxy* /usr/local/bin/
@@ -146,7 +149,7 @@ ARG OSSFS_VERSION=v1.91.12.ack.2
 ARG OSSFS2_VERSION=v2.0.9.ack.3
 ARG OSSFS_RPM_URL
 ARG OSSFS2_RPM_URL
-ARG ALINAS_UTILS_VERSION=2.4-0.20260813143608.fcc94d.generic
+ARG ALINAS_UTILS_VERSION=2.4-0.20260825141903.98d7f7.generic
 ARG EFC_VERSION=2.4-20260817135527.cb17f4.5b35e56.release
 ARG ALINAS_RPM_BASE_URL=https://aliyun-alinas-eac.oss-cn-beijing.aliyuncs.com/efc-release/2.4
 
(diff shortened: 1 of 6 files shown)


pkg/mounter/proxy/server/alinas/driver.go
```diff
@@ -93,10 +96,71 @@ func (h *Driver) Mount(ctx context.Context, req *proxy.MountRequest) error {
...
+		klog.InfoS("NAS RSA private key already exists, skipping generation", "path", keyPath)
+		return nil
+	} else if !os.IsNotExist(err) {
+		return fmt.Errorf("stat %s: %w", keyPath, err)
+	}
+
+	klog.InfoS("Generating NAS RSA private key via alinas-keygen", "command", alinasKeygenCommand, "path", keyPath)
+
+	cmd := exec.Command(alinasKeygenCommand)
+	cmd.Env = os.Environ()
```
huww98: Is it necessary? I think this is the default?

pkg/mounter/proxy/server/handler.go
```diff
@@ -146,6 +146,21 @@ func CleanupNASMountsOnExit() bool {
...
+// SetInitNASRSAPEM controls whether the alinas driver should generate the NAS
+// RSA private key (privateKey.pem) during initialization.
+func SetInitNASRSAPEM(v bool) {
+	initNASRSAPEM = v
+}
+
+// InitNASRSAPEM reports whether NAS RSA private key initialization is enabled.
+func InitNASRSAPEM() bool {
+	return initNASRSAPEM
+}
```
huww98: If we don't need synchronization for this, just make this variable public? So that we don't need seperate `initNASRSAPEM` in main.go

build/mount-proxy/Dockerfile
```diff
@@ -158,7 +161,7 @@ RUN --mount=type=bind,from=rpm-builder,source=/root/rpmbuild/RPMS/noarch,target=
         linux/arm64) OSSFS_ARCH="aarch_64"; ALINAS_ARCH="aarch64" ;; \
         *) echo "unknown platform"; exit 1 ;; \
     esac; \
-    yum install -y /tmp/rpms/python3-provider-*.rpm fuse-devel util-linux mailcap procps tini; \
+    yum install -y /tmp/rpms/python3-provider-*.rpm fuse-devel util-linux mailcap procps tini openssl; \
```
huww98, line 164: alinas-utils should declare dependency on `openssl`, so this can be installed automatically?

build/mount-proxy/Dockerfile
```diff
@@ -146,7 +149,7 @@ ARG OSSFS_VERSION=v1.91.12.ack.2
 ARG OSSFS2_VERSION=v2.0.9.ack.3
 ARG OSSFS_RPM_URL
 ARG OSSFS2_RPM_URL
-ARG ALINAS_UTILS_VERSION=2.4-0.20260813143608.fcc94d.generic
+ARG ALINAS_UTILS_VERSION=2.4-0.20260902175145.d1c35c.generic
```
huww98, line 152: Update from `VERSIONS`?

pkg/mounter/proxy/server/alinas/driver.go
```diff
@@ -93,10 +96,71 @@ func (h *Driver) Mount(ctx context.Context, req *proxy.MountRequest) error {
...
+// The key it emits is an unencrypted PKCS#8 PEM block
+// ("-----BEGIN PRIVATE KEY-----").
+//
+// This function is idempotent regardless of the script's behavior: it stats the
+// key first and skips generation when it already exists, so an existing key
+// (persisted via the /etc/aliyun hostPath) is never rotated on restart.
+func (h *Driver) initRSAPrivateKey() error {
+	keyPath := h.rsaPrivateKeyPath()
+	if _, err := os.Stat(keyPath); err == nil {
+		klog.InfoS("NAS RSA private key already exists, skipping generation", "path", keyPath)
```
huww98, line 146: OK for now. But will we need to rotate it in the future?

pkg/mounter/proxy/server/alinas/driver.go
mowangdk: Good catch, removed. `exec.Command` already inherits the current environment when `Env` is nil, so the explicit `cmd.Env = os.Environ()` was redundant. Done in commit 8c0a4dcc4.

pkg/mounter/proxy/server/handler.go
mowangdk: Done. Replaced the getter/setter with an exported `var InitNASRSAPEM bool`, so `main.go` binds the flag straight to it (`flag.BoolVar(&server.InitNASRSAPEM, ...)`) and the duplicate local var is gone. It's set once before `Init` and read only during single-threaded init, so no synchronization is needed. Commit 8c0a4dcc4.

build/mount-proxy/Dockerfile
mowangdk: Agreed that would be cleaner, but the dependency declaration lives in the aliyun-alinas-utils RPM packaging, which is out of this repo's control. Explicitly installing `openssl` here is the safe workaround for now; I'll relay the request to declare it as an RPM dependency to the alinas-utils owners.

build/mount-proxy/Dockerfile
mowangdk: Fixed. I updated `VERSIONS` (the single source of truth) to `2.4-0.20260902175145.d1c35c.generic` and regenerated the Dockerfile via `hack/update-versions.sh`, so the two are now in sync. Commit 29ab74aac.

pkg/mounter/proxy/server/alinas/driver.go
mowangdk: Not in this PR. The key is intentionally persisted (via the /etc/aliyun hostPath) and never rotated on restart, so per-node identity stays stable. If rotation is needed later, it should be a separate, explicit policy (e.g. an age-based or on-demand `--rotate-nas-rsa-pem` flag) coordinated with re-registering the key server-side and re-establishing mounts — deliberately out of scope here.

mowangdk: Yes, if a future AliNAS version changes its command parameters, we'll need a version rotation strategy.

PR #1842, opened 2026-09-04
feat(nas): bind-mount subpath on fallback to avoid a second NFS mount

mowangdk:
## What this PR does / why we need it

The NAS subpath `NodePublishVolume` fallback path is only hit when the subpath directory does **not yet exist** on the NFS server (i.e. the controller-side `nas:CreateDir` was skipped, or a non-standard filesystem creates the subdir lazily on the node).

Today that fallback creates the directory under a temporary root mount, unmounts the root, then does a **second, independent NFS mount** of `server:/<subpath>` onto the target — an extra NFS round-trip on the publish critical path.

This PR bind-mounts the freshly created subpath directory onto the target before unmounting the temporary root:

```
MkdirTemp -> rw-mount root -> mkdir -> mount --bind subdir -> target -> unmount root
```

A bind mount shares the underlying NFS superblock with the temporary root, so the target stays usable after the root is unmounted; the underlying NFS is released when the target is finally unmounted. `NodeUnpublishVolume` is unchanged — the target is still a single mount point.

## Scope / no regression (no feature gate)

Restricted to the plain `nfs` fstype, where a local bind is a true equivalent of the second NFS mount:

- **The primary path is unchanged** (subpath already exists -> direct `server:/<subpath>` mount).
- **Plain NFS targets are always unmounted locally**, so the bind introduces no unmount-side change.
- **`alinas`/`cpfs`/`cpfs-nfs` (incl. AccessPoint and EFC) are excluded**: those targets are mounted through the mount broker, which records per-target ownership for `NodeUnpublishVolume`. A local bind is not broker-owned, so its unmount would fall back to the slow local `umount` path dropped by the `csi_mount_proxy` nftables rule (~3s). They keep the second-NFS-mount path.

## Does this PR introduce a user-facing change?

```release-note
NONE
```

## Testing

- `go test ./pkg/nas/...` — all pass, including existing `TestDoMount*` (no regression)
- New tests exercise the plain-NFS fallback bind path (`TestDoMount_PlainNFSFallbackUsesBind`, `TestDoMount_PlainNFSFallbackReadOnlyRemounts`)

### DIFF
--- a/pkg/nas/bindmount_test.go
+++ b/pkg/nas/bindmount_test.go
@@ -1,0 +1,65 @@
+//go:build !windows
+
+package nas
+
+import (
+	"context"
+	"errors"
+	"testing"
+
+	"github.com/kubernetes-sigs/alibaba-cloud-csi-driver/pkg/mounter"
+	"github.com/stretchr/testify/assert"
+	mountutils "k8s.io/mount-utils"
+)
+
+// fallbackMounter fails the first ExtendedMount with an NFS "path not found"
+// error to trigger the subpath-creation fallback, succeeds later ExtendedMounts
+// (the temporary root mount), and records local Mount (bind) calls.
+type fallbackMounter struct {
+	mountutils.FakeMounter
+	extendedCalls int
+	bindOptions   [][]string
+}
+
+func (m *fallbackMounter) ExtendedMount(_ context.Context, _ *mounter.MountOperation) error {
+	m.extendedCalls++
+	if m.extendedCalls == 1 {
+		return errors.New("reason given by server: No such file or directory")
+	}
+	return nil
+}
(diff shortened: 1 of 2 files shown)


huww98:
I only read the description.

> Step 6 costs an extra NFS negotiation / network round-trip on the publish critical path.

If we are mounting the same server, maybe we can just swap step 5 and step 6? Kernel will reuse the superblock and it can be fast? This has the advantage of handling readonly mount transparently, and we don't need to branch in our code.
* If user request RO, we need to mount root RW, mkdir, then mount subdir RO, then kernel will detect this and create a new superblock (in some kernel version).
* If user request RW, kernel should reuse the superblock directly.

PR #1843, opened 2026-09-04
build: bring the blkid stages up to date

huww98:
#### What type of PR is this?

/kind cleanup

Brings the blkid-from-source build stages in `build/multi/Dockerfile.multi` up to date.

- Snapshot mirror: `snapshot-cloudflare.debian.org` instead of the canonical `snapshot.debian.org` host, which serves China at ~20 kB/s and times out an uncached build; the CDN alias serves the same immutable archives at ~1.5 MB/s.
- util-linux tarball from `mirrors.cernet.edu.cn` instead of kernel.org, which served the 9.5 MB file at ~16 kB/s (609 s of an 11-minute CI build). The `--checksum` on the ADD is what makes the source trustworthy, so the host does not have to be; the SBOM still names the canonical kernel.org URL.
- util-linux 2.41.2 → 2.41.5, for two fixes to code we aim at guest-written disks as root: a use-after-free in nested partition probing (partition array grown with `reallocarray()` while nested probers held pointers into it) and EBR bounds validation against a `uint32_t` overflow that let a crafted image register partitions at arbitrary sectors.
- Hardening, which was missing entirely: blkid was the only unhardened binary in an image whose mkfs and fsck all come hardened from Debian, and it is the one that reads attacker-controlled bytes. Flags now come from `dpkg-buildflags` (so a base-image bump brings the newer set along on its own) plus `bindnow`, which is what util-linux's own `debian/rules` asks for. Nine `_chk` entry points show up in the result.
- Target libc from multiarch (`libc6-dev:$DEB_HOST_ARCH`) instead of `libc6-dev-*-cross`: those cross packages are frozen at the bookworm release revision and install libraries outside any path libtool treats as a system one, so it hardcoded a `DT_RUNPATH` into the arm64 binaries. Multiarch is a no-op when the target is the builder's arch.
- `dpkg-architecture` for the triplet, retiring the hand-maintained arch table and the two per-arch stages (`build-util-linux-amd64`/`-arm64`) plus `build-0` that only existed to be their base. `ARG TARGETARCH` now marks the line between shared and per-target steps.
- `touch -h` so symlinks take the release tarball's mtime; without it `libblkid.so.1` kept the wall clock of whenever the stage last ran, the only thing that stopped the layer digest from being a function of the built content.
- `--no-install-recommends` in the deps stage, dropping ca-certificates, krb5-locales, libgpm2, libldap-common, libsasl2-modules and openssl from a stage we only harvest binaries out of.

Fixes #

Verified on both arches with a local buildkitd (`buildctl ... --opt target=build-util-linux`): the cross-compiled (arm64 host → amd64) and native (arm64) binaries are both stripped, carry 9 fortify `_chk` entry points, and have no `DT_RUNPATH`. The `debian` stage's `gather-node-deps.sh` completes cleanly with `--no-install-recommends` — none of the dropped packages provide binaries or libs that the gather step harvests.

```release-note
NONE
```

```docs
NONE
```

### DIFF
--- a/build/multi/Dockerfile.multi
+++ b/build/multi/Dockerfile.multi
@@ -32,9 +32,9 @@ RUN --mount=type=cache,target=/var/cache/apt,sharing=locked,id=apt-cache-$TARGET
     --mount=type=cache,target=/var/lib/apt,sharing=locked,id=apt-lib-$TARGETARCH \
     rm -f /etc/apt/apt.conf.d/docker-clean && \
     echo 'Acquire::Check-Valid-Until false;' > /etc/apt/apt.conf.d/snapshot && \
-    sed -i '/^URIs:/d; s|^# \(http://snapshot.debian.org/\)|URIs: \1|' /etc/apt/sources.list.d/debian.sources && \
+    sed -i '/^URIs:/d; s|^# http://snapshot.debian.org/|URIs: http://snapshot-cloudflare.debian.org/|' /etc/apt/sources.list.d/debian.sources && \
     apt-get update && \
-    apt-get install -y nfs-common e2fsprogs xfsprogs fdisk util-linux
+    apt-get install -y --no-install-recommends nfs-common e2fsprogs xfsprogs fdisk util-linux
 
 RUN --mount=type=bind,from=distroless-base,target=/base \
     --mount=type=bind,source=build/gather-node-deps.sh,target=/deps.sh \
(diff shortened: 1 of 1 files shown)


mowangdk (commented):
Thanks for the detailed writeup. The build changes are technically sound overall — checksum, hardening, multiarch libc, and the reproducibility fixes all check out. A couple of points below, one of which I think is worth fixing before merge.

### Important

**SBOM `releaseDate` looks inconsistent with 2.41.5's own content.**

The SBOM sets `"releaseDate": "2026-06-16T12:20:59Z"` for util-linux 2.41.5. But the 2.41.5 release notes are headlined by three 2026 `mount(8)` CVEs (CVE-2026-53612/53613/53614). A release dated 2026-06-16 predating those CVE assignments suggests the field is stale/carried over rather than taken from the actual 2.41.5 tag. Since the SBOM's whole purpose is provenance accuracy, could you set this to the real 2.41.5 release timestamp?

### Minor

**The "two fixes" framing in the description is a bit off.**

- The **use-after-free in nested partition probing** is real and is in the 2.41.5 changelog — good, and it's exactly the attacker-facing path this image exposes via `blkid`.
- The **"EBR bounds validation against a uint32_t overflow"** fix isn't listed in the 2.41.5 release notes. It may have landed in an earlier 2.41.x, so attributing it to the 2.41.2→2.41.5 delta as written isn't quite accurate.
- For what it's worth, 2.41.5 is dominated by the SUID `mount(8)` TOCTOU CVEs, which don't apply here (we ship `blkid`, not SUID `mount`), so the bump is still justified purely on the libblkid UAF fix. Just worth tightening the prose.

**util-linux mirror resolves through a redirect chain.**

`mirrors.cernet.edu.cn` 302-redirects to a downstream mirror (observed: `chinanet.mirrors.ustc.edu.cn`). `ADD` follows it and `--checksum` keeps it safe, so no correctness/security concern — but CI now depends on two mirrors being up rather than the single canonical host. Fine as a speed tradeoff; flagging in case download flakiness ever shows up there.

### Verified / LGTM on the rest

- New checksum `f586e35d…` matches kernel.org's PGP-signed sums for 2.41.5.
- Removed identifiers (`HOST`, `build-0`, `build-util-linux-{amd64,arm64}`) are unreferenced elsewhere in the tree.
- `--no-install-recommends` drops nothing harvested by `gather-node-deps.sh`, and `set -e` there guards against a needed lib going missing.
- `snapshot-cloudflare.debian.org` is the official CDN alias; APT signature verification is unaffected by the host swap.
- `touch -h` + `SOURCE_DATE_EPOCH` correctly make the layer digest a function of built content.

huww98:
> SBOM releaseDate looks inconsistent with 2.41.5's own content.

```
curl --head https://www.kernel.org/pub/linux/utils/util-linux/v2.41/util-linux-2.41.5.tar.xz
HTTP/2 200
server: nginx
content-type: application/x-xz
last-modified: Tue, 16 Jun 2026 12:20:59 GMT
```
The releaseDate is very accurate.

> The "EBR bounds validation against a uint32_t overflow" fix isn't listed in the 2.41.5 release notes

It is in 2.41.4

mowangdk (commented):
Retracting my "Important" point on the SBOM `releaseDate` — I was wrong.

The kernel.org listing for `util-linux-2.41.5.tar.xz` is dated `16-Jun-2026 12:20`, so the SBOM value `2026-06-16T12:20:59Z` is correct and matches the actual tarball publication timestamp. My inference that a `CVE-2026-536xx` ID implied a later release date was invalid — no change needed here.

The Minor notes (CVE-framing wording, and the cernet mirror redirect chain being a two-mirror availability dependency) still stand as-is, and everything else verified clean.

mowangdk (commented):
One more retraction — my Minor note questioning the "EBR bounds / uint32_t overflow" fix was also wrong.

I'd only checked the 2.41.5 release notes. Since this PR bumps 2.41.2 → 2.41.5, the intermediate releases are part of the delta, and the fix is there: **v2.41.4** ChangeLog, `libblkid/src/partitions/dos.c` — "validate EBR data and links within extended partition", explicitly describing a crafted image registering "partitions at arbitrary sectors via uint32_t overflow." That matches the PR description almost verbatim, and it's in libblkid (the attacker-facing path here), so it's correctly attributed.

So both fixes cited in the description are legitimate and land in the 2.41.2→2.41.5 delta:
- libblkid nested-partition use-after-free — v2.41.5
- libblkid EBR uint32_t-overflow bounds validation — v2.41.4

The only remaining Minor note that stands: the util-linux tarball comes via `mirrors.cernet.edu.cn`, which 302-redirects to a downstream mirror (`chinanet.mirrors.ustc.edu.cn`). I verified the served bytes match the pinned checksum, so it's safe — just a two-mirror availability dependency for uncached CI builds. Not blocking. LGTM on the rest.

PR #1845, opened 2026-09-10
feat(ossfs2): support FUSE fd-passing and Crash Recovery

AlbeeSo:
#### What type of PR is this?
/kind feature

# ossfs2 FUSE fd-passing and Crash Recovery

Rebased version of #1769 onto current master, with conflict resolution, bug fixes, and comprehensive test coverage.

## 1. Overview

In RunC mode, the ossfs2 daemon runs inside a shared fuse pod. When the daemon crashes (OOM, panic, SIGKILL), the FUSE mount becomes stale and all pods using that volume suffer IO failures until a manual `NodeUnstageVolume` cleanup.

This PR introduces two complementary features:

1. **FUSE fd-passing**: The CSI client performs the kernel mount and passes the FUSE file descriptor to the fuse pod via SCM_RIGHTS, decoupling kernel mount ownership from daemon lifecycle.
2. **Automatic recovery**: On non-SIGTERM exit, the supervisor flushes the FUSE connection (interrupting in-flight requests with `-EINTR`), then restarts ossfs2 with exponential backoff (up to 5 attempts).

## 2. Feature Gates

| Feature Gate | Default | Stage | Description |
|---|---|---|---|
| `EnableFUSEFdPassing` | false | Alpha | Client performs kernel mount; fd passed to server |
| `EnableOssfs2Recovery` | false | Alpha | Auto-restart on crash (implies fd-passing) |

Both gates must be configured **identically** on csi-provisioner (controller) and csi-plugin (node). Mismatched configurations lead to undefined behavior.

## 3. Architecture and Key Components

```
CSI Client ──SCM_RIGHTS(fd)──► Fuse Pod (Server)
  1. open /dev/fuse              1. recvMsgWithFd parses OOB
  2. mount -t fuse (kernel)      2. dup fd to prevent GC finalizer close
  3. WriteMsgUnix(msg, oob)      3. start ossfs2 (ExtraFiles)
  4. wait for response           4. wait for readiness (stat/poll)
  5. SafeCleanup on failure      5. superviseProcess monitors lifecycle
```

| Package / File | Responsibility |
|---|---|
| `pkg/features/features.go` | Feature gate definitions |
| `pkg/utils/os/kernel.go` | Kernel/OS prerequisite checks (alinux3, >=5.10.134-18, x86_64) |
| `proxy/client/client.go` + `fuse_options.go` | Client: open /dev/fuse, kernel mount, SCM_RIGHTS send, FUSE/daemon option separation |
| `proxy/server/handler.go` | Server: recvMsgWithFd, fd lifecycle management |
| `ossfs2/cmd.go` | Command building and process startup (dup fd -> ExtraFiles) |
| `ossfs2/mount.go` | Unified mount entry, fd-passing/legacy mode dispatch |
| `ossfs2/readiness.go` | Readiness detection (fd-passing: blocking stat; legacy: polling) |
| `ossfs2/supervise.go` | Process monitoring + recovery loop (exponential backoff 1s->2s->4s->8s->16s, max 5 attempts) |
| `mounter/utils/fuse_cleanup.go` | Safe unmount (mountinfo + umount2 syscall, avoids D-state) |
| `mounter/utils/fuse_conn.go` | FUSE connection flush (write to /sys/fs/fuse/connections/\<id\>/flush) |
| `interceptors/ossfs_monitor.go` | Recovery lifecycle metrics callback registration |
| `oss/nodeserver.go` | Node-side wiring: kernel probe, option parsing, RunC-only enablement |

## 4. Recovery Flow

1. ossfs2 exits abnormally (not SIGTERM)
2. `superviseProcess` detects exit via `<-proc.exited`
3. `OnProcessExit` callback -> metrics: failover_count++, state=unhealthy
4. `FlushFuseConnection(chanId)` -> write flush file, interrupt in-flight FUSE requests
5. `recoveryRestart` with exponential backoff (max 5 attempts, 10s timeout each)
6. On success -> `OnRecoverySuccess` -> state=healthy, track new PID
7. On exhaustion -> `OnRecoveryFailed` -> state remains unhealthy, stop monitoring

## 5. Scope Limitations

- **RunC only**: fd-passing and recovery are not enabled for RunD/MicroVM (each pod has its own daemon, blast radius is small)
- **ossfs (libfuse2)**: Does not support fd-passing
- **x86_64 only**: Recovery kernel patch is alinux3 x86_64 specific
- **Data integrity**: In-flight write buffer data is lost during recovery. Applications requiring write durability must call `fsync()` to confirm writes.

## 6. Changes since #1769

- Rebased onto current master (168 commits ahead), resolved 8 merge conflicts
- **Bug fix**: overlay interceptor D-state hang when fd-passing is enabled (statfs on dead FUSE daemon blocks)
- **Test coverage added** (not in original PR):
  - UT: liveness probe dispatch, failover image gate selection, supervise edge cases (max attempts exhaustion, multi-crash cycles, fd closed during recovery, terminate race windows, backoff sleep interruption)
  - Integration: concurrent mount during recovery, terminate during mount
  - Container-level: FUSE stub binary + 3 podman tests (mount-kill, fd-passing, fd-leak)
- Removed premature deprecation warning for ossfs mountOptions (actively used by customers for debugging)

## 7. Backward Compatibility

Verified that all changes are gated behind `EnableFUSEFdPassing` and `EnableOssfs2Recovery` feature gates. When both gates are OFF (default), all code paths are functionally identical to the pre-PR behavior:
- `SafeIsNotMountPoint(m, path, false)` delegates directly to `IsNotMountPoint`
- Legacy mount readiness polling is unchanged
- Pod spec (privileged, MountPropagation) unchanged
- Image selection unchanged

Supersedes #1769

This is a rebased and enhanced version of #1769. The original PR had merge conflicts with master. This PR resolves all conflicts and adds comprehensive test coverage across UT, integration, and container layers.

```release-note
Add alpha support for ossfs2 FUSE fd-passing and automatic crash recovery behind feature gates EnableFUSEFdPassing and EnableOssfs2Recovery (both default off).
```

```docs

```


pkg/mounter/fuse_pod_manager/oss/ossfs2/manager.go
```diff
@@ -264,25 +264,76 @@ func (f *fuseOssfs) buildPodSpec(c *fpm.FusePodContext, target string) (spec cor
...
+	// If FdPassing is enabled, the fuse pod no longer needs privileged mode because
+	// the kernel mount is performed by the CSI client (csi-plugin or csi-agent) and
+	// the FUSE fd is passed to this pod via SCM_RIGHTS.
+	//
+	// Caveat: if EnableFUSEFdPassing is enabled on the node (csi-plugin) but NOT on
+	// the controller (csi-provisioner), the controller will NOT set FdPassing=true
+	// in the mount request. The pod will therefore be created with privileged=true,
+	// even though the feature gate is enabled on the node. Both components must have
+	// the feature gate enabled for non-privileged mode to take effect.
+	privileged := !c.FdPassing
```
mowangdk, line 300: Since the pod is privileged here, this feels like a good spot to also pin the OOM kill order. `csi-mount-proxy-server` is PID 1 and the ossfs2 daemon is its child in the same container/cgroup, so the fd-passing recovery only fires in the "daemon dies, server survives" case. Under a shared cgroup memory limit though, nothing currently steers the kernel OOM killer toward the daemon — I don't see any `oom_score_adj` wiring in the pod spec (only the 50Mi memory request) — so a cgroup OOM can take out the server alongside/before the daemon, which aborts the FUSE connection and defeats recovery entirely (it falls back to `NodeUnstageVolume` cleanup).

Since the fuse pod is privileged, the server can just write its own `/proc/self/oom_score_adj = -1000` at startup (and/or set a higher score on the ossfs2 child), so the OOM killer preferentially reaps the daemon and keeps the fd-holder alive to recover. That directly protects the recovery invariant this PR is built on.

WDYT — worth wiring in around here? And if OOM recovery is only meant to cover the sub-cgroup-OOM case where the killer happens to pick the daemon, it'd be worth calling that out in Scope Limitations, since the current list reads as if daemon OOM is fully covered.

pkg/mounter/fuse_pod_manager/oss/utils.go
```diff
@@ -42,7 +43,12 @@ func SetDefaultImage(fuseType string, m metadata.MetadataProvider, config *fpm.F
 				config.ImageTag = defaultOssfsImageTag
 			}
 		case mounterutils.OssFs2Type:
-			config.ImageTag = defaultOssfs2ImageTag
+			if features.FunctionalMutableFeatureGate.Enabled(features.EnableOssfs2Recovery) ||
```
mowangdk, line 46: Are there any concerns about using a recovery image by default?

mowangdk: BTW, crash recovery should be controlled by us, not the end user, so I propose removing the recovery field from the PV.

pkg/mounter/proxy/client/fuse_options.go
```diff
@@ -0,0 +1,129 @@
...
+
+	daemonOptions = daemonOptionSet.UnsortedList()
+	return
+}
+
+// parseOptionKey extracts the key from an option string ("key" or "key=value").
+// Returns ("", false) for empty / blank-key inputs and "rw" (the default mode,
+// silently ignored to avoid conflicts with "ro").
+func parseOptionKey(opt string) (string, bool) {
+	kv := strings.SplitN(opt, "=", 2)
```
mowangdk: Small reuse/robustness note here: `parseOptionKey` re-implements key extraction with `SplitN(opt, "=", 2)` + a bare `kv[0] == ""` check, but the repo already has the same "split an option string at the first `=`" convention in `pkg/mounter/utils` — `IndexMountOptions`/`MergeMountOptions` both use `strings.Cut(o, "=")` followed by `strings.TrimSpace(k)`. Worth aligning to that, for two concrete reasons:

- **Whitespace keys leak through.** Because this doesn't `TrimSpace` the key, a blank entry like `" "` passes the `kv[0] != ""` check and gets inserted as a literal `" "` daemon option (later forwarded to ossfs2). The doc comment says blank entries are ignored, but only the exact empty string is. `IndexMountOptions` already handles this via `TrimSpace`.
- **`len(kv) == 0` is dead code.** `strings.SplitN` never returns a zero-length slice for a non-empty separator, so that branch can't be hit.

Minimal change that reuses the existing convention and fixes both:

```go
func parseOptionKey(opt string) (string, bool) {
    k, _, _ := strings.Cut(opt, "=")
    k = strings.TrimSpace(k)
    if k == "" || k == "rw" {
        return "", false
    }
    return k, true
}
```

`setFuseOption` can then reuse the returned key instead of doing its own second `SplitN`. The one-way split into fuse vs daemon options (with the `nullVal` flag semantics and the defaults∪mountFlags precedence) is genuinely specific to this path and worth keeping — just the key-parsing primitive is what's duplicated.

pkg/utils/os/kernel.go
```diff
@@ -0,0 +1,187 @@
...
+
+// KernelVersion represents a parsed kernel version string from uname.
+// Example: "5.10.134-18.al8.x86_64" → {Major:5, Minor:10, Patch:134, Sublevel:18, OSDist:"al8", Arch:"x86_64"}
+type KernelVersion struct {
+	Major    int
+	Minor    int
+	Patch    int
+	Sublevel int    // the numeric part after the first hyphen (e.g. 18 in "5.10.134-18")
+	OSDist   string // the OS distribution tag (e.g. "al8", "el8")
+	Arch     string // the architecture (e.g. "x86_64", "aarch64")
```
mowangdk: `Arch` is parsed but never read — it looks like dead state. It's assigned in `ParseKernelVersion` (`kv.Arch = strings.Join(suffixParts[i+1:], ".")`), but the actual architecture gate in `checkKernelForRecoveryWithInputs` uses the separate `machine` value from `uname -m` (`if machine != "x86_64"`), not `kv.Arch`. So nothing consumes this field.

Two small downsides of keeping it: it invites a future reader to wire the arch check to `kv.Arch` (which is parsed from the release string, not `uname -m`, and can legitimately differ), and it's parsed via the multi-segment tolerance path that has no real input exercising it.

Suggest dropping the `Arch` field (and the `kv.Arch = ...` assignment) and keeping `machine` from `uname -m` as the single source of truth for the arch check. If you'd rather keep arch on the struct for symmetry/logging, a short comment noting it's informational-only (the gate uses `uname -m`) would avoid the "wire it up" trap.

pkg/mounter/proxy/server/ossfs2/readiness.go
```diff
@@ -0,0 +1,101 @@
...
+		notMnt, err := m.IsLikelyNotMountPoint(target)
+		if err != nil {
+			return fmt.Errorf("mount point verification failed: %w", err)
+		}
+		if notMnt {
+			return fmt.Errorf("target %s is not a mount point after stat succeeded", target)
+		}
+		logger.Info("Successfully mounted", "mountpoint", target, "fd", fuseFd)
+		return nil
+	case err := <-ossfsExited:
```
mowangdk, line 61: Minor DRY note — the "ossfs2 exited during init" handling is duplicated across the two readiness paths and the wording diverges. Here (fd-passing) it returns `"ossfs2 exited during initialization: %w"` / `"ossfs2 exited during initialization"`, while `checkMountReadiness` (legacy path) returns `"ossfs2 exited: %w"` / `"ossfs2 exited"`. Same event, two different messages, three call sites total.

Worth extracting a tiny helper so the message is defined once, e.g.:

```go
func ossfs2ExitError(err error) error {
    if err != nil {
        return fmt.Errorf("ossfs2 exited during initialization: %w", err)
    }
    return fmt.Errorf("ossfs2 exited during initialization")
}
```

and use it in both the fd-passing `select` case and `checkMountReadiness`.

To be clear, this is only about the exit-handling snippet — splitting `waitForFdPassingMountReady` vs `waitForLegacyMountReady` into two functions is the right call and I would not merge them: they are genuinely different readiness models (fd-passing uses a single blocking `stat` via goroutine+`select` to catch the FUSE_INIT edge, since CSI already created the kernel mount, and it carries the intentional bounded stat-goroutine leak; legacy uses `PollUntilContextCancel` because ossfs2 does its own `mount(2)` and the mountpoint appears gradually). Collapsing them would put two unrelated control-flow styles (`select` vs poll) into one body — less readable, not simpler. Only the `ossfsExited` snippet is the duplication worth removing.

pkg/mounter/proxy/server/ossfs2/driver.go
```diff
@@ -65,16 +65,20 @@ func (h *Driver) Fstypes() []string {
...
-		Overlay:     req.Overlay,
+		Source:          req.Source,
+		Target:          req.Target,
+		FsType:          req.Fstype,
+		Options:         req.Options,
+		Secrets:         req.Secrets,
+		MetricsPath:     req.MetricsPath,
+		VolumeID:        req.VolumeID,
+		Overlay:         req.Overlay,
+		FuseFd:          fuseFd,
```
mowangdk, line 79: This `MountOperation` sets `Overlay`, `FuseFd`, `Recovery`, `HasActiveDaemon` — but not `FdPassing`. I think that leaves the overlay D-state fix unreachable in production.

The trace:

- The overlay interceptor's fast path is guarded by `op.FdPassing` (`pkg/mounter/interceptors/overlay.go`, `if op.FdPassing { op.Target = lowerDir; MkdirAll(lowerDir); return handler(ctx, op) }`). That branch exists specifically to skip the `IsNotLiveMountPoint`/statfs liveness probe on the lower FUSE mount — i.e. it's the "overlay interceptor D-state hang when fd-passing is enabled" fix called out in the PR description.
- `op.FdPassing` is only ever read at that one spot on the server side. Since this `MountOperation` never populates it, `op.FdPassing` is always the zero value `false` when the interceptor runs.
- So for an overlay + fd-passing ossfs2 mount whose daemon has crashed (fd still held, connection alive), the interceptor falls through to `raw.IsLikelyNotMountPoint(mergedDir)` → `IsNotLiveMountPoint(raw, lowerDir)`, which does a statfs on the lower FUSE mount and blocks in uninterruptible D-state — exactly the hang the branch was written to prevent.

Fix looks like adding the field right here, e.g. `FdPassing: fuseFd > 0,` (or `FdPassing: req.FdPassing,`), so the interceptor guard actually fires.

One caveat on reachability: this only bites when overlay and fd-passing are both in effect for the same ossfs2 mount, both alpha-gated. If that combination is intentionally impossible, then the `if op.FdPassing` branch (and its two unit tests) is effectively dead code and it'd be worth a comment saying so. Either way the current state — a shipped D-state fix that never triggers because the flag driving it is never set — seems worth reconciling.

pkg/mounter/fuse_pod_manager/oss/ossfs2/manager.go
AlbeeSo: Good catch. The fuse pod is already privileged, so writing `/proc/self/oom_score_adj` requires no additional capabilities. Will add this as a follow-up — mount-proxy-server sets -999 at startup (not -1000, so a genuine server leak can still be reaped), and the daemon inherits the default score. The kernel OOM killer then preferentially reaps the daemon (higher memory footprint + higher score), preserving the fd-holder for recovery.

Will also add a note to Scope Limitations clarifying that cgroup-level OOM that kills the server alongside the daemon defeats recovery.

pkg/mounter/fuse_pod_manager/oss/utils.go
AlbeeSo: No concerns — the failover image is built from ack/master + recovery code. The non-recovery code path is identical to the regular ossfs2 image. When feature gates are off, the failover image is not selected at all (line 45-47: only when EnableOssfs2Recovery or EnableFUSEFdPassing is enabled). Enabling the feature gate is the explicit intent to use this capability.

AlbeeSo: The PV-level `recovery` field is intentional — it gives per-volume granularity for gradual rollout and rollback. The feature gate controls the default (when enabled, all ossfs2 volumes get recovery). The PV field lets operators disable recovery on specific problematic volumes without turning off the gate cluster-wide. We control the default; the PV field is the escape hatch.

pkg/mounter/proxy/client/fuse_options.go
AlbeeSo: Fixed in 5072e3c. Switched to `strings.Cut` + `TrimSpace`, removed the dead `len(kv)==0` branch, and reused the returned key in `setFuseOption` to eliminate the duplicate `SplitN`.

pkg/utils/os/kernel.go
AlbeeSo: Fixed in 5072e3c. Removed the `Arch` field and its assignment. `uname -m` is the single source of truth for the architecture check — it is POSIX standard and returns the CPU architecture directly (`x86_64`/`aarch64`), whereas the kernel release string suffix format varies across distros and custom kernels, making it unreliable for gating.

pkg/mounter/proxy/server/ossfs2/readiness.go
AlbeeSo: Fixed in 5072e3c. Extracted `ossfs2ExitError(err)` helper used by both fd-passing and legacy readiness paths. Agreed that the two readiness functions themselves should stay separate — different control-flow models (blocking stat vs polling).

pkg/mounter/fuse_pod_manager/oss/ossfs2/manager.go
AlbeeSo: Correction on my previous reply: in fd-passing mode the fuse pod is non-privileged (`privileged = !c.FdPassing`, line 300), so writing `/proc/self/oom_score_adj` requires `CAP_SYS_RESOURCE` which is not available by default.

Options for follow-up:
1. Add `CAP_SYS_RESOURCE` to the fuse pod security context (minimal capability, well-scoped)
2. Use pod spec `resources.limits.memory` to influence OOM scoring without extra capabilities — the kernel OOM killer already factors in memory consumption relative to limits

Will investigate the right approach in a follow-up. For now, the recovery invariant (daemon dies, server survives) holds for SIGKILL/SIGSEGV crashes; cgroup OOM that kills the server is an acknowledged limitation that should be called out in Scope Limitations.

pkg/mounter/proxy/server/ossfs2/driver.go
AlbeeSo: Good catch — the overlay interceptor was using `op.FdPassing` (client-side intent flag) while all other server-side checks use `op.FuseFd > 0` (server-side fact: "I received an fd via SCM_RIGHTS"). Fixed in 5072e3c by changing overlay.go:43 from `op.FdPassing` to `op.FuseFd > 0`, consistent with mount.go, ossfs_secret.go, and readiness.go.

The design is intentional: `FdPassing` is a client-side concept (user opts into fd-passing mode). On the server side, the presence of a received fd (`FuseFd > 0`) is the authoritative signal. The driver does not need to propagate `req.FdPassing` into the MountOperation — the fd itself is the proof.

build/lib/init.sh
```diff
@@ -70,6 +70,11 @@ do
                 echo "Running oss plugin...."
                 mkdir -p "$KUBELET_ROOT_DIR/csi-plugins/ossplugin.csi.alibabacloud.com"
                 rm -rf "$KUBELET_ROOT_DIR/plugins/ossplugin.csi.alibabacloud.com/csi.sock"
+                # Pre-load fuse module so /sys/fs/fuse/connections exists before
+                # fuse pods start; otherwise the HostPathDirectory check races
+                # with the first kernel FUSE mount and emits FailedMount warnings.
+                host_cmd modprobe fuse
```
mowangdk, line 76: This is only necessary if the module hasn't been loaded yet. Rather than running modprobe every time, I think we should check if module is loaded when enabling the feature

pkg/mounter/fuse_pod_manager/oss/utils.go
mowangdk: > No concerns — the failover image is built from ack/master + recovery code. The non-recovery code path is identical to the regular ossfs2 image. When feature gates are off, the failover image is not selected at all (line 45-47: only when EnableOssfs2Recovery or EnableFUSEFdPassing is enabled). Enabling the feature gate is the explicit intent to use this capability.

Given that the non-recovery path is identical, could we just make the failover image the default for ossfs 2.0?

mowangdk: > The PV-level `recovery` field is intentional — it gives per-volume granularity for gradual rollout and rollback. The feature gate controls the default (when enabled, all ossfs2 volumes get recovery). The PV field lets operators disable recovery on specific problematic volumes without turning off the gate cluster-wide. We control the default; the PV field is the escape hatch.

Makes sense, but I'd prefer putting it in mountOptions so operators can easily modify it on the PV.

build/lib/init.sh
AlbeeSo: You are right that this runs on every csi-plugin restart. However, `modprobe` is internally idempotent — kmod checks `/proc/modules` before calling `init_module()`, so when the module is already loaded it is a no-op with zero kernel overhead. The only cost is one extra shell invocation.

Note that bmcpfs (line 89) also runs `modprobe fuse` unconditionally. If we add a guard here, we should be consistent and guard both. But given modprobe built-in idempotency, the lsmod pre-check would just be saving one shell call — not sure it is worth the added complexity.

Happy to add the guard if you prefer, but also fine leaving as-is since modprobe handles the already-loaded case internally.

pkg/mounter/fuse_pod_manager/oss/utils.go
AlbeeSo: Agree on the direction — making the failover image the default for ossfs2 is the right end state. However, the recovery support has some invasiveness into the ossfs2 mainline implementation. I would prefer to stabilize the current approach first, then unify the image once we have confidence from production usage.

Tracking as a future TODO.

AlbeeSo: Actually, `spec.mountOptions` (which arrives as `MountFlags` in CSI) does end up in the same `Options` list on the `MountOperation` — they get merged via `AddDefaultMountOptions` and passed to the proxy server. So putting `recovery=true` in mountOptions would technically reach the mount-proxy-server.

But the problem is timing: `recovery` controls **pod spec generation** on the controller side — image selection (failover vs regular), extra volume mounts (`/sys/fs/fuse/connections`), and privileged mode. This happens in `ControllerPublishVolume` / fuse pod creation, long before `NodePublishVolume` where mountOptions are processed. By the time mountOptions reach the node, the fuse pod is already created with or without recovery support.

So `recovery` must be in volumeAttributes (available to both controller and node), not mountOptions (node-only, mount-time only). The current design is correct for this reason.

AlbeeSo: Agree on the direction — making the failover image the default for ossfs2 is the right end state. However, the recovery support has some invasiveness into the ossfs2 mainline implementation. I would prefer to stabilize the current approach first, then unify the image once we have confidence from production usage.

Tracking as a future TODO.

AlbeeSo: `spec.mountOptions` (which arrives as `MountFlags` in CSI) does end up in the same `Options` list on the `MountOperation` — they get merged via `AddDefaultMountOptions` and passed to the proxy server. So putting `recovery=true` in mountOptions would technically reach the node.

But the problem is timing: `recovery` controls **pod spec generation** on the controller side — image selection (failover vs regular), extra volume mounts (`/sys/fs/fuse/connections`), and privileged mode. This happens in `ControllerPublishVolume` / fuse pod creation, long before `NodePublishVolume` where mountOptions are processed. By the time mountOptions reach the node, the fuse pod is already created with or without recovery support.

So `recovery` must be in volumeAttributes (available to both controller and node), not mountOptions (node-only, mount-time only). The current design is correct for this reason.

AlbeeSo: Agree on the direction — making the failover image the default for ossfs2 is the right end state. However, the recovery support has some invasiveness into the ossfs2 mainline implementation. I would prefer to stabilize the current approach first, then unify the image once we have confidence from production usage.

Tracking as a future TODO.

AlbeeSo: `spec.mountOptions` (which arrives as `MountFlags` in CSI) does end up in the same `Options` list on the `MountOperation` — they get merged via `AddDefaultMountOptions` and passed to the proxy server. So putting `recovery=true` in mountOptions would technically reach the node.

But the problem is timing: `recovery` controls **pod spec generation** on the controller side — image selection (failover vs regular), extra volume mounts (`/sys/fs/fuse/connections`), and privileged mode. This happens in `ControllerPublishVolume` / fuse pod creation, long before `NodePublishVolume` where mountOptions are processed. By the time mountOptions reach the node, the fuse pod is already created with or without recovery support.

So `recovery` must be in volumeAttributes (available to both controller and node), not mountOptions (node-only, mount-time only). The current design is correct for this reason.

build/lib/init.sh
AlbeeSo: Agreed. Replaced the kernel version check with a capability-based approach: `CheckKernelForRecovery` now scans `/proc/kallsyms` for the `fuse_flush_pq` symbol (the kernel function behind `/sys/fs/fuse/connections/<id>/flush`), similar to how bmcpfs checks `fuse_dev_ioctl_recover`. This directly detects whether the running kernel has the FUSE recovery flush support, regardless of version/OS/arch.

The `modprobe fuse` here remains as-is — it just ensures the fuse module is loaded before pod HostPathDirectory checks. The actual capability detection happens at nodeServer startup in Go code.

pkg/mounter/proxy/server/handler.go
```diff
@@ -93,7 +94,57 @@ type rawRequest struct {
...
+// Returns fuseFd=0 when no fd is received (backward compatible with old clients).
+func recvMsgWithFd(conn *net.UnixConn, req *rawRequest) (fuseFd int, err error) {
+	fuseFd = 0
+	// Buffer must be large enough for an entire request in one ReadMsgUnix call,
+	// because SCM_RIGHTS (fd) is only delivered with the first read.
+	const maxBufSize = 1 << 20 // 1MB — typical requests are <10KB; 1MB covers extreme cases
+	buf := make([]byte, maxBufSize)
+	oob := make([]byte, unix.CmsgSpace(4)) // space for one fd
+
+	n, oobn, _, _, err := conn.ReadMsgUnix(buf, oob)
```
mowangdk, line 107: Could we retain streaming request decoding here? This socket is `SOCK_STREAM`: a single client `Write`/`WriteMsgUnix` does not guarantee that one `ReadMsgUnix` receives the entire JSON message. Increasing the receive buffer to 1 MiB does not provide that guarantee either. The later `json.Unmarshal(buf[:n], req)` therefore rejects valid requests when the first read returns only a prefix.

Before this change, `Handle` called `proxy.ReadMsg`, whose `json.Decoder` continued reading until it had a complete JSON value. This is consequently a regression even with both feature gates disabled, not just an fd-passing issue.

I reproduced it in a Linux container with a valid ~128 KiB request sent in a **single Write**, after setting the client's send buffer to 4 KiB. The server returned `read request: unmarshal request: unexpected end of JSON input`.

Suggested minimal fix: keep the initial `ReadMsgUnix` and SCM_RIGHTS extraction, then replay the received bytes before continuing to read the socket:

```go
reader := io.MultiReader(bytes.NewReader(buf[:n]), conn)
if err := proxy.ReadMsg(reader, req); err != nil {
    if fuseFd > 0 {
        _ = unix.Close(fuseFd)
    }
    return -1, fmt.Errorf("read request message: %w", err)
}
return fuseFd, nil
```

This needs a `bytes` import and replaces the trailing-newline stripping/direct `json.Unmarshal` block. Also remove the `n == maxBufSize` rejection: filling the first buffer is not an oversized message; `proxy.ReadMsg` can enforce the existing total `proxy.MaxMsgSize` limit. The initial buffer can then be a modest chunk buffer rather than a whole-message buffer.

Please add regression tests for fragmented requests both with and without an fd, a single-write request larger than the first-read buffer, and fd cleanup on a truncated/timed-out request. `TestNoMessageEnd` should also restore the original expectation that a missing message delimiter is an error.

pkg/oss/nodeserver.go
```diff
@@ -285,11 +326,18 @@ func (ns *nodeServer) NodePublishVolume(ctx context.Context, req *csi.NodePublis
...
+		// 2. Recovery requires extra volumes (/sys/fs/fuse/) mounted in fuse pod
+		// 3. In RunC the FUSE daemon is shared across pods, making recovery more
+		//    critical; RunD/MicroVM has per-pod daemons with lower blast radius
+		// TODO: enable for RunD after stabilization in RunC
 		err := ossfsMounter.ExtendedMount(ctx, &mounter.MountOperation{
 			Source:      mountSource,
 			Target:      targetPath,
 			FsType:      opts.FuseType,
 			Options:     mountOptions,
+			Args:        mountFlags,
```
mowangdk, line 340: Could we avoid forwarding kernel mount flags through `Args` in the MicroVM path? This block serves both RunD and MicroVM, but their mounters interpret `Args` differently: `ProxyMounter` sends them as protocol `MountFlags`, whereas `OssCmdMounter.getArgs` appends them verbatim after `ossfs2 mount <target>`.

For example, an existing PV with `mountOptions: [nodev]` now produces:

```text
ossfs2 mount <target> nodev -c <passwd-file> --oss_endpoint=... ...
```

Before this PR, `makeMountOptions` explicitly ignored MountFlags for ossfs2, so the same PV did not inject this extra positional argument. This regression occurs with **both feature gates disabled**.

I verified the full NodePublish argument-building path against the real `v2.0.9.ack.3 / a1f29fc` ossfs2 binary, replacing only the systemd launcher with a script that executes the generated command. Bare `nodev`, `nosuid`, `ro`, `rw`, and `max_read=131072` each fail with `fuse: invalid argument '<value>'`; that binary's error path then aborts, and NodePublish returns Internal. The no-MountFlags control passes argument parsing and reaches filesystem initialization (the test container intentionally had no network, so this was not an OSS mount-success test).

Please preserve the legacy MicroVM behavior by passing these flags only through the proxy path, or introduce a separate `MountOperation.MountFlags` field instead of overloading `Args`. Do not remove `op.Args` from `OssCmdMounter.getArgs` entirely: the secret interceptor also uses it for the required `-c <passwd-file>` arguments. A gate-off MicroVM regression test with non-empty MountFlags should cover this.

pkg/oss/nodeserver.go
```diff
@@ -168,10 +197,21 @@ func (ns *nodeServer) NodePublishVolume(ctx context.Context, req *csi.NodePublis
...
+	// - Legacy (no fd-passing): a crashed daemon leaves a mount whose root inode stat is still
+	//   cached, making the mount look healthy. IsNotLiveMountPoint adds a statfs probe that
+	//   reaches the daemon and detects this, auto-unmounting unserviced mounts so they can be
+	//   repaired. This is safe because there is no open FUSE fd keeping the connection alive.
+	var notMntTarget bool
+	if opts.FdPassing {
+		notMntTarget, err = checkMountPointFdPassing(ns.rawMounter, targetPath)
+	} else {
+		notMntTarget, err = checkMountPointLegacy(ns.rawMounter, targetPath)
+	}
```
mowangdk: Should this dispatch use the *effective runtime-specific* fd-passing mode rather than `opts.FdPassing` alone? `parseOptions` sets this flag when the feature gate is enabled, including in RunD/MicroVM, but the mount operation below intentionally does not enable fd-passing/recovery for those runtimes. Consequently they still use a legacy daemon while losing the legacy `IsNotLiveMountPoint`/statfs probe.

I reproduced this with a real legacy FUSE mount and SIGKILL of its daemon. While the root attributes were still cached, the same RunD NodePublish request behaved as follows:

- Gate off: statfs detects the disconnected mount, the driver unmounts it and attempts a remount.
- Gate on in the RunD node-server: this branch selects the presence-only probe, logs `already mounted`, and returns success without attempting repair.

Thus a feature intended to affect RunC changes RunD's existing failure-repair behavior, despite not providing the new automatic recovery there. This is not a gate-off regression; it requires enabling the gate in the RunD node-server/csi-agent process.

Please derive the effective mode after runtime selection, e.g. `fdPassingEnabled := runtimeType == RuntimeTypeRunC && opts.FdPassing`, and use it consistently for probe selection and mounting. RunD/MicroVM should retain the legacy liveness check until they actually support fd-passing. Please also add runtime cases to the probe-dispatch tests; the current tests cover only the RunC setup.

mowangdk (commented):
Two compatibility issues in the existing MicroVM/RunD paths. The MicroVM issue also reproduces with both new feature gates disabled.

pkg/mounter/fuse_pod_manager/oss/ossfs/manager.go
```diff
@@ -366,7 +366,10 @@ const (
 	KeyListObjectsV2 = "listobjectsv2"
 )
 
-func (f *fuseOssfs) AddDefaultMountOptions(options []string) []string {
+func (f *fuseOssfs) AddDefaultMountOptions(options []string, mountFlags []string) []string {
+	// For ossfs, mountFlags (from PV spec mountOptions) are appended to daemon options.
+	options = append(options, mountFlags...)
```
mowangdk, line 371: Please preserve the pre-PR option ordering for ossfs 1.x. Moving the MountFlags append here reverses their order relative to the options generated by `MakeMountOptions`, including the `ro` flag generated for a read-only CSI request. This changes existing behavior even with **both new feature gates disabled**.

Before this PR, the order was:

```text
otherOpts -> PV MountFlags -> generated volume options -> defaults
```

Now `makeMountOptionsAndFlags` merges the generated options first, and this method appends MountFlags afterward:

```text
otherOpts -> generated volume options -> PV MountFlags -> defaults
```

Concrete reproduction: an ossfs PV has `mountOptions: [rw]`, while the CSI publish request specifies `Readonly=true`. I verified the resulting access-option sequence:

```text
before: rw,ro
after:  ro,rw
```

These are not equivalent. libfuse 2 processes the flags in order: `ro` sets `MS_RDONLY`, while a subsequent `rw` clears it ([mount.c, fuse-2.9.9](https://github.com/libfuse/libfuse/blob/fuse-2.9.9/lib/mount.c#L167-L197)). Thus the newly appended `rw` cancels the driver's generated read-only mount flag. Whether the container runtime additionally enforces a read-only bind is separate; the CSI-level read-only behavior must not regress. The option-order regression is tested, and the flag interpretation is confirmed from libfuse source; I have not performed an actual OSS write test.

Could we keep the original merge order for ossfs 1.x while introducing the separate kernel-flags path for ossfs2? Please add a gate-off regression test combining `Readonly=true` with `MountFlags: []string{"rw"}`. This should not require users to change an existing PV merely to retain the pre-upgrade behavior.

pkg/mounter/proxy/server/handler.go
AlbeeSo: Fixed. `recvMsgWithFd` now does a small initial `ReadMsgUnix` (64KB) to capture SCM_RIGHTS, then replays the received bytes via `io.MultiReader` into `proxy.ReadMsg` (json.Decoder) for streaming decode. The `n == maxBufSize` rejection is removed.

Added regression tests:
- `TestFragmentedRequest`: message split across two writes with delay
- `TestLargeRequest`: 128KB request exceeding the initial buffer
- `TestNoMessageEnd`: restored original expectation (missing delimiter = error)

pkg/oss/nodeserver.go
AlbeeSo: Fixed both issues:

**MicroVM mountFlags leak**: Removed `Args: mountFlags` from the MicroVM MountOperation. `OssCmdMounter` appends Args verbatim to the ossfs2 command line, so kernel flags like `nodev` caused `fuse: invalid argument`. MicroVM does not use fd-passing and has no use for kernel mount flags.

**RunD probe dispatch** (related comment on the same file): Derived `fdPassingEffective := runtimeType == RuntimeTypeRunC && opts.FdPassing` and use it for probe selection and `SafeIsNotMountPoint`. RunD/MicroVM now always use the legacy statfs liveness probe regardless of feature gate state. Added RunD test cases to the probe dispatch test.

The RunC path still passes `Args: mountFlags` to ProxyMounter, which maps them to `MountRequest.MountFlags` for the client-side kernel mount in fd-passing mode. When fd-passing is off, the client skips `prepareFuseFd` and MountFlags are not consumed. Server-side `MountOperation.Args` is populated only by interceptors (e.g. `-c <passwdFile>`), not from `MountRequest.MountFlags`.

pkg/mounter/fuse_pod_manager/oss/ossfs/manager.go
AlbeeSo: Fixed. Restored the pre-PR merge order for ossfs 1.x: `otherOpts → mountFlags → generated → defaults`.

mountFlags are merged into daemon options inside `makeMountOptionsAndFlags` for ossfs 1.x only (matching the original `makeMountOptions` behavior). ossfs2 and customfuse do not merge mountFlags into daemon options — they are returned separately for kernel mount use in fd-passing mode.

ossfs 1.x `AddDefaultMountOptions` no longer handles mountFlags (already merged upstream). The interface now delegates per-implementation semantics.

Added `TestMountOptionsPipeline_Regression` covering ossfs1/ossfs2 × mountFlags/no-mountFlags × readonly combinations, verifying that ossfs 1.x `rw` (from MountFlags) appears before generated `ro` (libfuse2 last-wins), and that ossfs2 daemon options never contain mountFlags.

pkg/mounter/proxy/server/ossfs2/supervise.go
```diff
@@ -0,0 +1,173 @@
...
+
+		ctx, cancel := context.WithTimeout(context.Background(), 10*time.Second)
+		logger.Info("Restarting ossfs2 for recovery", "attempt", attempt, "target", target)
+		proc, err := m.startAndWaitReady(ctx, op, true, nil)
+		cancel()
+
+		if err != nil {
+			lastErr = err
+			logger.Error(err, "Recovery restart failed", "attempt", attempt, "target", target)
+			continue
```
mowangdk, line 159: Could we flush the connection after a failed replacement process has fully exited and before retrying it? Currently `superviseProcess` flushes only once before calling `recoveryRestart`; a failed attempt here goes directly to the next startup without another flush.

A replacement daemon can also read FUSE requests and then crash before replying or passing readiness. Those requests are already in the kernel processing queue and will not be delivered again by ordinary reads from the next daemon. This is the same orphaned-request problem that the initial flush addresses.

I reproduced the sequence using a real FUSE connection: attempt 1 reads a STATFS request and exits without replying; attempt 2 starts a working daemon. The result was `launches=2, flushes=1`: fresh STATFS requests succeeded through the recovered daemon, but the request consumed by attempt 1 remained blocked until connection teardown. For this test, initial flush was a no-op hook with an initially empty processing queue, and first-attempt readiness was held pending to exercise a startup failure. This is not a full alinux recovery E2E test. The public cloud-kernel flush implementation also explicitly describes processing requests as requests already sent to userspace that need to be ended during failover.

Please include flushing in the per-attempt lifecycle, after the preceding process is known to have exited, rather than only before the entire retry batch. A regression test should cover a failed startup that has consumed an application request, followed by a successful restart, and verify that the old request is completed/failed rather than left hanging.

pkg/mounter/proxy/server/ossfs2/supervise.go
```diff
@@ -0,0 +1,173 @@
...
+			// with shutdown, Terminate()'s wg.Wait() blocks until this sleep finishes
+			// (worst case ~15s across all retries). Probability is very low and the delay
+			// is within typical pod terminationGracePeriodSeconds. Replace with a select
+			// on a terminate channel if faster shutdown is needed in the future.
+			time.Sleep(delay)
+		}
+
+		ctx, cancel := context.WithTimeout(context.Background(), 10*time.Second)
+		logger.Info("Restarting ossfs2 for recovery", "attempt", attempt, "target", target)
+		proc, err := m.startAndWaitReady(ctx, op, true, nil)
```
mowangdk, line 153: Please recheck termination after the backoff sleep, before starting another process. The check at the top of the loop has already passed when `time.Sleep` begins, so setting `terminating=true` during that sleep still allows this `startAndWaitReady` call to run.

I reproduced this by waiting until `recoveryRestart` was actually sleeping, then setting the terminating flag. The launch count increased from 1 to 2 after shutdown was requested. The existing `TestSuperviseProcess_TerminateDuringBackoffSleep` verifies eventual completion but does not assert that no further process is started.

The post-recovery terminating check does handle a successfully started process, so this is not necessarily a permanent shutdown hang. However, shutdown can include another startup/readiness wait (up to the 10-second attempt timeout plus cleanup), not merely the remaining sleep duration described in the comment. The attempt context is also derived from `context.Background()`, so shutdown cannot cancel an in-progress readiness wait.

At minimum, add a terminating check immediately after sleeping. Prefer a cancellation-aware backoff and a supervisor lifecycle context that `Terminate` cancels, and add an assertion that shutdown during backoff does not start another daemon.

pkg/mounter/proxy/server/ossfs2/supervise.go
```diff
@@ -0,0 +1,173 @@
...
+			// Other break paths rely on recovery callbacks for metrics instead.
+			trulyExited <- exitErr
+			break
+		}
+
+		var exitErrTyped *exec.ExitError
+		if errors.As(exitErr, &exitErrTyped) {
+			if status, ok := exitErrTyped.Sys().(syscall.WaitStatus); ok && status.Signal() == syscall.SIGTERM {
+				logger.Info("ossfs2 terminated by SIGTERM, not recovering", "pid", currentPid)
+				break
```
mowangdk, line 66: Non-blocking exit-policy clarification: should a SIGTERM sent directly to ossfs2 always suppress recovery, or only when the final wait status says the process was killed by SIGTERM? These differ when a daemon installs a signal handler and exits normally after cleanup.

I verified both cases through the current startup/supervision code: an unhandled SIGTERM skips recovery, while a process that traps SIGTERM and calls `exit(0)` enters recovery. `cmd.Wait()` returns nil in the latter case, and `mount.go` converts it to `ossfs2 exited with no error`, which cannot match this `*exec.ExitError`/SIGTERM check.

Normal mount-proxy shutdown is separately protected by `terminating=true`; this question concerns terminating only the child while the supervisor stays alive. If the intended policy is that any child exit is recoverable unless the supervisor explicitly stopped the mount, please document that rather than implying all SIGTERM exits are excluded. If direct SIGTERM is intended as a reliable stop mechanism, an explicit per-mount stop state/control path is more robust than inferring intent solely from WaitStatus.

mowangdk (commented):
Follow-up on supervisor retry/termination handling: two implementation issues and one non-blocking clarification of the exit policy.

pkg/mounter/proxy/server/ossfs2/supervise.go
AlbeeSo: Good catch. Verified in the ossfs2 source: the mount main flow calls libfuse3 `fuse_set_signal_handlers(session)`, which registers a SIGTERM handler that calls `fuse_session_exit()` and lets the process exit normally with code 0. So `cmd.Wait()` returns nil, not an `*exec.ExitError` with SIGTERM — the current check is indeed dead code for ossfs2.

However, the correct way to stop a FUSE daemon is to unmount the mountpoint (which triggers `FUSE_DESTROY`), not to SIGTERM the process directly. In the CSI flow, normal shutdown goes through `Terminate()` which sets `terminating=true` before any process cleanup, so recovery is already suppressed. Direct `kill <pid>` by an operator is an edge case that can be worked around by unmounting the mountpoint instead.

For a proper fix, ossfs2 should override the libfuse3 SIGTERM handler to exit with code 143 (128+SIGTERM) instead of 0, so CSI can detect it. ossfs2 already has `set_one_signal_handler` infrastructure for this. Will track as a follow-up in the ossfs2 repo.

For this PR: adding a comment documenting the current policy assumption (SIGTERM detection relies on the process being killed by the signal, not trapping it and exiting cleanly).

pkg/mounter/proxy/server/ossfs2/supervise.go
AlbeeSo: Fixed. Moved flush from `superviseProcess` (once before the retry batch) into `recoveryRestart` (before each attempt). A failed replacement daemon may read FUSE requests and crash without replying — those orphaned requests need to be flushed before the next attempt starts.

Flush failure now causes `continue` to the next attempt (with backoff) instead of immediately aborting recovery. This handles transient flush errors.

Added `TestRecoveryRestart_FlushesBeforeEachAttempt`: all 5 attempts fail at readiness, verifies flush is called 5 times (once per attempt). Updated `TestSuperviseProcess_FlushFailureAbortsRecovery`: persistent flush failures now exhaust all attempts instead of aborting on first failure.

pkg/mounter/proxy/server/ossfs2/supervise.go
AlbeeSo: Fixed. Added `terminating` check immediately after `time.Sleep(delay)` in `recoveryRestart`. Setting `terminating=true` during backoff now returns early without starting another daemon.

Added `TestRecoveryRestart_TerminateDuringBackoffPreventsNextAttempt`: first attempt fails, backoff starts (500ms), terminating is set during sleep, verifies `attemptCount == 1` (no second attempt). Also updated `TestSuperviseProcess_TerminateDuringBackoffSleep` to assert `attemptCount == 1`.

Re: cancellation-aware backoff — `time.Sleep` is still not interruptible by `Terminate()`, so `wg.Wait()` blocks until the current sleep finishes. The post-sleep check prevents the next attempt but does not shorten the current sleep. A `select` on a terminate channel would fix this, but the worst-case delay (~15s across all retries) is within typical `terminationGracePeriodSeconds`. Can follow up if faster shutdown is needed.

mowangdk:
It hasn't been fully reviewed yet, so we need a gradual rollout.
/lgtm
/approve

PR #1846, opened 2026-09-14
feat(nas): support dynamic provisioning for AgenticFS

mowangdk:
#### What type of PR is this?

/kind feature

Adds a `volumeAs: agenticfs` controller mode to the NAS driver. Each PVC provisions
one AgenticSpace plus one AccessPoint bound to it, on an AgenticFS filesystem named either
directly by the `fileSystemId` StorageClass parameter or, through `containerNetworkFileSystem`, by
a CNFS object (`spec.type: agenticfs`) that owns it. This is control plane only: the node side is
unchanged, mounting goes through the AccessPoint domain with `tls,ram`.

Design points worth reviewing:

- **Idempotence.** `CreateAgenticSpace` carries `ClientToken=<PV name>`, so a retry returns the
  same space. `CreateAccessPoint` has no ClientToken and no uniqueness constraint on its name, so
  it is made idempotent by discovery instead: every attempt first lists the accesspoints already
  bound to the space and reuses one. Without that step each retry orphans an accesspoint, and
  since `DeleteAgenticSpace` requires all of a space's accesspoints to be detached, the leftovers
  would block `DeleteVolume` forever.
- **Compensation.** A failure after the space was created deletes only the accesspoint this call
  created, and only on a terminal gRPC code - on a retryable code external-provisioner retries, and
  compensating would destroy an accesspoint the next attempt is about to reuse. The space is
  reported and left to an external reaper rather than deleted. Three stable, mutually
  non-substring log prefixes (`agenticfs-resource-created`, `agenticfs-resource-retained-for-retry`,
  `agenticfs-orphan-resource`) let a reaper count each leak exactly once.
- **Timeout budget.** The compensation is bounded so no request it issues is still in flight once
  `CreateVolume` returns, because the per-volume lock is released after it.
- **Filesystem source.** The driver never creates the filesystem, only AgenticSpaces and
  accesspoints on it, so naming one with `fileSystemId` is sufficient and no CNFS is required: a
  filesystem managed outside the cluster works directly. `containerNetworkFileSystem` stays
  supported and keeps its `spec.type` / `status.fsAttributes.storageType` validation, and
  `fileSystemId` wins when both are set so a StorageClass can be migrated without a gap. The
  direct path does not pre-check the filesystem type with `DescribeFileSystems`, leaving that to
  `CreateAgenticSpace`, which rejects a wrong ID precisely and without an extra round trip on
  every `CreateVolume`. Zone and network placement (`zoneId`, `vpcId`, `vSwitchId`) are
  StorageClass parameters on both paths.

Also fixes a unit misuse in `NasClientV2`: `connTimeout` is the SDK `ConnectTimeout`, which the
darabonba runtime applies in **milliseconds**, so the value `10` meant a 10ms bound rather than the
intended 10s. It survived in production only because a warm same-region VPC connection answers in
single-digit milliseconds. This affects every NAS v2 call path, not just agenticfs.

Fixes #

- The vendored `nas-20170626/v4` and `darabonba-openapi/v2` bumps are what expose
  `CreateAgenticSpace` / `CreateAccessPoint` / `ListAccesspoints(AgenticSpaceId=...)`.
- Test coverage leans on wire-level contract tests rather than hand-built error fixtures: they
  drive the real vendored SDK against a local `httptest` stand-in, so they pin the actual wire
  shape (including that one NAS call is bounded by `httpClient.Timeout`, and that `ctx` is not
  threaded into the SDK).
- The `connTimeout` value and its millisecond unit are pinned by a deliberately literal constant,
  so raising it turns the tripwire red instead of silently re-deriving every bound from it.
- Examples are under `examples/nas/agenticfs/`.

```release-note
Added `volumeAs: agenticfs` to the NAS driver: each PVC provisions one AgenticSpace plus one bound AccessPoint on an AgenticFS filesystem named by the `fileSystemId` StorageClass parameter or owned by a CNFS. Also fixed the NAS SDK v2 client timeout, which was applied in milliseconds instead of the intended seconds.
```

```docs
NONE
```


pkg/nas/agenticfs_controller.go
```diff
@@ -0,0 +1,1337 @@
...
+			if outcome.err != nil {
+				outcomeText = fmt.Sprintf("answered NotFound (%v), the accesspoint was already gone", outcome.err)
+			}
+			reportResolved(fmt.Sprintf("CreateVolume failed terminally with code %s and the compensating nas:DeleteAccesspoint %s; the accesspoint is reconciled, the agenticspace is still left to the reaper by design", code, outcomeText))
+		}
+	case <-cleanupCtx.Done():
+		// Neither can outlive this function, so the lock is never released while a delete of this volume is in flight.
+		reportLeak(fmt.Sprintf("CreateVolume failed terminally with code %s and the compensating nas:DeleteAccesspoint did not return within %s (%v); the delete may still be in flight OR may never have been sent at all (the goroutine can still be queued on the shared rate limiter when the bound fires, in which case limiter.Wait returns context.DeadlineExceeded and the wire request is never issued). Neither case can outlive this call: the wait phase is capped at %s and the wire phase at %s - reconcile against the accesspoint's actual cloud state before assuming either", code, c.compTimeout, cleanupCtx.Err(), waitBudget, nasAPICallBound), cleanupCtx.Err())
+	}
+}
```
AlbeeSo: **Concern: `compensateCreateVolume` complexity is very high (lines 648-780).**

The compensation goroutine with atomic latch (`reported`), budget gates, panic recovery, and multiple reporting exits (leak vs resolved vs diagnostic) is _correct_ but hard to reason about — 130 lines of interleaved concurrency, error classification, and observability in a single function.

Future maintainers will need to understand:
- The `reported` latch ensures exactly-once orphan/resolved reporting across the goroutine boundary
- The `compensationWindowOpen` budget gate interacts with `rpcBudget` vs `compTimeout` vs `nasAPICallBound`
- The `select` on `done` vs `cleanupCtx.Done()` can race with the goroutine's own `reportLeak`
- Three sentinel errors (`errCompensateDeletePanicked`, `errCompensateDeleteNeverSent`, `errCompensateDeleteNoOutcome`) each drive a different `select` branch

Suggestion: extract the goroutine body into a named method that returns `compensateOutcome`, and consider a small state machine (e.g. `enum { neverSent, sent, succeeded, failed, panicked }`) so each reporting exit maps to exactly one state. A diagram in the doc comment would also help.

pkg/nas/agenticfs_controller.go
```diff
@@ -0,0 +1,1337 @@
...
+	// True only once the call got past the rate limiter, i.e. a wire request may exist and be in flight.
+	sent bool
+	// Kept for diagnostics; the sentinel err is what the select branches on.
+	panicValue any
+}
+
+// pkg/nas/cloud's wait() wraps every rate-limiter failure before the request is built - the only signal.
+func deleteNeverSent(err error) bool {
+	return err != nil && strings.Contains(err.Error(), nasRateLimiterWaitPrefix)
+}
```
AlbeeSo: **Concern: `deleteNeverSent` relies on fragile string matching.**

```go
func deleteNeverSent(err error) bool {
	return err != nil && strings.Contains(err.Error(), nasRateLimiterWaitPrefix)
}
```

This distinguishes "the delete was never issued" from "it was issued and failed" by checking whether the error message contains the rate-limiter wait prefix from `pkg/nas/cloud`. If the `wait()` error format in `nas_client_v2.go` changes (e.g. someone wraps it differently, or the message is localized), this silently misclassifies "never sent" as "sent and failed", changing the orphan reporting semantics.

Suggestion: use a sentinel error or a typed wrapper instead of string matching:
```go
var ErrRateLimiterWait = errors.New("rate limiter wait")
// in wait():
return fmt.Errorf("%w: %w", ErrRateLimiterWait, err)
// in deleteNeverSent():
return errors.Is(err, ErrRateLimiterWait)
```
This makes the contract explicit and survives message rewording.

pkg/nas/agenticfs_controller.go
```diff
@@ -0,0 +1,1337 @@
...
+	}
+
+	// No tier prefix: the orphan or resolved line comes only once the outcome is known, so a reaper counts one.
+	logger.Info("compensating: about to delete the accesspoint this call created (the agenticspace is left to the reaper by design)",
+		c.compensationFields(filesystemId, agenticSpaceId, fileSystemPath, createdAccesspointId,
+			fmt.Sprintf("CreateVolume failed terminally with code %s; deleting the accesspoint this call created", code),
+			cause)...)
+
+	// Neither inherits the (usually expired) request deadline nor can be cancelled, and is bounded.
+	cleanupCtx, cancel := context.WithTimeout(context.WithoutCancel(ctx), c.compTimeout)
```
AlbeeSo: **Concern: no context propagation into the SDK means the compensation bound is only as reliable as `connTimeout`.**

The code comments acknowledge this: `"ctx never reaches the SDK"`. The `compensationTimeout` (15s) and `waitBudget` (5s) are the _only_ bounds on the compensation goroutine. If the SDK's internal retry logic kicks in (the darabonba runtime retries on certain HTTP errors), or if `connTimeout` somehow doesn't apply to a particular code path, the goroutine can outlive both bounds.

The `connTimeout=10000ms` fix helps a lot — but since it's the SDK `ConnectTimeout` that darabonba uses as the _whole HTTP call_ deadline, any SDK behavior change (e.g. switching to `ReadTimeout` separately, or adding automatic retries) could break this invariant silently.

The `compensationTimeout > nasAPICallBound` assertion (line 104-105) is good, but it only holds if the SDK honors the bound. Consider adding a defensive `context.WithTimeout` on `cleanupCtx` that is strictly shorter than `compensationTimeout`, so even if the SDK ignores `connTimeout`, the goroutine is force-cancelled.

pkg/nas/agenticfs_controller.go
```diff
@@ -0,0 +1,1337 @@
...
+		logger.Info(retainedForRetryLogPrefix+": CreateVolume failed with a retryable code, the accesspoint/agenticspace are kept for the next attempt - NOT an orphan, do not clean them up by hand",
+			c.compensationFields(filesystemId, agenticSpaceId, fileSystemPath, createdAccesspointId,
+				fmt.Sprintf("CreateVolume failed with retryable code %s; the agenticspace and the accesspoint are retained for the next attempt, which reuses them via the CreateAgenticSpace ClientToken replay and findReusableAccessPoint", code),
+				cause)...)
+		return
+	}
+
+	// Terminal: the provisioner will not retry, so delete exactly the accesspoint this call created.
+
+	if createdAccesspointId == "" {
```
AlbeeSo: nit: should this also check `agenticSpaceId == ""` the same way line 826 `spaceMayExist` does? Currently if we reach here with `createdAccesspointId == ""` but `agenticSpaceId != ""`, we report an orphan — which is correct. Just noting the two checks serve different purposes and the interaction is non-obvious.

AlbeeSo (commented):
Review comments on the AgenticFS-specific changes. The design is sound overall — idempotency via ClientToken + discovery, compensation that never deletes the space, and the stable log-prefix contract are all well thought out. Four concerns below.

AlbeeSo:
**Concern: No RAM policy documentation for the new AgenticFS API calls.**

`CreateAgenticSpace`, `DeleteAgenticSpace`, `GetAgenticSpace`, `SetAgenticSpaceQuota` are new NAS OpenAPI actions introduced by this PR, but I do not see corresponding updates to the RAM policy examples under `docs/ram-policies/`. Users deploying with least-privilege RAM policies will get `Forbidden` errors at runtime when provisioning AgenticFS volumes.

Suggestion: add the four new actions to the NAS controller RAM policy example (likely `docs/ram-policies/nas/` or similar), so users know exactly which permissions are required for AgenticFS provisioning.

PR #1858, opened 2026-09-21
feat(mounter): compile the FUSE recovery helpers on non-Linux

huww98:
#### What type of PR is this?

/kind cleanup

`SafeCleanupFuseMount`, `SafeIsNotMountPoint`, `GetFuseConnectionID` and `FlushFuseConnection` live in `//go:build linux` files, so every package that depends on them — `proxy/client`, `interceptors`, `oss`, `ossfs2` — builds and unit-tests only on Linux, even for the parts with no Linux dependency at all (option assembly, path math, secret handling). That makes it impossible to run those unit tests natively on a macOS dev machine.

This adds a `//go:build !linux` stub, mirroring the existing `pkg/disk/datacache/data_cache_unsupported.go` precedent, so those packages compile and their portable logic can be tested off Linux. The ordinary (non-`fuseUnsafe`) paths delegate to the portable `CleanupMountPoint` / `IsNotMountPoint`; the `fuseUnsafe` recovery paths and the sysfs connection control are Linux-only kernel mechanisms (direct `umount2`, mountinfo scans, `/sys/fs/fuse/connections`) and return a clear "only supported on Linux" error rather than silently succeeding.

The driver still only runs on Linux in production; this changes nothing there.

Verified on darwin that `pkg/mounter/utils`, `interceptors`, `proxy/client`, `oss` and `fuse_pod_manager/...` now build, and the `fuse_pod_manager/oss/ossfs` + `ossfs2` unit tests pass.

```release-note
NONE
```


AlbeeSo:
/lgtm
/;approve

PR #1860, opened 2026-09-22
feat(substrate): add substrate-mode support for CSI Node and Controller

AlbeeSo:
## What

Add substrate-mode support so the CSI driver works when called by Substrate's atelet instead of kubelet. Substrate bypasses kubelet and calls CSI Node plugins directly via Unix socket, which requires several adaptations.

**Depends on:** #1859 (first two commits are from that PR)

### Changes

| File | Change |
|---|---|
| `pkg/common/nodeserver.go` | Skip staging/target path validation against `/var/lib/kubelet` when `substrate-mode` is in VolumeContext |
| `pkg/common/nodeserver_test.go` | Unit tests for substrate path bypass |
| `pkg/nas/nodeserver.go` | Skip pod runtime lookup (`GetPodRunTime`) for substrate actors; inject STS token from node metadata service for AgenticFS mount (POC workaround) |
| `pkg/nas/utils.go` | `fetchNodeSTSToken()` — retrieve STS credentials from ECS metadata service |
| `pkg/nas/controllerserver.go` | Handle PV-not-found in `DeleteVolume` (substrate does not create K8s PersistentVolume objects) |

### Why substrate-mode is needed

1. **Path validation**: Substrate uses `/var/lib/ateom-gvisor/` for staging and mount targets, not `/var/lib/kubelet/`. The existing validator rejects these paths.

2. **Pod runtime lookup**: `GetPodRunTime` queries the K8s API for pod metadata. Substrate actors are not K8s pods, so this lookup fails.

3. **STS token injection**: AgenticFS mount (`alinas` protocol with `tls,vers=3,ram`) requires STS credentials. In standard K8s this comes from Agent Identity. In substrate-mode, atelet injects `csi.alibabacloud.com/substrate-mode=true` into VolumeContext, and the CSI plugin falls back to the node metadata service for STS tokens. This is a POC workaround until Agent Identity integration is complete.

4. **No PV object**: Substrate manages volumes through its own API (ateapi), not K8s PVCs/PVs. `DeleteVolume` must handle the case where the PV doesn't exist in the K8s API.

### Testing

**Unit tests:**
- `TestIsSubstrateVolumeContext` — nil, empty, wrong value, correct value
- `TestNodeStageVolumeSubstrateBypassesStagingPathCheck` — non-kubelet path accepted
- `TestNodeStageVolumeNonSubstrateRejectsBadPath` — non-kubelet path rejected without substrate-mode
- `TestNodePublishVolumeSubstrateBypassesTargetPathCheck` — same for publish
- `TestNodePublishVolumeNonSubstrateRejectsBadPath` — same for publish

**E2E on ACK cluster:**
- Controller passthrough: `"AgenticFS: server provided, using existing AccessPoint"` ✅
- Node mount: `"substrate mode detected, skipping pod runtime lookup"` → `"injected STS token from node metadata service"` → `"Mount success"` ✅


AlbeeSo:
Closing: this PR incorrectly depended on #1859. Substrate CSI support will be submitted as an independent PR covering all storage types (NAS, OSS, AgenticFS), not just AgenticFS.

